PR #38, opened 2020-11-24
Merge nightly changes to daemon v2

ongrid:


vshvsh:
I don't think we should proceed with this algorithm. It doesn't offer any additional robustness in the fair weather and won't work right in the event of delayed finality on the mainnet (akin to one that happened in medalla).  

Finality issues in eth2 can be imagined as digital tv with bad signal. Finality suddenly stops on epoch X, doesn't happen for, potentially, weeks, and then catches up for weeks in a moment. Then it can be lost again and again for a long time. 

We don't want oracle to run uninterrupted for weeks, nor do we want it to use week-ago stale data to the oracle. 

The pseudocode for the single run of an oracle reporting that works in bad and good network weather alike should be something like this, IMO (it follow directly from the spec):

```
first_reportable_frame = get_last_reported_frame_from_eth1() + 1
last_final_frame = calculate_frame(last_eth2_final_epoch()) - 1
if last_final_frame>=first_reportable_frame:
    validator_keys = get_validator_keys_from_eth1()
    report(validator_keys = validator_keys, epoch = last_frame_epoch(last_final_frame))
```

Dry runs, demonization, scheduling, interactive mode (send tx if the operator agrees with the data), etc can be implemented seamlessly on top of that single run.

vshvsh (changes requested):
I think we should change the algorithm before merging that

vshvsh (approved):

vshvsh:
On second thought, it's a PR into a dev branch; let's merge and continue to improve from there.

app/beacon.py
```diff
@@ -41,23 +50,40 @@ def get_actual_slot(self):
...
+            logging.info(f'Pubkey: {pubkey[:12]} Balance: {balance} Gwei')
         balances = sum(balance_list)
+        logging.info(f'Validator balances sum: {balances} Gwei')
+
         # Convert Gwei to wei
         balances *= 10 ** 9
-        return balances
+        total_validators_on_beacon = len(found_on_beacon_pubkeys)
+
+        return balances, total_validators_on_beacon
```
skozin, line 86: Prysm is still broken right now since `Prysm.get_balances` should return two values: `balances` and `total_validators_on_beacon`, like `Lighthouse.get_balances` does. But `Prysm.get_balances` only returns `balances` in this branch. Have you tried running it with Prysm before committing https://github.com/lidofinance/lido-oracle/pull/38/commits/9e703b97e58c4efe129ed9e383224b47a3ca5cfa#diff-c54a407f57e30d59dbaffdba7dc0bb249b021ff7de1073d7d94ab525b022ff5fL20?

ongrid:
@skozin Thanks. Cherry-picked related hunk from @ujenjt's commit 3ba6a8c0e4e471da4e98f051c5e4d024b6243d3c
```
@@ -18,7 +23,10 @@ def get_beacon(provider, slots_per_epoch):
         return Lighthouse(provider, slots_per_epoch)
     version = requests.get(urljoin(provider, 'eth/v1alpha1/node/version')).text
     if 'Prysm' in version:
-        return Prysm(provider, slots_per_epoch)
+        logging.error(f'Not supporting Prysm beacon node')
+        exit(1)
+        # TODO: fix me 
+        # return Prysm(provider, slots_per_epoch) 
     raise ValueError('Unknown beacon')
```

Issue #46, opened 2020-11-30
Sanity checks on inferred data


vshvsh:
There should be sanity checks on inferred data that forbid submitting a tx unless `--force` flag is set. More or less along the lines of "rewards can't exceed 20%/365  per day, slashings beyond reported only with `--force` flag, do not report duplicates" etc.

ongrid:
Some notes:
Now the Oracle contract doesn't check the relative numbers like reward - it's in backlog and hasn't been implemented. When it will be there, the daemon will discover that TX is insane via TX calling (it's there in both `pyrmont` and `daemon_v2`). 

The ability to execute the TX is already checked just before sending on the net by replaying it in the VM (see https://github.com/lidofinance/lido-oracle/blob/86d343dc415d373892c7aefdd38713787b55fc9a/app/oracle.py#L195 ). Now it looks as traceback (without process exit), but it's planned to be wrapped in human-readable log message (See #44). 

The on-chain reverts are possible and depend on timing (more aggressive WAIT - more probable tx races).
These effects can be minified by:
* improving the algorithm to avoid synchronisation around the global event (now daemons tend to synchronize around the moment of frame switch in ETH1). Working on it. If we switch to `look back` behavior instead of `waiting` the new frame, this won't be the issue.
* some sort of collision detection with rollbacks (watch for the pool, and if there is already the number of TXes that exceed the quorum, the exceeded ones should be self-identified and cancelled by initiators. Need to specify some kind of pseudorandom tiebreaker, stepping and feedback similarly to Physical collision detection on the wire protocols. This also will need to have the node with web3.pool exposed (this won't work via infura and other public endpoints).

vshvsh:
We don't have the capacity to implement it in the contracts so we're doing it in the daemon. The issue is to prevent massively wrong (due to a bug) data to propagate on chain.

ongrid:
Discussed on the calls - it's (at some degree) possible within the contracts and relies on the ability to virtually play transaction in EVM before serializing on the net. This costs no gas and we already use this approach. Related issues:


Actual logic and math:
[lido-dao#176](https://github.com/lidofinance/lido-dao/issues/176)
[lido-dao#175](https://github.com/lidofinance/lido-dao/issues/175)
[lido-dao#39](https://github.com/lidofinance/lido-dao/issues/39)

Cosmetics (hide EVM traceback, show human-readable log):
#44

[...]

Issue #44, opened 2020-11-27
Informative log instead of printing  'execution reverted: EPOCH_IS_TOO_OLD' traceback


ongrid:
When local TX call fails on the line,

https://github.com/lidofinance/lido-oracle/blob/86d343dc415d373892c7aefdd38713787b55fc9a/app/oracle.py#L195 

the daemon prints traceback, and it can be perceived by the user that something went wrong (Looks exactly as the "traceback of the death"). Moreover, the process after it pauses for WAIT time.

```
ERROR 2020-11-27 13:06:46 <daemon> Unexpected exception. 
Traceback (most recent call last):
  File "oracle.py", line 191, in <module>
    w3.eth.call(tx)
  File "/usr/local/lib/python3.8/site-packages/eth_utils/functional.py", line 45, in inner
    return callback(fn(*args, **kwargs))
  File "/usr/local/lib/python3.8/site-packages/web3/eth.py", line 432, in call
    return self.web3.manager.request_blocking(
  File "/usr/local/lib/python3.8/site-packages/web3/manager.py", line 153, in request_blocking
    raise ValueError(response["error"])
ValueError: {'code': 3, 'message': 'execution reverted: EPOCH_IS_TOO_OLD', 'data': '0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001045504f43485f49535f544f4f5f4f4c4400000000000000000000000000000000'}
    INFO 2020-11-27 13:06:46 <daemon> Sleep 60 s.
```

What should be done just now:
* If Traceback matches "execution reverted: EPOCH_IS_TOO_OLD" substring, the daemon should display descriptive one-line WARN ing
* Recommend longer WAIT time for production (about one hour will be enough to make such race of reports very unlikely)

How to improve algorithm more dramatically:
* Preliminary reason of this - oracle  daemons tend to synchronize their loops to the moment of frame switching on ETH1 side. With aggressive timing - oracles observe this nearly simultaneously and try to poll and report. Only first N (where N=quorum) will succeed.
* In the wider scope, the algorithm should be improved to avoid such synchronisation issues, will clarify the proposal in #21

skozin:
Do you mean the check implemented here?

[...]

Issue #41, opened 2020-11-27
Supervised (interactive) mode with TX confirmation


ongrid:
For manual debugging we should implement interactive mode supervised by human to avoid issues like contract state poisoning.

If the process started with `DAEMON=0` env it should print out the TX object via `__repr__()` then break before actual TX sending with the following interactive prompt.

```
Should we sent this TX? [y/n]: 
Please respond with [y/n]: y
```

Need additional checks if docker needs any tricks to expose interactive terminal capabilities. 

Initiated by @ujenjt on behalf of @vshvsh

ongrid:
the flow within Docker should be the following
* image runs in detached mode (default behavior)
* if `DAEMON=0` it stops in the situation which needs attendant's approval
* attendant sees that it happened and process paused via `docker logs`. Container is still up and running.
* attendant jount to the process via `docker attach`
* attendant sends `y` character
* docker instance continues its operation
* attendant closes the terminal docker continues
* docker instance continues its operation

@onionglass check if that's true.

ongrid:
Need to make dirty auto-test withing separate action (allowed to fail yet). Should fail with `connection refused`.

PR #48, opened 2020-11-30
WIP: Daemon v0.2.0 release

ongrid:
**This PR will be merged just prior bumping v0.2.0 release**
**The conflicts will be resolved by accepting incoming changes where applicable (just prior v0.2.0 release also)**
Reason: code review + hightlighting the required actions prior v0.2.0 mainnet release. Requested by @vshvsh 

Release scope:

- [ ] Sanity checks on inferred data #46 (needs to be implemented in Oracle contract. Daemon should just decorate failing calls with calling TX locally, need to implement checks in [lido-dao#39](https://github.com/lidofinance/lido-dao/issues/39)
- [ ] Informative logs if TX calling throws #44
- [ ] Falling into interactive mode before sending TX to the network. For v0.2.0 all network interactions should be intercepted by human confirmation when running in non-daemon mode `DAEMON=0` #41
- [ ] Testing with Prysm in E2E and testnet.
- [ ] Testing with upgraded contracts (lido-dao v0.2.1 ? which includes gas optimizations and in-contract sanity checks) in E2E and testnet.
- [ ] Polling algorithm refactoring #43 

Out of scope for v 0.2.0:

- [ ] Stateful algorithm that is able to intellectually treat more complex issues not leading to EVM reverts
- [ ] Quorum-finalizing tx-es are still subject to race, just one will succeed. This can be partially solved by increasing `WAIT` time from seconds to hours. With longer wait times race is much less likely to happen.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+wheel==0.36.2
 attrs==20.3.0
 base58==2.0.1
 bitarray==1.2.2


vshvsh:
Will repeat what I said in the previous pull request: this polling algorithm is not the one we can proceed with. There shouldn't be a significant pause between polling eth1 and eth2 data.
https://github.com/lidofinance/lido-oracle/pull/38#issuecomment-73382700

vshvsh:
Waiting for finality in the daemon will not work correcty in the majority of events when the network actually loses finality.

PR #54, opened 2020-12-03
Write informative one-line warning instead of traceback #44

vsmelov:
Write informative one-line warning instead of traceback.
Increase SLEEP in production to 3600sec.

### DIFF
--- a/app/oracle.py
+++ b/app/oracle.py
@@ -213,13 +213,17 @@ while True:
                 logging.info(f'Tx data: {tx.__repr__()}')
                 if prompt('Should we sent this TX? [y/n]: ', ''):
                     sign_and_send_tx(tx)
-        except:
-            logging.error('Unexpected exception. ')
-            traceback.print_exc()
-            if run_as_daemon:
-                logging.info(f'Sleep {await_time_in_sec} s.')
-                time.sleep(await_time_in_sec)
-                continue
+        except Exception as exc:
+            if isinstance(exc, ValueError) and 'execution reverted: EPOCH_IS_TOO_OLD' in str(exc):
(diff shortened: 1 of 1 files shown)


ongrid (changes requested):
* The issue-related changes should be combined into the separate commit. Other ideas should be wrapped into separate ones.

* Follow the simple rule - one finalized idea per commit. In this project we are actively cherry-picking commits so they should be absolutely clear and concise. If someone picks a single commit of yours and leave 'fixing' commit untransferred, he will definitely pollute the destination code.

* Commit message should be easy readable in online log mode `git log --oneline`. Follow these recommendation https://chris.beams.io/posts/git-commit/ and reword.

In your case we should have 2 commits

* on the top (HEAD) - your proposal to change ENV. It should be easily droppable via `git reset HEAD^1` 
* in the bottom  commit - your changes related to the task.

You can rewrite the git history via interactive rebase (not so easy) or just resetting and staging changes back in correct order (easier).

docker-compose.yml
```diff
@@ -3,6 +3,7 @@ services:
   oracle:
     build: .
     environment:
+      SLEEP: 3600
```
ongrid, line 6: I'd suggest to keep the default value. It's optimal for E2E. Drop the HEAD commit.

ongrid (changes requested):

ongrid:
@Klyaus @onionglass could you please fix the tests? Our approach with patching is too fragile :-(. I think we should discard it and use primitive server for mocking JSON-RPC of web3 and REST of beacon. Please drop interaction tests for now.

Issue #57, opened 2020-12-05
Fix logging STDOUT/STDERR streams


ongrid:
All logs by default should go to STDOUT (especially logs with INFO level of severity)
Interaction with the user should go to STDOUT
errors should go to STDERR

Tests should assert the exact streams

ongrid:
* Base: `daemon_v2 HEAD` (after merging prysm support)
* PR target: `daemon_v2`
* Tests should cover the changes
* Tests should pass

ongrid:
@onionglass could you please adapt @vsmelov 's code to our updated `daemon_v2` branch and fix the issue you highlighted in PR?

PR #63, opened 2020-12-08
split logging to stdout and stderr #57

vsmelov:
add init_log function

### DIFF
--- a/app/log.py
+++ b/app/log.py
@@ -1,0 +1,31 @@
+import sys
+import logging
+
+
+def init_log():
+    """ initialize root logger """
+    assert not hasattr(init_log, '_called'), 'init_log must be called only once!'
+    init_log._called = True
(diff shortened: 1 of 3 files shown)


ongrid (changes requested):
Add a single test that runs the app and asserts at least two messages : one STDOUT and one STDERR. Suitable place for such test is before interaction tests https://github.com/lidofinance/lido-oracle/blob/a56f7155fb68de275785e5b3ba105694ea7ea790/tests/test_oracle.py#L4

onionglass:
beacon.py ans contract.py also import logging

ongrid (changes requested):
@onionglass could you please adapt this?

Klyaus (approved):

ongrid (approved):

Klyaus (approved):

Issue #68, opened 2020-12-11
Tests. Mute curl connect failure messages


ongrid:
```
./run_tests.sh 
Performing all tests
Run ETH1/ETH2 mock webservice
Switch mock to Lighthouse
curl: (7) Failed to connect to 127.0.0.1 port 8080: Connection refused
Run python tests
```

Base: `daemon_v2`

ongrid:
Fixed in https://github.com/lidofinance/lido-oracle/commit/82ef2cbed0e7841f684b3f19d2e767121517ac61 .
Changes are in `daemon_v2` now

Issue #67, opened 2020-12-11
eth_account. privateKeyToAccount is deprecated. Switch to from_key method.


ongrid:
The notice about `privateKeyToAccount` method removal.
https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.account.Account.privateKeyToAccount 

Need to switch to the new `Account.from_key` interface
https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.account.Account.from_key

We use it to calculate the member's account here
https://github.com/lidofinance/lido-oracle/blob/a56f7155fb68de275785e5b3ba105694ea7ea790/app/oracle.py#L75 

Base Branch: `daemon_v2`

PR #70, opened 2020-12-11
Use from_key instead of privateKeyToAccount #67

Klyaus:
Switch privateKeyToAccount method to from_key https://github.com/lidofinance/lido-oracle/issues/67
Mute curl connect failure messages in tests https://github.com/lidofinance/lido-oracle/issues/68


run_tests.sh
```diff
@@ -26,6 +26,7 @@ fi
 
 echo "Run ETH1/ETH2 mock webservice"
 python -m aiohttp.web -H localhost -P 8080 helpers.eth_nodes_mock:main > /dev/null 2>&1 &
```
ongrid: the same came with just-merged code. Not actual anymore

ongrid (approved):

PR #101, opened 2021-01-09
Fix timestamps of last epoch. Increase MAX_APR

### LINKED ISSUE #100: Wrong timestamp of last epoch in APR calculations. Increase MAX_APR
APR calculation currently (v0.1.2) uses the timestamp of the event of the previous report. Since the oracle reports reflect the balances on the start of the epoch, this should also be used in APR calculations.

BTW need to increase `MAX_APR` to 0.15

ongrid:
closes #100

### DIFF
--- a/app/metrics.py
+++ b/app/metrics.py
@@ -30,11 +30,10 @@ class PoolMetrics:
         return self.getTransientValidators() * self.DEPOSIT_SIZE
 
 
-def get_previous_metrics(w3, pool, oracle, beacon_spec, from_block = 0):
+def get_previous_metrics(w3, pool, oracle, beacon_spec, from_block=0):
     """Since the contract lacks a method that returns the time of last report and the reported numbers
     we are using web3.py filtering to fetch it from the contract events."""
     logging.info('Getting previously reported numbers (will be fetched from events)...')
-    epochs_per_frame = beacon_spec[0]
     genesis_time = beacon_spec[3]
     result = PoolMetrics()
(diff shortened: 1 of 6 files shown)


app/metrics.py
```diff
@@ -44,24 +45,27 @@ def get_previous_metrics(w3, pool, oracle, genesis_time, epochs_per_frame, from_
...
+            break
 
-    # If the list of events is empty, we consider it's the first run
-    result.epoch = 0
-    result.timestamp = latest_block['timestamp']
+    # If the epoch has been assigned from the last event (not the first run)
+    if result.epoch:
+        result.timestamp = get_timestamp_by_epoch(beacon_spec, result.epoch)
+    else:
+        # If it's the first run, we set timestamp to genesis time
```
vshvsh, line 61: Don't we need to assign something to result's epoch?

ongrid: It's assigned in the class https://github.com/lidofinance/lido-oracle/blob/cf0622714752deaf88866d4ea92a114c4e55cf64/app/metrics.py#L14 . The default value of 0 is a pretty good marker for future use on consumer side (to check for corner-cases similar to https://github.com/lidofinance/lido-oracle/blob/cf0622714752deaf88866d4ea92a114c4e55cf64/app/metrics.py#L122 and https://github.com/lidofinance/lido-oracle/blob/cf0622714752deaf88866d4ea92a114c4e55cf64/app/metrics.py#L126)

vshvsh (approved):

Issue #102, opened 2021-01-09
Don't fetch beacon balances for non-reportable epochs


ongrid:
Oracle should avoid beacon balance polling if the epoch has been reported already.

skozin:
Seems to be fixed in 0c9d989?

skozin:
Yep, fixed.

PR #103, opened 2021-01-09
WIP: Introduce FSM and partial states #102

ongrid:


vshvsh:
A bit too complex for my taste. Will think what we can do with it.

PR #104, opened 2021-01-09
Introduce dual-stage get_current_metrics() #102

### LINKED ISSUE #102: Don't fetch beacon balances for non-reportable epochs

ongrid:
The first call of get_current_metrics(...) returns just partial result
by default. Validators keys and balances get fetched only if
the function called with the result of the previous call (partial).

It's a fast workaround to have 'stages' without complex FSM logic.

closes #102

### DIFF
--- a/app/metrics.py
+++ b/app/metrics.py
@@ -62,33 +62,38 @@ def get_previous_metrics(w3, pool, oracle, beacon_spec, from_block=0):
     return result
 
 
-def get_current_metrics(w3, beacon, pool, oracle, registry, beacon_spec):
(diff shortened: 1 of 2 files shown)


vshvsh:
Hacky but works. I think we'll use it.

ongrid:
@vsmelov please handover this. Actually, it's ready to be merged as-is. Regarding a more clear and idiomatic approach see the idea to introduce FSM #103 (just a sketch to describe the idea)

Issue #56, opened 2020-12-03
Prometheus Metrics


joe-bowman:
The oracle should expose a prometheus metrics endpoint, to track vital values (e.g. eth balance, block heights, etc.)

ongrid:
Hey @joe-bowman. We are considering to implement one of two possible options:
* Make the oracle daemon "scrapeable". It should expose HTTP Endpoint that will be polled by Prometheus. You mean this option, right? This needs more configuration and more efforts to protect the endpoint from external threats.
* Exporting (pushing) metrics to the Prometheus gateway. This approach seems to be more secure (less exposure, less attack surface). Does it suit your and typical Prometheus setup?

joe-bowman:
Scrapable (pull model) is preferred, push model comes with caveats (less trivial to assert something is ‘down’ because push gateway will keep the last value. It can be worked around, by pushing a counter with the current unix time as a value (last_push) so you can alert on that not increasing.

Pull is preferred from a ops point of view, but appreciate the desire to minimise attack surface.

vshvsh:
@joe-bowman we're now discussing a decision which is a blocker to implementing the metrics. 
Should an oracle run as a daemon (and export metrics over an endpoint) or a cronable single-run executable (can write metrics to file that is served by a separate exporter)? 
Given that the oracle is stateless currently, and should fire once per day at a predictable time, I'd prefer the second option for the time being. What do you think?

ongrid:
The daemon runs in the docker container and won't rely on any task sheduler. The container and the process inside are always up (the main oracle coroutine sleeps and wakes up once a day or several hours). This doesn't prevent another coroutine to serve external scrap requests or export values to the Prometheus GW using built-in async mechanisms. So both options are possible. The Exporting (pushing) just needs less efforts in terms of security and more efforts on mainainer's side to configure Prometheus GW.

I'd prefer to give our operators option 1 (scrapable) but with the strong requirement to run the container inside the perimeter + introduce in-process whitelisting + kernel-level firewall rules. On the later stage I'd extend the option 1 with some proven authorization mechanism.

joe-bowman:
Push works for me; I appreciate the desire to minimise attack surface.

Presumably it fires more frequently than once a day though? (i.e. each time it polls for new blocks). We don't want to only find out it is down when it has failed to update. Being alerted early to the fact it is not syncing is very useful.

vshvsh:
[...]

PR #105, opened 2021-01-14
prometheus metrics #56

vsmelov:
add prometheus metrics


app/oracle.py
```diff
@@ -266,4 +285,17 @@ def prompt(prompt_message, prompt_end):
         break
 
     logging.info(f'We are in DAEMON mode. Sleep {await_time_in_sec} s and continue')
-    time.sleep(await_time_in_sec)
+
+    # sleep and countdown
+    awake_at = time.time() + await_time_in_sec
+    while time.time() < awake_at:
```
vshvsh, line 291: I think the main metric update be done in the main loop to reuse calculations done there.
 
getPreviousMetrics are easy on the node so can be done at any time but, alas, changes only once in 24h.
getCurrentMetrics is very heavy load, it's better to reuse the crunched numbers.

I can see simple metrics like "current block of eth1/current slot of eth2" exported in this loop though.

vshvsh (changes requested):

app/oracle.py
vsmelov: done, I removed recalculation of get_current_metrics, get_previous_metrics

PR #128, opened 2021-11-22
eip-1559 implemented

F4ever:


F4ever:
Done:
- Switched to tx type 2
- Upgraded web3
- Remove unused and redundant libs

Psirex (approved):

PR #130, opened 2021-12-08
Fix dockerignore file

F4ever:
Dockerfile doesn't work because app and assets folders are excluded incorrectly.


avsetsin (approved):
lgtm

PR #131, opened 2021-12-10
Add prefixes for prometheus metrics

F4ever:
Added `PROMETHEUS_PREFIX` env variable that can be used to configure prefixes for metrics

### DIFF
--- a/app/prometheus_metrics.py
+++ b/app/prometheus_metrics.py
@@ -1,7 +1,6 @@
 import os
 import logging
 
-from prometheus_client import Gauge
 from prometheus_client.metrics import Gauge, Histogram
 
 from pool_metrics import PoolMetrics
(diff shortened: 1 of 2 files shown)


app/prometheus_metrics.py
```diff
@@ -1,8 +1,7 @@
 import os
 import logging
-import resource
 
-from prometheus_client import start_http_server, Gauge, Summary
+from prometheus_client import Gauge
 from prometheus_client.metrics import Gauge, Histogram
```
avsetsin, line 5: Looks like there are two Gauge imports

avsetsin (approved):
lgtm

app/prometheus_metrics.py
F4ever: Fixed

PR #135, opened 2021-12-21
Use modern lido-sdk library, increase performance

F4ever:
- Moved from Lido to lido-sdk library
- Remove base dependencies from req-test and fixed workflow
- Increase performance of get_balances for Lighthouse client


avsetsin (approved):
lgtm

PR #148, opened 2022-03-02
fix(build): build script improvements

sambacha:
## Build Script fixes

- fixes silent failure in which the value of $? is overwritten by [/[[ i.e.  you can't get the original value in the META_INFO check
- sets DOCKER_BUILDKIT env to leverage buildkit build system
- explicitly set exit status code to `exit 1` on failure 

### Additional stuff
- consider: using docker buildx build
- consider comparing python script to git command, e.g.

```sh
GIT_SHA=${GITHUB_COMMIT:-$(git rev-parse HEAD)}
export GIT_SHA=$GIT_SHA

if [ "$GIT_SHA" = "$VERSION" ]; then
    echo "GIT SHA Versions are equal"
else
    echo "ERROR: GIT SHA Versions are not equal"
fi
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -15,32 +15,31 @@ export DOCKER_CONFIG=$HOME/.lidofinance
 
 #### Version info ####
 META_INFO=$(python ./helpers/get_git_info.py 2>&1)
-
-if [[ $? != 0 ]]
-then
-    echo "ERROR: python script get_git_info.py"
+if ! META_INFO=$(python ./helpers/get_git_info.py 2>&1); then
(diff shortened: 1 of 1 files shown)


avsetsin (commented):
@F4ever seems like we don't use this script anymore, right? These changes probably need to be moved to our CI

sambacha:
> @F4ever seems like we don't use this script anymore, right? These changes probably need to be moved to our CI

Thats what I was thinking to be honest, but it was referenced in the README for building

PR #149, opened 2022-03-02
fix(docker): healthcheck and python error

sambacha:
fixes health check command
fixes python to force the stdout and stderr streams to be unbuffered (see logs in console) 
fixes pip to not use cache

consider: init system usage like tini

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,26 @@
-FROM python:3.8-slim as builder
+FROM python:3.8-buster-slim as builder
 
-RUN apt-get update && apt-get install -y gcc && apt-get install -y g++
+ENV LANG=C.UTF-8 \
+    DEBIAN_FRONTEND=noninteractive \
+    PIP_NO_CACHE_DIR=true
+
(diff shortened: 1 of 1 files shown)


avsetsin (commented):
@sambacha thanks for the PR, LGTM

sambacha:
> @sambacha thanks for the PR, LGTM

Would you be open to accepting an additional PR for adding goss for smoke testing purposes? 

Thanks for the quick turnaround, much appreciated

avsetsin:
> Would you be open to accepting an additional PR for adding goss for smoke testing purposes?
Thank you for your efforts! I'm not sure we need it since we already have a batch of smoke tests in our infra (it's a separate private repository, so I can't reference it)

sambacha:
> > Would you be open to accepting an additional PR for adding goss for smoke testing purposes?
> 
> Thank you for your efforts! I'm not sure we need it since we already have a batch of smoke tests in our infra (it's a separate private repository, so I can't reference it)

that was what I was suspecting :) 

cheers

PR #153, opened 2022-06-10
Feature/move to poetry

F4ever:


.github/workflows/linters.yml
```diff
@@ -8,30 +8,30 @@ jobs:
...
-        # This path is specific to Ubuntu
-        path: ~/.cache/pip
-        # Look to see if there is a cache hit for the corresponding requirements file
-        key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
-        restore-keys: |
-          ${{ runner.os }}-pip-
-          ${{ runner.os }}-
+        python-version: 3.9
+
+    - uses: Gr1N/setup-poetry@v7
```
avsetsin, line 17: As far as I remember the last agreement was not to use third-party actions directly. Only by using a fork

Dockerfile
```diff
@@ -1,31 +1,39 @@
...
 
 ENV LANG=C.UTF-8 \
     DEBIAN_FRONTEND=noninteractive \
     PIP_NO_CACHE_DIR=true
 
-RUN apt-get update && apt-get install --no-install-recommends -qq -y gcc g++ && apt-get clean autoclean \
-  && apt-get autoremove -y \
-  && rm -rf /var/lib/apt/lists/* \
-  && rm -f /var/cache/apt/archives/*.deb
+RUN apt-get update && apt-get install -y --no-install-recommends -qq gcc=4:10.2.1-1 libffi-dev=3.3-6 g++=4:10.2.1-1 git=1:2.30.2-1 curl=7.74.0-1.3+deb11u1 \
```
avsetsin, line 7: Are extra packages needed here for sure?

Dockerfile
```diff
@@ -1,31 +1,39 @@
...
+FROM python:3.9-slim as base
 
 ENV LANG=C.UTF-8 \
     DEBIAN_FRONTEND=noninteractive \
     PIP_NO_CACHE_DIR=true
 
-RUN apt-get update && apt-get install --no-install-recommends -qq -y gcc g++ && apt-get clean autoclean \
-  && apt-get autoremove -y \
-  && rm -rf /var/lib/apt/lists/* \
-  && rm -f /var/cache/apt/archives/*.deb
```
avsetsin, line 10: Is the new version a proper replacement for this one?

pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[tool.poetry]
+name = "lido-oracle"
+version = "2.1.1"
+description = "Oracle daemon for Lido decentralized staking service. Collects and reports Ethereum 2.0 beacon chain states (the number of visible validators and their summarized balances) to the Lido dApp contract running on Ethereum 1.0 side."
+authors = ["Raman Siamionau <raman.s@lido.fi>"]
```
avsetsin, line 5: Could you please also mention folks who were involved in development before the oracle was transferred to the tooling team for support?

Dockerfile
```diff
@@ -1,31 +1,39 @@
...
+
+RUN . /venv/bin/activate && poetry install --no-dev --no-root
+
+FROM base as production
+
+COPY --from=builder /venv /venv
 
 RUN mkdir -p /var/www && chown www-data /var/www && \
-    apt-get update && apt-get install --no-install-recommends -qq -y curl && \
     apt-get clean && find /var/lib/apt/lists/ -type f -delete && \
```
avsetsin, line 28: - don't we need `curl`? it is used in the healthcheck
- there is still `apt-get clean' but `apt-get update && apt-get install` was removed

Dockerfile
```diff
@@ -50,14 +58,14 @@ ENV BRANCH=${BRANCH}
...
+EXPOSE $PROMETHEUS_PORT
 USER www-data
 
 COPY --from=builder /usr/local/ /usr/local/
 COPY assets ./assets
 COPY app ./
 
 HEALTHCHECK --interval=10s --timeout=3s \
-    CMD curl -f http://localhost:8000/healthcheck || exit 1
+    CMD curl -f http://localhost:$PULSE_SERVER_PORT/healthcheck || exit 1
```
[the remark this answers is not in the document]
F4ever: Yeah, we don't have correct health check. Currently healthcheck fetches prometheus url in oracle.
I'll add correct healthcheck in next updates

.github/workflows/linters.yml
F4ever: Removed

Dockerfile
F4ever: Installing curl is going in "base" stage
clean and lists removed, because executed on base stage

pyproject.toml
F4ever: Oh, generated automatically. I removed myself from list. 
Added top 3 developers to list, but they don't work in Lido now, so I just added their names.

Dockerfile
F4ever: Not sure what this is used for. After builded container check this folder and there is no deb files, so this command is redundant

Dockerfile
F4ever: Removed

app/beacon.py
```diff
@@ -125,6 +125,10 @@ def __init__(self, url, slots_per_epoch):
         self.slots_per_epoch = slots_per_epoch
         self.version = requests.get(urljoin(url, self.api_version)).json()
 
+        # if send_tx_yesterday than sleep 5 minutes
+        # + poetry
+        # + flashbots
```
avsetsin, line 130: Why do we need flashbots for oracle?

avsetsin (approved):

PR #156, opened 2022-06-21
Feature: alternate oracles

F4ever:


app/oracle.py
```diff
@@ -152,6 +152,16 @@
...
 
+quorum_size = oracle.functions.getQuorum().call()
+oracle_members = oracle.functions.getOracleMembers().call()
+
+if not dry_run and account:
+    try:
+        oracle_holder_index = oracle_members.index(account.address)
+    except ValueError:
+        oracle_holder_index = 0
+        logger.warning('Account is not oracle member.')
```
avsetsin, line 163: I think it's better to get the quorum and member index just before the report. This would only happen once a day, but we wouldn't need to ask oracle owners to restart in case of any parameter changes in the contract

avsetsin (approved):
lgtm

PR #157, opened 2022-07-21
[Lido oracle] Oracle v2.3.0 release

F4ever:
- Switch to poetry
- Alternate oracles (sleep for 8 minutes) depends on date. To make sure all oracles are online


Dockerfile
```diff
@@ -1,31 +1,37 @@
...
 
 RUN mkdir -p /var/www && chown www-data /var/www && \
-    apt-get update && apt-get install --no-install-recommends -qq -y curl && \
-    apt-get clean && find /var/lib/apt/lists/ -type f -delete && \
-    chown www-data /app/
+    chown -R www-data /app/ && chown -R www-data /venv
 
-ENV PATH=$PATH:/usr/local/bin
-ENV PYTHONPATH="/usr/local/lib/python3.8/site-packages/"
+ENV PYTHONPATH="/venv/lib/python3.9/site-packages/"
```
sergeyWh1te, line 29: It's common question. Why do we need to hard code docker envs? In most case we prepared .env config which will be uploaded into bash profile on deployed machine.

So, I wanted to highlight that if we posted them in .env file => ci/di could provide them into docker setup/config and etc

sergeyWh1te (approved):

avsetsin (approved):
lgtm

PR #158, opened 2022-07-22
[Lido oracle] Add multi provider and split function

F4ever:
- Split big function in two
- Add multi provider support

### DIFF
--- a/app/metrics.py
+++ b/app/metrics.py
@@ -44,9 +44,7 @@ def get_previous_metrics(w3, pool, oracle, beacon_spec, from_block=0) -> PoolMet
 
 
 def get_light_current_metrics(w3, beacon, pool, oracle, beacon_spec):
-    """If the result of previous get_current_metrics call isn't given
-     create and return partial metric.mSince it doesn't get keys from
-     registry and doesn't retrieve beacon state, it's much faster."""
+    """Fetch current frame, buffered balance and epoch"""
     epochs_per_frame = beacon_spec[0]
     partial_metrics = PoolMetrics()
     partial_metrics.blockNumber = w3.eth.getBlock('latest')['number']  # Get the the epoch that is both finalized and reportable
@@ -66,7 +64,7 @@ def get_light_current_metrics(w3, beacon, pool, oracle, beacon_spec):
 
 
 def get_full_current_metrics(w3, beacon, beacon_spec, partial_metrics) -> PoolMetrics:
-    """If partial result provided, the oracle fetches all the required states from ETH1 and ETH2"""
+    """The oracle fetches all the required states from ETH1 and ETH2 (validator balances)"""
     slots_per_epoch = beacon_spec[1]
     slot = partial_metrics.epoch * slots_per_epoch
     logging.info(f'Reportable state: epoch:{partial_metrics.epoch} slot:{slot}')
--- a/app/metrics.py
+++ b/app/metrics.py
@@ -47,7 +47,7 @@ def get_light_current_metrics(w3, beacon, pool, oracle, beacon_spec):
     """Fetch current frame, buffered balance and epoch"""
     epochs_per_frame = beacon_spec[0]
     partial_metrics = PoolMetrics()
-    partial_metrics.blockNumber = w3.eth.getBlock('latest')['number']  # Get the the epoch that is both finalized and reportable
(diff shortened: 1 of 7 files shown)


app/oracle.py
```diff
@@ -101,15 +103,7 @@
...
-if eth1_provider.startswith('http'):
-    provider = HTTPProvider(eth1_provider)
-elif eth1_provider.startswith('ws'):
-    provider = WebsocketProvider(eth1_provider)
-else:
-    logging.error('Unsupported ETH provider!')
-    exit(1)
-
-w3 = Web3(provider)
+w3 = Web3(MultiProvider(eth1_provider.split(',')))
```
F4ever, line 106: I can also rename WEB3_PROVIDER_URI -> WEB3_RPC_ENDPOINTS
What do you think?

app/metrics.py
```diff
@@ -43,29 +43,29 @@ def get_previous_metrics(w3, pool, oracle, beacon_spec, from_block=0) -> PoolMet
...
-        partial_metrics.timestamp = get_timestamp_by_epoch(beacon_spec, partial_metrics.epoch)
-        partial_metrics.depositedValidators = pool.functions.getBeaconStat().call()[0]
-        partial_metrics.bufferedBalance = pool.functions.getBufferedEther().call()
-        return partial_metrics
-
+     create and return partial metric.mSince it doesn't get keys from
+     registry and doesn't retrieve beacon state, it's much faster."""
+    epochs_per_frame = beacon_spec[0]
+    partial_metrics = PoolMetrics()
+    partial_metrics.blockNumber = w3.eth.getBlock('latest')['number']  # Get the the epoch that is both finalized and reportable
```
sergeyWh1te, line 52: Get the epoch that is finalized and reportable

app/metrics.py
```diff
@@ -43,29 +43,29 @@ def get_previous_metrics(w3, pool, oracle, beacon_spec, from_block=0) -> PoolMet
...
+    partial_metrics.epoch = min(potentially_reportable_epoch,
+                                (finalized_epoch_beacon // epochs_per_frame) * epochs_per_frame)
+    partial_metrics.timestamp = get_timestamp_by_epoch(beacon_spec, partial_metrics.epoch)
+    partial_metrics.depositedValidators = pool.functions.getBeaconStat().call()[0]
+    partial_metrics.bufferedBalance = pool.functions.getBufferedEther().call()
+    return partial_metrics
+
+
+def get_full_current_metrics(w3, beacon, beacon_spec, partial_metrics) -> PoolMetrics:
     """If partial result provided, the oracle fetches all the required states from ETH1 and ETH2"""
```
sergeyWh1te, line 69: If a(article) partial result IS(passive voice) provided, the oracle fetches all the required states from ETH1 and ETH2

app/oracle.py
sergeyWh1te: I suggest for using just "Provider" name with []rpc_endpoints signature.
Provider([]rpc_endpoints signature), and if it's possiblie make annottation for "Provider" like this.

```
@endpoints - an array with bla...bla
@returns bla bla bla
func Provider(endpoints []rpc_endpoints signature): Provider
```

app/oracle.py
```diff
@@ -363,14 +357,15 @@ def update_beacon_data():
...
     # Get minimal metrics that are available without polling
-    current_metrics = get_current_metrics(w3, beacon, pool, oracle, registry, beacon_spec)
+    current_metrics = get_light_current_metrics(w3, beacon, pool, oracle, beacon_spec)
     metrics_exporter_state.set_current_pool_metrics(current_metrics)
+
     if current_metrics.epoch <= prev_metrics.epoch:  # commit happens once per day
         logging.info(f'Currently reportable epoch {current_metrics.epoch} has already been reported. Skipping it.')
         return
 
     # Get full metrics using polling (get keys from reggistry, get balances from beacon)
```
sergeyWh1te, line 367: Get full metrics using polling (getting keys from registry and balances from the beacon chain)

app/metrics.py
F4ever: Fixed

app/oracle.py
F4ever: Function overloading is not nice to do in python.

app/metrics.py
F4ever: Fixed

sergeyWh1te (approved):

PR #159, opened 2022-07-25
[Lido oracle] Add timeout exceptions

F4ever:


app/beacon.py
```diff
@@ -17,13 +17,13 @@
 
 
 def get_beacon(provider, slots_per_epoch):
-    version = requests.get(urljoin(provider, 'eth/v1/node/version')).text
+    version = requests.get(urljoin(provider, 'eth/v1/node/version'), timeout=60).text
```
sergeyWh1te, line 20: It appears more than 10 times.
I strongly believe in idea that in has to be a const.

F4ever: Fixed

sergeyWh1te (approved):

PR #164, opened 2022-10-04
New beacon client for Prysm

F4ever:


app/beacon.py
```diff
@@ -61,156 +58,48 @@ def inner(*args, **kwargs):
...
+        validator_pub_keys = self._from_bytes_to_pub_keys(keys_list)
 
-        for validator in response_json['data']:
-            pubkey = validator['validator']['pubkey']
-            # Log all validators along with balance
-            if pubkey in pubkeys:
-                validator_balance = int(validator['balance'])
-                total_balance += validator_balance
-                found_on_beacon_pubkeys += 1
+        df = pd.DataFrame.from_records([{**v, **v['validator']} for v in all_validators])
```
avsetsin, line 81: I don't really like that we're adding a new dependency here. The logic here wasn't complicated, I'm not sure if the extra library made the code much better, but we got an extra point of failure and potential vulnerability. We also need to retest this part of the logic, it seems to me that refactoring in this place is not much needed now

app/beacon.py
```diff
@@ -61,156 +58,48 @@ def inner(*args, **kwargs):
...
-        actual_slots = {}
-        response = session.get(urljoin(self.url, self.api_beacon_head), timeout=DEFAULT_TIMEOUT).json()
-        actual_slots['actual_slot'] = int(response['headSlot'])
-        actual_slots['finalized_slot'] = int(response['finalizedSlot'])
-        return actual_slots
-
-    @proxy_connect_timeout_exception
-    def get_balances(self, slot, key_list):
-        params = {}
-        pubkeys = []
```
avsetsin, line 176: Wasn't there a problem here? It seems that the last time we solved the performance problem, only the lighthouse version was fixed.

The problem is that searching in the array at each iteration in the loop is a very costly operation, since we are searching in an array of 160+ keys

avsetsin (approved):
lgtm

PR #171, opened 2022-11-28
Feat/oracle v3

F4ever:
New Oracle V3


src/oracle.py
```diff
@@ -0,0 +1,135 @@
+from time import sleep
```
sergeyWh1te, line 1: I'd like to offer to rename "oracle.py" into "main.py". 
I got it from the Dockerfile. "oracle.py" is a start point of all oracle applications and singular. Let's name it with a strict name correlation.

src/oracle.py
```diff
@@ -0,0 +1,135 @@
...
+            self.run_cycle()
+
+    def run_cycle(self):
+        self._fetch_beacon_specs()
+        epoch = self._fetch_next_finalized_epoch()
+        logger.info({'msg': 'Get finalized epoch.', 'value': epoch})
+
+        try:
+            self.run_once(epoch)
+        except KeyboardInterrupt as error:
```
sergeyWh1te, line 56: I'm not python expert. Moreover, "KeyboardInterrupt" tells me about user's or system app termination.
I'd prefer to move this signal under "Unexpected exception." and rename it like "SIGTERM" (signal for termination)

src/oracle.py
```diff
@@ -0,0 +1,135 @@
...
+        _, self.slots_per_epoch, self.seconds_per_slot, _ = contracts.oracle.functions.getBeaconSpec().call()
+        # _, self.slots_per_epoch, self.seconds_per_slot, _ = contracts.validator_exit_bus.functions.getBeaconSpec().call()
+
+    def run_as_daemon(self):
+        while True:
+            pulse()
+            logger.info({'msg': 'Run cycle.'})
+            self.run_cycle()
+
+    def run_cycle(self):
```
sergeyWh1te, line 49: "run_cycle" is a nice. 
Somewhere on bottom, I found a possibility to run this function only once by demanding. It represents an idea that one tact or cycle. Let's call it just "fire" or "handle"

src/oracle.py
```diff
@@ -0,0 +1,135 @@
...
+        self.modules: List[OracleModule] = [
+            Accounting(self._w3, self._beacon_chain_client),
+            Ejector(self._w3, self._beacon_chain_client),
+        ]
+
+    def _fetch_beacon_specs(self):
+        _, self.slots_per_epoch, self.seconds_per_slot, _ = contracts.oracle.functions.getBeaconSpec().call()
+        # _, self.slots_per_epoch, self.seconds_per_slot, _ = contracts.validator_exit_bus.functions.getBeaconSpec().call()
+
+    def run_as_daemon(self):
```
sergeyWh1te, line 43: Similar to "run_cycle". I'd like tp prefer to rename just "worker"

src/oracle.py
```diff
@@ -0,0 +1,135 @@
...
+        logger.info({
+            'msg': 'Fetch first proposed slot in epoch.',
+            'epoch': epoch,
+            'slot': slot_number,
+            'block_hash': block_hash,
+        })
+
+        for module in self.modules:
+            try:
+                module.run_module(slot_number, block_hash)
```
sergeyWh1te, line 97: Is it ok, that every time when our code runs, "the accounting module" works first, and after the "ejector module" works. So, In the case when the Accounting module throws an exception, it means that the "ejector" module won't do its work every time. Also, if the oracle has more "modules" or workers themselves, that pattern will add affect cascade interrupting behavior all application.

sergeyWh1te: I thought those "modules" or "workers" had to work independently of each other.

sergeyWh1te, line 97: Here, we're able to see that word "module" appears twice. 
For a cleaner style, make it just 
`module.run(slot_number, block_hash). 
`
In modules rename run_module func for just "run" or "handle" or "fire".

src/modules/accounting.py
```diff
@@ -0,0 +1,251 @@
...
+            last_requests_id_to_finalize,
+            finalization_shares_amount,
+        ))
+
+        return transaction
+
+    def _get_beacon_validators_stats(self, slot: SlotNumber, block_hash: HexBytes) -> Tuple[int, int]:
+        # Should return balance in gwei!
+        lido_validators = get_lido_validators(self._w3, block_hash, self._beacon_chain_client, slot)
+        return len(lido_validators), sum(int(validator['validator']['balance']) for validator in lido_validators)
```
sergeyWh1te, line 163: Is it ok sum validators' balances and cast value into (int). I'm afraid that that string can lose some affective balance, like 165.3242 ETH.

src/contract_utils/lido_keys.py
```diff
@@ -0,0 +1,58 @@
...
+
+def _get_dict_of_keys(validators: List[OperatorKey]) -> Dict[str, OperatorKey]:
+    """Key is pubkey in hex format. Value is data from key registry."""
+    return {
+        '0x' + binascii.hexlify(validator['key']).decode(): validator for validator in validators
+    }
+
+
+@lru_cache(maxsize=2)
+def get_lido_node_operators(w3: Web3, block_hash: HexBytes) -> List[ModifiedOperator]:
```
sergeyWh1te, line 22: I pos this code about lido validator, not the operator. Because of the - node operator, it's a human=)

src/contract_utils/lido_keys.py
```diff
@@ -0,0 +1,58 @@
...
+    # TODO Fix library - add fetch by block configuration and rewrite using not just NO module (in future)
+    lido = Lido(w3)
+    lido.get_operators_indexes()
+    return [ModifiedOperator(module_id='0x9D4AF1Ee19Dad8857db3a45B0374c81c8A1C6320', **op) for op in lido.get_operators_data()]
+
+
+@lru_cache(maxsize=2)
```
sergeyWh1te, line 29: Might it be a better strategy for caching is LFU? 
(last frequently used)

src/modules/ejection.py
```diff
@@ -0,0 +1,322 @@
...
+
+        return validators_to_eject
+
+    def _get_validator_to_eject(
+        self,
+        lido_validators: List[MergedLidoValidator],
+    ) -> Generator[MergedLidoValidator, None, None]:
+        """
+        Filter all exiting and exited validators.
+        Sort them by index.
```
sergeyWh1te, line 235: Here. Sorting by index might be outdated. Check it please

src/contract_utils/withdrawal_queue.py
```diff
@@ -0,0 +1,29 @@
...
+
+def get_withdrawal_requests_wei_amount(block_hash: HexBytes) -> int:
+    total_pooled_ether = contracts.lido.functions.getTotalPooledEther().call(block_identifier=block_hash)
+    logger.info({'msg': 'Get total pooled ether.', 'value': total_pooled_ether})
+
+    total_shares = contracts.lido.functions.getSharesByPooledEth(total_pooled_ether).call(block_identifier=block_hash)
+    logger.info({'msg': 'Get total shares.', 'value': total_shares})
+
+    queue_length = contracts.withdrawal_queue.functions.queueLength().call(block_identifier=block_hash)
+    logger.info({'msg': 'Get last id to finalize.', 'value': total_pooled_ether})
```
sergeyWh1te, line 19: "Get last id to finalize."
It could be better to log about queue length and size.  Also, the code doesn't log request_id from the el_queue.

src/contract_utils/withdrawal_queue.py
```diff
@@ -0,0 +1,29 @@
...
+
+    if queue_length == 0:
+        logger.info({'msg': 'Withdrawal queue is empty.'})
+        return 0
+
+    return contracts.withdrawal_queue.functions.calculateFinalizationParams(
```
sergeyWh1te, line 25: According to https://github.com/lidofinance/lido-dao/blob/feature/shapella-upgrade/contracts/0.8.9/WithdrawalQueue.sol#L399. This fn from smart contract returns a couple of return values.

src/modules/accounting.py
```diff
@@ -0,0 +1,251 @@
...
+        return self._w3.eth.get_balance(
+            contracts.lido.functions.getELRewardsVault().call(block_identifier=block_hash),
+            block_identifier=block_hash,
+        )
+
+    def _get_wei_to_reserve(self, block_hash: HexBytes) -> int:
+        return get_withdrawal_requests_wei_amount(block_hash)
+
+    def _get_requests_finalization_report(self, block_hash: HexBytes) -> Tuple[List[int], List[int]]:
+        last_finalized_request = contracts.withdrawal_queue.functions.finalizedQueueLength().call(block_identifier=block_hash)
```
sergeyWh1te, line 215: finalizedQueueLength - didn't find references in smart-contracts.

src/modules/accounting.py
```diff
@@ -0,0 +1,251 @@
...
+        ) = self._get_requests_finalization_report(block_hash)
+        logger.info({
+            'msg': 'Get withdrawals queue stats.',
+            'last_requests_id_to_finalize': last_requests_id_to_finalize,
+            'finalization_shares_amount': finalization_shares_amount,
+        })
+        if last_requests_id_to_finalize:
+            LAST_FINALIZED_WITHDRAWAL_REQUEST.set(last_requests_id_to_finalize[-1])
+
+        transaction = contracts.oracle.functions.handleCommitteeMemberReport((
```
sergeyWh1te, line 145: https://github.com/lidofinance/lido-dao/blob/feature/shapella-upgrade/contracts/0.8.9/LidoOracleNew.sol#L337

contracts.oracle.functions.handleCommitteeMemberReport has another signature.

src/modules/accounting.py
```diff
@@ -0,0 +1,251 @@
...
+            return True
+
+        quorum_size = contracts.oracle.functions.getQuorum().call(block_identifier=block_hash)
+        oracle_members = contracts.oracle.functions.getOracleMembers().call(block_identifier=block_hash)
+
+        try:
+            oracle_holder_index = oracle_members.index(ACCOUNT.address)
+        except ValueError:
+            logger.warning('Account is not oracle member.')
+            return True
```
sergeyWh1te, line 244: Mistype? 
I think it has to be "false"

src/modules/accounting.py
```diff
@@ -0,0 +1,251 @@
...
+
+        self._update_beacon_specs()
+
+    def _update_beacon_specs(self, block_identifier: Union[str, HexBytes] = 'latest'):
+        (
+            self.epochs_per_frame,
+            self.slots_per_epoch,
+            self.seconds_per_slot,
+            self.genesis_time,
+        ) = contracts.oracle.functions.getBeaconSpec().call(block_identifier=block_identifier)
```
avsetsin, line 51: https://github.com/lidofinance/lido-dao/blob/060f006f62ffebdb365a77891c9b2d12b3c7df42/contracts/0.8.9/oracle/HashConsensus.sol#L157-L168

Now `getChainConfig`, `getFrameConfig` are responsible for this. `getChainConfig` can only be called when starting the oracle, because these constants are specific to the network and cannot change over time.

src/modules/accounting.py
avsetsin: new interface: https://hackmd.io/@skozin/BJ7mDVuoo

src/modules/accounting.py
```diff
@@ -0,0 +1,251 @@
...
+            'slots_per_epoch': self.slots_per_epoch,
+            'seconds_per_slot': self.seconds_per_slot,
+            'genesis_time': self.genesis_time,
+        })
+
+    def run_module(self, slot: SlotNumber, block_hash: HexBytes):
+        """Check if epoch is reportable and try to send report if it is."""
+        self._update_beacon_specs(block_identifier=block_hash)
+
+        if not is_current_epoch_reportable(self._w3, contracts.oracle, slot, block_hash):
```
avsetsin, line 64: This check in the two-phase oracle will be more difficult. We must consider that the oracle must be able to change its report. This assumes an emergency situation if we have a bug in the consensus algorithm. In this case we are supposed to change the consensus algorithm in the oracle daemon and ask the committee members to update. 

So at startup the oracle should check if the current report is finalized, and if not, it should rebuild the report. If the hashes do not match what it sent before, send a new

src/modules/accounting.py
```diff
@@ -0,0 +1,251 @@
...
+        return (
+            stacking_module_id,
+            node_operator_id,
+            exited_validators_number,
+        )
+
+    def _get_exited_validators(self, slot: SlotNumber, block_hash: HexBytes) -> List[MergedLidoValidator]:
+        lido_validators = get_lido_validators(self._w3, block_hash, self._beacon_chain_client, slot)
+
+        withdrawable_validators = filter(lambda validator: int(validator['validator']['validator']['exit_epoch']) <= int(slot / self.slots_per_epoch), lido_validators)
```
avsetsin, line 201: We cannot rely on `exit_epoch` since a validator may still have eth on the balance
1) There may be validators with non-0x01 type WC
2) There is a time gap between `exit` and `withdrawable`

https://kb.beaconcha.in/ethereum-2.0-depositing#exited
https://notes.ethereum.org/7CFxjwMgQSWOHIxLgJP2Bw#4311-Initiate-exit

An example of a slashed and exited validator with eth:
http://unstable.mainnet.beacon-api.nimbus.team/eth/v1/beacon/states/head/validators/3280

```json
{
  "execution_optimistic": false,
  "data": {
    "index": "3280",
    "balance": "34118878082",
    "status": "withdrawal_possible",
    "validator": {
      "pubkey": "0xb3622c1180e86dcfe82e3e2a9d76a27614c002ba0dc9c058f8d6d5c75b95b5e834a30d9b84269ec8171419009c4cc5c3",
      "withdrawal_credentials": "0x00c4bc4e7b2e72415bf73c6af63cc13e4ebc98d1c87167d2a539771e38047381",
      "effective_balance": "32000000000",
      "slashed": true,
      "activation_eligibility_epoch": "0",
      "activation_epoch": "0",
      "exit_epoch": "148724",
      "withdrawable_epoch": "156911"
    }
  }
}
```

src/protocol_upgrade_checker.py
```diff
@@ -0,0 +1,39 @@
...
+from web3 import Web3
+
+from src import variables
+from src.contracts import contracts
+from src.metrics.healthcheck_server import pulse
+
+logger = logging.getLogger(__name__)
+
+
+def wait_for_withdrawals(w3: Web3):
```
avsetsin, line 13: We need to implement the contract and the consensus versions check anyway, after which this method will be unnecessary.

src/oracle.py
```diff
@@ -0,0 +1,135 @@
...
+            except Exception as error:
+                EXCEPTIONS_COUNT.labels(module.__class__.__name__).inc()
+                logger.error({'msg': f'Module {module.__class__.__name__} failed.', 'error': str(error)})
+
+
+if __name__ == '__main__':
+    logger.info({'msg': 'Oracle startup.'})
```
avsetsin, line 104: It would be great to log some env vars, consensus version at startup. It might make it easier to find issues in the logs

src/modules/accounting.py
TheDZhon: https://github.com/lidofinance/lido-dao/blob/feature/shapella-upgrade/contracts/0.8.9/WithdrawalQueue.sol#L220

All of the changes should be relevant to the `shapella-upgrade` branch.

pyproject.toml
```diff
@@ -1,24 +1,21 @@
...
 name = "lido-oracle"
-version = "2.4.0"
+version = "3.0.0"
 description = "Oracle daemon for Lido decentralized staking service. Collects and reports Ethereum 2.0 beacon chain states (the number of visible validators and their summarized balances) to the Lido dApp contract running on Ethereum 1.0 side."
-authors = ["Kirill Varlamov", "Smelov Vladimir", "Dmitry Suldin"]
+authors = ["Raman Siamionau", "Kirill Varlamov", "Smelov Vladimir", "Dmitry Suldin"]
 license = "GPL 3.0"
 
 [tool.poetry.dependencies]
 python = "^3.8,<4.0.0"
```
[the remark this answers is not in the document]
F4ever: It already freezed in poetry.lock

src/oracle.py
F4ever: This default python exception

src/services/extra_data.py
```diff
@@ -0,0 +1,129 @@
...
+                for operator_id, validators in chunk:
+                    operator_ids.append(operator_id.to_bytes(Lengths.NODE_OPERATOR_IDS))
+                    vals_count.append(len(validators).to_bytes(Lengths.STUCK_VALS_COUNTS))
+                payloads.append(ItemPayload(
+                    module_id=module.id.to_bytes(Lengths.MODULE_ID),
+                    node_ops_count=len(chunk).to_bytes(Lengths.NODE_OPS_COUNT),
+                    node_operator_ids=b"".join(operator_ids),
+                    vals_counts=b"".join(vals_count),
+                ))
+        return payloads
```
avsetsin, line 129: there is an extra check for amount of items in one group in the contract 

https://github.com/lidofinance/lido-dao/blob/feature/shapella-upgrade/contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol#L403-L423

https://discord.com/channels/761182643269795850/1029294901487214593/1075778491347382293

src/services/extra_data.py
```diff
@@ -0,0 +1,129 @@
...
+    def _get_oracle_report_limits(self):
+        output_names = get_function_output_names(self.w3.lido_contracts.oracleReportSanityChecker.abi, 'getOracleReportLimits')
+        result = self.w3.lido_contracts.oracleReportSanityChecker.functions.getOracleReportLimits().call()
+        return OracleReportLimits(**dict(zip(output_names, result)))
+
+    def to_bytes(self, extra_data: list[ExtraDataItem]) -> bytes:
+        extra_data_bytes = b''
+        for item in extra_data:
+            extra_data_bytes += item.item_index
+            extra_data_bytes += item.item_type.value.to_bytes(Lengths.ITEM_TYPE)
```
avsetsin, line 78: maybe it's better to move `.to_bytes(Lengths.ITEM_TYPE)` to `build_extra_data`

src/services/extra_data.py
```diff
@@ -0,0 +1,129 @@
...
+from src.web3_extentions.typings import Web3
+
+
+# Extra data is an array of items, each item being encoded as follows:
+# |  3 bytes  | 2 bytes  |   X bytes   |
+# | itemIndex | itemType | itemPayload |
+#
+# itemPayload format:
+#  | 3 bytes  |   8 bytes    |  nodeOpsCount * 8 bytes  |  nodeOpsCount * 16 bytes  |
+#  | moduleId | nodeOpsCount |      nodeOperatorIds     |      stuckValsCounts      |
```
avsetsin, line 19: stuck or exited

src/services/extra_data.py
```diff
@@ -0,0 +1,129 @@
...
+        for mapping in sorted(module_to_operators.values(), key=lambda x: x["module"].id):
+            module = mapping["module"]
+            operators = mapping["operators"]
+            operators = sorted(operators.items(), key=itemgetter(0))
+            for chunk in chunks(operators, max_list_items_count):
+                operator_ids = []
+                vals_count = []
+                for operator_id, validators in chunk:
+                    operator_ids.append(operator_id.to_bytes(Lengths.NODE_OPERATOR_IDS))
+                    vals_count.append(len(validators).to_bytes(Lengths.STUCK_VALS_COUNTS))
```
avsetsin, line 122: `STUCK_VALS_COUNTS` -> `STUCK_AND_EXITED_VALS_COUNTS` ?

src/web3_extentions/contracts.py
```diff
@@ -0,0 +1,62 @@
...
+
+        self.lido_execution_layer_rewards_vault = self.w3.eth.contract(
+            address=self.lido_locator.functions.elRewardsVault().call(),
+            abi=self.load_abi('LidoExecutionLayerRewardsVault'),
+        )
+
+        self.withdrawal_vault = self.w3.eth.contract(
+            address=self.lido_locator.functions.withdrawalVault().call(),
+            abi=self.load_abi('WithdrawalVault'),
+        )
```
avsetsin, line 38: i guess we don't need these contracts

src/web3_extentions/contracts.py
```diff
@@ -0,0 +1,62 @@
+import json
+
+from web3 import Web3
+from web3.module import Module
+
+from src import variables
+
+
+class LidoContracts(Module):
```
avsetsin, line 9: we need a OracleDaemonConfig contract as well

src/providers/keys/client.py
```diff
@@ -0,0 +1,51 @@
...
+    ALL_OPERATORS = 'v1/operators'
+
+    def _get_with_blockstamp(self, url: str, blockstamp: BlockStamp, params: Optional[dict] = None) -> dict:
+        """
+        Returns response if blockstamp < blockNumber from response
+        """
+        for i in range(self.RETRY_COUNT):
+            data, meta = self._get(url, params)
+            if meta['meta']['elBlockSnapshot']['blockNumber'] >= blockstamp.block_number:
+                return data
```
avsetsin, line 34: might be a good idea to check that total keys from Keys API >= total amount of keys that we fetches from staking router with `getAllNodeOperatorsReports` / `getAllNodeOperatorsDigests`

src/web3_extentions/lido_validators.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @lru_cache(maxsize=1)
+    def get_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
+        lido_keys = self.w3.kac.get_all_lido_keys(blockstamp)
+        validators = self.w3.cc.get_validators(blockstamp.state_root)
+
+        return self._merge_validators(lido_keys, validators)
+
+    @staticmethod
+    def _merge_validators(keys: list[LidoKey], validators: list[Validator]) -> list[LidoValidator]:
```
avsetsin, line 36: maybe `filter_validators_by_pubkeys`?

src/web3_extentions/lido_validators.py
```diff
@@ -0,0 +1,79 @@
...
+                lido_validators.append(LidoValidator(
+                    key=key,
+                    validator=validators_keys_dict[key.key],
+                ))
+
+        return lido_validators
+
+    @lru_cache(maxsize=1)
+    def get_lido_validators_by_node_operators(self, blockstamp: BlockStamp) -> Dict[NodeOperatorIndex, LidoValidator]:
+        merged_validators = self.get_lido_validators(blockstamp)
```
avsetsin, line 53: `merged_validators` -> `lido_validators`

src/web3_extentions/lido_validators.py
```diff
@@ -0,0 +1,79 @@
...
+        no_validators = {(operator.stakingModuleAddress, operator.index): [] for operator in no_operators}
+
+        for validator in merged_validators:
+            no_validators[(validator.key.moduleAddress, validator.key.operatorIndex)].append(validator)
+
+        return dict(no_validators)
+
+    @lru_cache(maxsize=1)
+    def get_lido_node_operators(self, blockstamp: BlockStamp) -> list[OperatorExpanded]:
+        operators_by_modules: list[OperatorResponse] = self.w3.kac.get_operators(blockstamp)
```
avsetsin, line 66: let's fetch them from the staking router

src/web3_extentions/tx_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+class TransactionUtils(Module):
+    @staticmethod
+    def check_transaction(transaction, from_address: str) -> bool:
+        """
+        Returns:
+        True - transaction succeed.
+        False - transaction reverted.
+        """
+        logger.info({"msg": "Check transaction locally.", "value": transaction.args})
```
avsetsin, line 23: not sure that it's "locally" since it makes a static call to the node. just a thought, not need to fix it if you mind

src/web3_extentions/tx_utils.py
```diff
@@ -0,0 +1,77 @@
...
+                "maxFeePerGas": pending_block.baseFeePerGas * 2 + self.w3.eth.max_priority_fee,
+                "maxPriorityFeePerGas": self.w3.eth.max_priority_fee,
+                "nonce": self.w3.eth.get_transaction_count(account.address),
+            }
+        )
+
+        signed_tx = self.w3.eth.account.sign_transaction(tx, account._private_key)
+
+        tx_hash = self.w3.eth.send_raw_transaction(signed_tx.rawTransaction)
+        TX_SEND.inc()
```
avsetsin, line 59: do we have a separate label sets for metrics for each oracle modules? if not may be we need to have them

src/web3_extentions/tx_utils.py
```diff
@@ -0,0 +1,77 @@
...
+        logger.info({"msg": "Transaction sent.", "value": tx_hash.hex()})
+
+        tx_receipt = self.w3.eth.wait_for_transaction_receipt(tx_hash)
```
avsetsin, line 62: what if the transaction will be in the queue for the long time?

src/modules/submodules/consensus.py
```diff
@@ -0,0 +1,350 @@
...
+    CONSENSUS_VERSION: int
+
+    def __init__(self, w3: Web3):
+        self.w3 = w3
+
+        if self.report_contract is None:
+            raise NotImplementedError('report_contract attribute should be set.')
+
+        if self.CONSENSUS_VERSION is None or self.CONSENSUS_VERSION is None:
+            raise NotImplementedError('CONSENSUS_VERSION and CONSENSUS_VERSION should be set.')
```
avsetsin, line 40:
```suggestion
        if self.CONTRACT_VERSION is None or self.CONSENSUS_VERSION is None:
            raise NotImplementedError('CONTRACT_VERSION and CONSENSUS_VERSION should be set.')
```

assets/IStakingModule.json
```diff
@@ -0,0 +1 @@
...idatorsKeysForDeposits","outputs":[{"internalType":"uint256","name":"returnedKeysCount","type":"uint256"},{"internalType":"bytes","name":"publicKeys","type":"bytes"},{"internalType":"bytes","name":"signatures","type":"bytes"}],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"_nodeOperatorId","type":"uint256"},{"internalType":"uint256","name":"_exitedValidatorsKeysCount","type":"uint256"},{"internalType":"uint256","name":"_stuckValidatorsKeysCount","type":"uint256"}],"name":"unsafeUpdateValidatorsKeysCount","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"_nodeOperatorId","type":"uint256"},{"internalType":"uint256","name":"_exitedValidatorKeysCount","type":"uint256"}],"name":"updateExitedValidatorsKeysCount","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"_nodeOperatorId","type":"uint256"},{"internalType":"uint256","name":"_stuckValidatorKeysCount","type":"uint256"}],"name":"updateStuckValidatorsKeysCount","outputs":[],"stateMutability":"nonpayable","type":"function"}]
```
avsetsin, line 1: it doesn't look like we use it anywhere

src/modules/ejector/data_encode.py
```diff
@@ -0,0 +1,32 @@
...
+    #     /// |  moduleId  |  nodeOpId  |  validatorIndex  | validatorPubkey |
+
+    result = b''
+
+    for (module_id, op_id), validator in validators:
+        result += module_id.to_bytes(MODULE_ID_LENGTH)
+        result += op_id.to_bytes(NODE_OPERATOR_ID_LENGTH)
+        result += int(validator.index).to_bytes(VALIDATOR_INDEX_LENGTH)
+
+        if len(validator.validator.pubkey) != VALIDATOR_PUB_KEY_LENGTH:
```
avsetsin, line 27: according to `pubkey` type here it's a string, so its length will be = 98 
2 ('0x') + 48 bytes * 2 = 98

src/modules/ejector/data_encode.py
```diff
@@ -0,0 +1,32 @@
+from eth_typing import HexStr
+
+from src.utils.types import hex_str_to_bytes
+from src.web3py.extentions.lido_validators import NodeOperatorGlobalIndex, LidoValidator
+
+
+DATA_FORMAT_LIST = 1
```
avsetsin, line 7: might be 0 for current deployed contracts

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,211 @@
...
+        validators_to_eject = []
+        validator_to_eject_balance_sum = 0
+
+        validators_iterator = ValidatorToExitIterator(
+            w3=self.w3,
+            blockstamp=blockstamp,
+            c_conf=chain_config,
+        )
+
+        for validator in validators_iterator:
```
avsetsin, line 94: need to limit validators with `maxValidatorExitRequestsPerReport` from `OracleReportSanityChecker - getOracleReportLimits`

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,211 @@
...
+            withdrawal_epoch = self._get_epoch_validators_will_be_withdrawn(blockstamp, len(validators_to_eject) + 1)
+            future_rewards = (blockstamp.ref_epoch - (withdrawal_epoch + epochs_to_sweep)) * rewards_speed_per_epoch
```
avsetsin, line 96:
```suggestion
            future_rewards = ((withdrawal_epoch + epochs_to_sweep) - blockstamp.ref_epoch) * rewards_speed_per_epoch
```

avsetsin: and we need to handle the case when `rewards_speed_per_epoch` < 0

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,211 @@
...
+            self.w3.lido_contracts.lido.functions.getBufferedEther().call(
+                block_identifier=blockstamp.block_hash
+            )
+        )
+
+    def get_total_unfinalized_withdrawal_requests_amount(self, blockstamp: BlockStamp) -> Wei:
+        steth_to_finalize = self.w3.lido_contracts.withdrawal_queue_nft.functions.unfinalizedStETH().call(
+            block_identifier=blockstamp.block_hash,
+        )
+        logger.info({'msg': 'Wei to finalize.'})
```
avsetsin, line 141: add value to the log

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,211 @@
...
+            w3=self.w3,
+            blockstamp=blockstamp,
+            c_conf=chain_config,
+        )
+
+        for validator in validators_iterator:
+            withdrawal_epoch = self._get_epoch_validators_will_be_withdrawn(blockstamp, len(validators_to_eject) + 1)
+            future_rewards = (blockstamp.ref_epoch - (withdrawal_epoch + epochs_to_sweep)) * rewards_speed_per_epoch
+
+            if future_rewards + total_current_balance + validator_to_eject_balance_sum >= to_withdraw_amount:
```
avsetsin, line 98: We do not take into account the balance of validators who previously requested to exit and validators that are now in the exit queue

avsetsin: described here (get_inflight_validators_eth): https://hackmd.io/PbUr-orDTAqH63F4-n8-pw?view#Predict-available-ETH-before-next-withdrawn

src/utils/slot.py
```diff
@@ -0,0 +1,138 @@
...
+            _check_block_header(existed_header)
+            break
+
+    if not existed_header:
+        raise NoSlotsAvailable('No slots available for current report. Check your CL node.')
+
+    if ref_slot_is_missed:
+        # Ref slot is missed, and we have next non-missed slot.
+        # We should get parent root of this non-missed slot
+        # and get details of its parent slot until we found slot < ref_slot.
```
avsetsin, line 96: this line is outdated

src/utils/slot.py
```diff
@@ -0,0 +1,138 @@
...
+class SlotNotFinalized(Exception):
+    pass
+
+
+def get_first_non_missed_slot(
+    cc: ConsensusClient,
+    ref_slot: SlotNumber,
+    last_finalized_slot_number: SlotNumber,
+    ref_epoch: Optional[EpochNumber] = None,
+) -> BlockStamp:
```
avsetsin, line 31: We use this not only with `ref_slot`, in bunker for example. And since this method returns blockstamp with ref_slot and ref_epoch this can cause issues.

I would suggest splitting it into two methods. One finds the closest block to the slot, the other expands blockstamp with additional fields: ref_slot and ref_epoch

src/modules/accounting/accounting.py
```diff
@@ -0,0 +1,222 @@
...
+
+    # # Consensus module: if contract got report data
+    def is_main_data_submitted(self, blockstamp: BlockStamp) -> bool:
+        processing_state = self._get_processing_state(blockstamp)
+        logger.info({'msg': 'Check if main data was submitted.', 'value': processing_state.main_data_submitted})
+        return processing_state.main_data_submitted
+
+    # Consensus module: if contract could accept any sort of report
+    def is_contract_reportable(self, blockstamp: BlockStamp) -> bool:
+        is_reportable = not self.is_main_data_submitted(blockstamp) or not self.is_extra_data_submitted(blockstamp)
```
avsetsin, line 76: **last_finalized_blockstamp** -> `execute_module` -> `get_blockstamp_for_report` -> `is_contract_reportable`

`is_contract_reportable` is called from `get_blockstamp_for_report` that is called with `last_finalized_blockstamp` in `execute_module`

in turn `is_contract_reportable` calls `is_main_data_submitted` and `is_extra_data_submitted` with `last_finalized_blockstamp` but i guess it should perform these checks on the latest block

src/modules/accounting/accounting.py
```diff
@@ -0,0 +1,222 @@
...
+        ps = named_tuple_to_dataclass(
+            self.report_contract.functions.getProcessingState().call(block_identifier=blockstamp.block_hash),
+            ProcessingState,
+        )
+        logger.info({'msg': 'Fetch processing state.', 'value': ps})
+        return ps
+
+    def is_extra_data_submitted(self, blockstamp: BlockStamp) -> bool:
+        processing_state = self._get_processing_state(blockstamp)
+        return processing_state.extra_data_items_count == processing_state.extra_data_items_submitted
```
avsetsin, line 91: TODO: we should check it in a separate method after new deploy. Since the third phase will distribute rewards, and we need to send tx even if no extra data

src/services/safe_border.py
```diff
@@ -0,0 +1,268 @@
...
+    def _get_withdrawal_request_status(self, request_id: int) -> any:
+        return self.w3.lido_contracts.withdrawal_queue_nft.functions.getWithdrawalRequestStatus(request_id).call(
+            block_identifier=self.blockstamp.block_hash)
+
+    def _get_last_successful_report_slot(self) -> SlotNumber:
+        return self.w3.lido_contracts.accounting_oracle.functions.getLastProcessingRefSlot().call(
+            block_identifier=self.blockstamp.block_hash)
+
+    def _retrieve_constants(self):
+        # Check self.w3.to_int works
```
avsetsin, line 215: TODO

src/modules/accounting/accounting.py
```diff
@@ -0,0 +1,222 @@
...
+        last_withdrawal_id_to_finalize = self._get_last_withdrawal_request_to_finalize(blockstamp)
+
+        exited_validators = self.lido_validator_state_service.get_lido_newly_exited_validators(blockstamp)
+
+        # Here report all exited validators even they were reported before.
+        if exited_validators:
+            node_operator_global_ids_list, exit_validators_count_list = zip(*exited_validators.items())
+            staking_module_ids_list = map(lambda no_global_id: no_global_id[0], node_operator_global_ids_list)
+        else:
+            staking_module_ids_list = exit_validators_count_list = []
```
avsetsin, line 106: 1) we can't relay on `exited_validators` here since `get_lido_newly_exited_validators` returns only changed values, but we need to report total exited validators by module
2) `exit_validators_count_list` should be a map with total exited keys by modules

Let's look at an example:

Current contract state:
Module 1: total exited keys: 10
Module 2: total exited keys: 15
Module 3: total exited keys: 20

CL state: 
Module 1: total exited keys: 12
Module 2: total exited keys: 15
Module 3: total exited keys: 21

2 newly exited validators appears in `Module 1` (10->12), and 1 in `Module 3` (20 -> 21)

This means we have to report:
`stacking_module_id_with_exited_validators=[1,3]`
`count_exited_validators_by_stacking_module=[12,21]`

We do not report Module 2 since there was no change

src/services/safe_border.py
```diff
@@ -0,0 +1,268 @@
...
+                SlotNumber(mid_slot),
+                self.blockstamp.slot_number
+            )
+            validators = self.w3.lido_validators.get_lido_validators(mid_non_missed_blockstamp)
+            slashed_validators = filter_slashed_validators(validators)
+
+            if slashed_validators:
+                end_slot = mid_non_missed_blockstamp.slot_number
+            else:
+                start_slot = mid_non_missed_blockstamp.slot_number + 1
```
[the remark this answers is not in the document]
rkolpakov: Here we check if these slots are close than one epoch and if so, we stop a loop
https://github.com/lidofinance/lido-oracle/blob/fe98b3e79a1c736b3395c01919e12a943f263358/src/services/safe_border.py#L158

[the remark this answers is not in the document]
avsetsin: It's still fragile, there could be a situation where an entire epoch is missed.

And if we know for sure that the slots in the interval `(mid_non_missed_blockstamp.slot_number, mid_slot]` are missed, it makes no sense to set `start_slot = mid_non_missed_blockstamp.slot_number + 1`, it's more correct to set `start_slot = mid_slot + 1` instead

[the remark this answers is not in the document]
avsetsin: `in_one_epoch = end_slot - start_slot <= self.chain_config.slots_per_epoch` is also not correct 

```
  |<-- epoch -->|
--|-------------|-------------|-------------|--->
            ^      ^
          slot1  slot2
```

`slot2 - slot1 <= slots_per_epoch` but they are not "in_one_epoch"

PR #176, opened 2023-02-08
Feat/lido validators

F4ever:


src/providers/http_provider.py
```diff
@@ -42,13 +47,18 @@ def _get(self, url: str, params: Optional[dict] = None) -> Tuple[dict | list, di
                 timeout=self.REQUEST_TIMEOUT,
             )
 
+        if response.status_code != HTTPStatus.OK.value:
```
skhomuti, line 50: Direct checking with enum also works
```suggestion
        if response.status_code != HTTPStatus.OK:
```

tests/conftest.py
```diff
@@ -5,16 +5,18 @@
 from web3 import Web3
```
skhomuti, line 5: For better typing
```suggestion
from src.web3_extentions.typings import Web3
```

skhomuti: And also, `Web3` doesn't have `lido_validators` attribute. Could you add it?

src/web3_extentions/lido_validators.py
```diff
@@ -33,18 +36,39 @@ def _merge_validators(keys: List[LidoKey], validators: List[Validator]) -> List[
...
+
+        operators = []
+
+        for module in operators_by_modules:
+            operators.extend([
+                {'stakingModuleAddress': module['module']['stakingModuleAddress'], **operator}
+                for operator in module['operators']
+            ])
+
+        return operators
```
skhomuti, line 74:
```suggestion
    def get_lido_node_operators(self, blockstamp: BlockStamp) -> List[OperatorExpanded]:
        operators_by_modules: list[OperatorResponse] = self.w3.kac.get_operators(blockstamp)

        operators = []

        for module in operators_by_modules:
            operators.extend([
                OperatorExpanded(stakingModuleAddress=module['module']['stakingModuleAddress'], **operator)
                for operator in module['operators']
            ])

        return operators
```

skhomuti (changes requested):

tests/conftest.py
F4ever: Fixed

skhomuti (approved):

PR #180, opened 2023-02-10
Fix minor typing errors

skhomuti:
- fixed enum values checks. Now it uses an iterator instead of private attributes
- Fixed confusion with block root and state root
- Fixed w3 typing by avoiding circular imports

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -25,8 +25,8 @@ logger = logging.getLogger()
 
 if __name__ == '__main__':
     module_name = sys.argv[-1]
-    if module_name not in OracleModule._value2member_map_:
-        msg = f'Last arg should be one of {list(OracleModule._value2member_map_.keys())}, received {module_name}.'
+    if module_name not in iter(OracleModule):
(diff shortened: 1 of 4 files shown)


src/web3_extentions/tx_utils.py
```diff
@@ -47,7 +47,7 @@ def sign_and_send_transaction(
             {
                 "from": account.address,
                 "gas": gas_limit,
-                "maxFeePerGas": pending_block.baseFeePerGas * 2 + self.w3.eth.max_priority_fee,
```
F4ever, line 50: image

F4ever (approved):

src/web3_extentions/tx_utils.py
skhomuti: yea, I also doubt about it
there's a typing warning because returned type is a TypedDict
but actually, it's AttributedDict

skhomuti: reverted

PR #183, opened 2023-02-10
feat: calc reward prediction

sergeyWh1te:


src/variables.py
```diff
@@ -28,3 +28,10 @@
 HEALTHCHECK_SERVER_PORT = int(os.getenv('HEALTHCHECK_SERVER_PORT', 9010))
 
 MAX_CYCLE_LIFETIME_IN_SECONDS = int(os.getenv("MAX_CYCLE_LIFETIME_IN_SECONDS", 300))
+
+# Oracle constants
+# REWARDS_HISTORICAL_PERIOD default - a one week
+REWARDS_HISTORICAL_PERIOD_IN_SECONDS = int(604800)
```
F4ever, line 34: Lets move it to constants.py

pyproject.toml
```diff
@@ -11,6 +11,7 @@ web3 = "6.0.0b9"
 prometheus-client = "^0.16.0"
 web3-multi-provider = "^0.5.0"
 timeout-decorator = "^0.5.0"
+pytest-mock = "^3.10.0"
```
skhomuti, line 14: looks like unused lib

src/variables.py
sergeyWh1te: Sure, but I didn't have constants.py in my branch =). Of course I will

tests/test_prediction.py
```diff
@@ -0,0 +1,149 @@
...
+    eth_event = MagicMock()
+    eth_event.build_filter = MagicMock(return_value=eht_contract_event)
+
+    p = Prediction(eth_event)
+
+    report_timestamp = BlockStamp(block_number=100000)
+
+    got = p.get_rewards_per_epoch(report_timestamp)
+
+    assert type(got) is type(expected) and got.args == expected.args
```
skhomuti, line 149: Split it to get a more obvious error message

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,94 @@
...
+        # Rough estimate, since there could be missing blocks in the slots
+        # 604800 / 12 = 50400
+        max_period_block_range = REWARDS_HISTORICAL_PERIOD_IN_SECONDS / SECONDS_PER_SLOT
+
+        from_block = ref_block - max_period_block_range
+        to_block = ref_block
+        return int(from_block), to_block
+
+    @staticmethod
+    def median(rewards: List[Wei]) -> Wei:
```
skhomuti, line 85: probably it would be suitable
```
import statistics
statistics.median()
```

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,94 @@
...
+        from_block, to_block = self.get_block_for_filtering(report_timestamp['block_number'])
+
+        logs: Iterable[EventData]
+        try:
+            logs = self.contract_event.build_filter().get_logs(
+                from_block=from_block,
+                to_block=to_block,
+            )
+        except Exception as e:
+            return e
```
skhomuti, line 62: it doesn't look like a "pythonic" way, better just throw it and catch on the top-levels

skhomuti (changes requested):

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,94 @@
...
+        :param contract_event: ETHDistributed(
+                uint256 indexed reportTimestamp,
+                uint256 preCLBalance,
+                uint256 postCLBalance,
+                uint256 withdrawalsWithdrawn
+                uint256 executionLayerRewardsWithdrawn
+                uint256 postBufferedEther
+            );
+        """
+        self.contract_event = contract_event
```
F4ever, line 20: Why contract_event is here?
Replace it with web3 instance.

F4ever, line 20: Why contract_event is here?
Replace it with web3 instance.

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def get_rewards_per_epoch(self, report_timestamp: BlockStamp) -> Wei | Exception:
```
F4ever, line 22: It's not timestamp, it's blockstamp

F4ever, line 22: It's not timestamp, it's blockstamp

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            logs = self.contract_event.build_filter().get_logs(
+                from_block=from_block,
+                to_block=to_block,
+            )
+        except Exception as e:
+            return e
+
+        rewards: List[Wei] = list()
+        for log in logs:
```
F4ever, line 65: Is this approved by Analytics team?

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,61 @@
...
+from web3.contract import ContractEvent
+from web3.types import Wei, EventData
+
+from modules.accounting.constants import REWARDS_HISTORICAL_PERIOD_IN_SECONDS, SECONDS_PER_SLOT
+from src.typings import BlockStamp
+
+
+class Prediction:
+    _eth_distributed: ContractEvent
+    _eth_distributed_address: web3.eth.Address = 'address'
```
sergeyWh1te, line 14: @F4ever 
I found an example in depositBot. 
https://github.com/lidofinance/depositor-bot/blob/main/scripts/utils/interfaces.py, that address and abi - is a GlobalConstant. This why it'll be change when constants come like in an examlpe

src/modules/accounting/prediction.py
sergeyWh1te: dealed

src/modules/accounting/prediction.py
sergeyWh1te: @F4ever told that it's ok to use a small custom median function(cause it's pretty simple) instead of using additional lib.
Ps: the first variant was with using the external method from the library. Honestly, it was provided by "numpy".

src/modules/accounting/constants.py
```diff
@@ -1,2 +1,9 @@
 CONTRACT_VERSION = 1
 CONSENSUS_VERSION = 1
+
+# Oracle constants
+# REWARDS_HISTORICAL_PERIOD default - a one week
+REWARDS_HISTORICAL_PERIOD_IN_SECONDS = int(604800)
+
+# Blockchain
+SECONDS_PER_SLOT = int(12)
```
F4ever, line 9: - SECONDS_PER_SLOT - you can fetch from contract

src/modules/accounting/constants.py
```diff
@@ -1,2 +1,9 @@
 CONTRACT_VERSION = 1
 CONSENSUS_VERSION = 1
+
+# Oracle constants
+# REWARDS_HISTORICAL_PERIOD default - a one week
+REWARDS_HISTORICAL_PERIOD_IN_SECONDS = int(604800)
```
F4ever, line 6: I hope this value also will be in OracleConfigs

F4ever: But for now lets place it directly in prediction.py

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,61 @@
...
+            rewards.append(Wei(reward))
+
+        median: Wei = self.median(rewards)
+        return Wei(median)
+
+    @staticmethod
+    def get_block_for_filtering(ref_block: int) -> (int, int):
+        # Rough estimate, since there could be missing blocks in the slots
+        # 604800 / 12 = 50400
+        max_period_block_range = REWARDS_HISTORICAL_PERIOD_IN_SECONDS / SECONDS_PER_SLOT
```
F4ever, line 45: Why this constant in seconds, not in Slots?

F4ever: Or epochs

src/modules/accounting/prediction.py
F4ever: If this built-in lib why not? Numpy is not included in base python packages.
Let's write less code and use more available bicycles.

F4ever: And if you use lib, you don't have to test it

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,61 @@
...
+            address=self._eth_distributed_address,
+            abi=self._eth_distributed_abi
+        ).events.ETHDistributed
+
+    def get_rewards_per_epoch(self, blockstamp: BlockStamp) -> Wei:
+        from_block, to_block = self.get_block_for_filtering(blockstamp['block_number'])
+
+        logs: Iterable[EventData]
+
+        logs = self._eth_distributed.build_filter().get_logs(
```
F4ever, line 28: What if there were no events? I think we should report 0 in that case

src/variables.py
```diff
@@ -27,4 +27,4 @@
 
 HEALTHCHECK_SERVER_PORT = int(os.getenv('HEALTHCHECK_SERVER_PORT', 9010))
 
-MAX_CYCLE_LIFETIME_IN_SECONDS = int(os.getenv("MAX_CYCLE_LIFETIME_IN_SECONDS", 300))
+MAX_CYCLE_LIFETIME_IN_SECONDS = int(os.getenv("MAX_CYCLE_LIFETIME_IN_SECONDS", 300))
```
F4ever, line 30: eof broken

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,61 @@
...
+            from_block=from_block,
+            to_block=to_block,
+        )
+
+        rewards: List[Wei] = list()
+        for log in logs:
+            reward = log['args']['withdrawalsWithdrawn'] + log['args']['executionLayerRewardsWithdrawn']
```
F4ever, line 35: If report skipped, this won't work, because one of the events will have twice of rewards.
Probably next should work:
- get all events
- calculate for each event eth_per_epoch
- And then median([...medians_per_epochs])

F4ever: @avsetsin add pls name of the event where report_perioud is

src/modules/accounting/prediction.py
skhomuti: yea, statistics is a built-in lib

src/modules/accounting/prediction.py
```diff
@@ -0,0 +1,61 @@
...
+from src.typings import BlockStamp
+
+
+class Prediction:
+    _eth_distributed: ContractEvent
+    _eth_distributed_address: web3.eth.Address = 'address'
+    _eth_distributed_abi = 'abi'
+
+    def __init__(self, w3: Web3):
+        self._eth_distributed = w3.eth.contract(
```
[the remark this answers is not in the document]
sergeyWh1te: Then it could be left as it was initially. Because the class depends only on the event and not the whole web3.

src/modules/accounting/prediction.py
sergeyWh1te: Didn't get you. Let's discuss it

src/modules/accounting/prediction.py
sergeyWh1te: Discussed here https://discord.com/channels/761182643269795850/1072446086423257088/1073220253473050767

src/modules/accounting/prediction.py
sergeyWh1te: done

src/modules/accounting/prediction.py
F4ever: TokenRebased rebased

PR #185, opened 2023-02-11
feat: bunker mode

vgorkavenko:


src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+            )
+            for v in lido_validators_slashed_in_epoch:
+                effective_balance = int(v.validator.effective_balance)
+                penalty_numerator = effective_balance // EFFECTIVE_BALANCE_INCREMENT * adjusted_total_slashing_balance
+                penalty = penalty_numerator // total_balance * EFFECTIVE_BALANCE_INCREMENT
+                per_epoch_lido_midterm_penalties[epoch] = per_epoch_lido_midterm_penalties.get(epoch, 0) + penalty
+
+        # Put per epoch buckets into per frame buckets to calculate lido midterm penalties impact in each frame
+        per_frame_buckets: dict[int, int] = {}
+        index = 0
```
skhomuti, line 136: move it nearly to the loop

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+                total_balance
+            )
+            for v in lido_validators_slashed_in_epoch:
+                effective_balance = int(v.validator.effective_balance)
+                penalty_numerator = effective_balance // EFFECTIVE_BALANCE_INCREMENT * adjusted_total_slashing_balance
+                penalty = penalty_numerator // total_balance * EFFECTIVE_BALANCE_INCREMENT
+                per_epoch_lido_midterm_penalties[epoch] = per_epoch_lido_midterm_penalties.get(epoch, 0) + penalty
+
+        # Put per epoch buckets into per frame buckets to calculate lido midterm penalties impact in each frame
+        per_frame_buckets: dict[int, int] = {}
```
skhomuti, line 135: it also might be `defaultdict(int)`

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+
+MIN_VALIDATOR_WITHDRAWABILITY_DELAY = 256
+EPOCHS_PER_SLASHINGS_VECTOR = 8192
+PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX = 3
+EFFECTIVE_BALANCE_INCREMENT = 2 ** 0 * 10 ** 9
+MIN_DEPOSIT_AMOUNT = 32 * 10 ** 9
+
+FIRST_DEPOSIT_EVENT_LOOKUP_MAXIMUM_BLOCKS_DISTANCE = 225 * 7
+
+NORMILIZED_CL_PER_EPOCH = 64
```
skhomuti, line 20: `NORMALIZED` typo

skhomuti (changes requested):

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+        )
+
+        cl_rebase = ref_lido_balance_with_vault - corrected_prev_lido_balance_with_vault
+
+        logger.info({"msg": f"CL rebase from {prev_epoch=}: {cl_rebase} Gwei"})
+
+        return cl_rebase
+
+    def _calculate_lido_balance_with_vault(
+        self, ref_blockstamp: BlockStamp, lido_validators: List[Validator], vault_balance_at: BlockStamp
```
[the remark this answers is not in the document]
vgorkavenko: The vault address should be received only once somewhere. Then we don't need to pass `ref_blockstamp`

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+            "_clBalance": self.w3.to_wei(ref_lido_balance, 'gwei'),
+            "_withdrawalVaultBalance": ref_withdrawal_vault_balance,
+            "_elRewardsVaultBalance": 0,
+            "_lastFinalizableRequestId": 0,
+            "_simulatedShareRate": 0,
+        }
+        before_report_total_pooled_ether = self.w3.lido_contracts.lido.functions.totalSupply().call(
+            block_identifier=cdtp.ref_blockstamp.block_hash
+        )
+        after_report_total_pooled_ether, _ = self.w3.lido_contracts.lido.functions.handleOracleReport(**args).call(
```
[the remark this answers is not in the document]
vgorkavenko: I thought the current version is easier to read

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+
+        return frame_cl_rebase
+
+    def _is_high_midterm_slashing_penalty(self, cdtp: CommonDataToProcess, cl_rebase: Gwei) -> bool:
+        logger.info({"msg": "Detecting high midterm slashing penalty"})
+        all_slashed_validators = self._not_withdrawn_slashed_validators(cdtp.ref_all_validators, cdtp.ref_epoch)
+        lido_slashed_validators: List[Validator] = [
+            v.validator for v in self.w3.lido_validators.filter_lido_validators(cdtp.lido_keys, all_slashed_validators)
+        ]
+        logger.info({"msg": f"Slashed - all: {len(all_slashed_validators)} lido: {len(lido_slashed_validators)}"})
```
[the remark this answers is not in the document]
vgorkavenko: `total validators` are often used in the context of all validators of all statuses, not just slashed. Mb I will write something with {all_slashed_validators, lido_slashed_validators=}

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+
+        # If no one lido in current slashed validators - no need to bunker
+        if len(lido_slashed_validators) == 0:
+            return False
+
+        # Fill per epoch buckets by possible slashed epochs
+        per_epoch_buckets: dict[int, List[Validator]] = {}
```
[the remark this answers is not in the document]
vgorkavenko: naming is controversial, I agree, but in this step of reading it is more important to us that this "something" per epoch, because it's just an intermediate state of data. I expand a code comment above

src/services/bunker.py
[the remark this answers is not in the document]
skhomuti: in this case just use it from `lido_contracts` module

src/services/bunker.py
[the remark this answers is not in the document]
skhomuti: sounds reasonable. So just put a separator between these values, yea

src/utils/dataclass.py
```diff
@@ -15,8 +15,10 @@ def __post_init__(self):
                 field_type = field.type.__args__[0]
                 if is_dataclass(field_type):
                     setattr(self, field.name,
-                            field.type.__origin__(map(lambda x: field_type(**x), getattr(self, field.name))))
-            elif is_dataclass(field.type):
+                            field.type.__origin__(map(
+                                lambda x: field_type(**x) if not is_dataclass(x) else x,
+                                getattr(self, field.name))))
+            elif is_dataclass(field.type) and not is_dataclass(getattr(self, field.name)):
```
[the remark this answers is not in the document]
vgorkavenko: Merged

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+
+        return cl_rebase
+
+    def _calculate_lido_balance_with_vault(
+        self, ref_blockstamp: BlockStamp, lido_validators: List[Validator], vault_balance_at: BlockStamp
+    ) -> tuple[Gwei, Gwei]:
+        """
+        Calculate Lido validators balance and balance in withdrawal vault contract in particular block
+        """
+        lido_balance = sum([int(validator.balance) for validator in lido_validators])
```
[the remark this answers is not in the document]
vgorkavenko: This function doesn't do two actions. It returns the balance amount with rewards which go to vault. This function is needed to decrease the same lines with the same actions in the main functions

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
-from src.web3_extentions.typings import Web3
+import math
+import logging
+from typing import List, Sequence
+
+from ens.utils import Web3
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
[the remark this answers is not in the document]
vgorkavenko: done

tests/services/test_bunker.py
```diff
@@ -0,0 +1,72 @@
...
+
+from src.providers.consensus.typings import Validator, ValidatorStatus, ValidatorState
+from src.services.bunker import BunkerService
+
+
+# Static functions
+
+test_data_calculate_total_effective_balance = [
+    (
+        [Validator('0', '1', ValidatorStatus.ACTIVE_ONGOING.value, ValidatorState('', '', '2', False, '', '', '', '')),
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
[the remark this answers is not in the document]
vgorkavenko: done

src/services/bunker.py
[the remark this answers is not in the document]
vgorkavenko: done

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+        # If no one lido in current slashed validators - no need to bunker
+        if len(lido_slashed_validators) == 0:
+            return False
+
+        # Fill per epoch buckets by possible slashed epochs
+        per_epoch_buckets: dict[int, List[Validator]] = {}
+        for validator in all_slashed_validators:
+            possible_epochs_with_slashings = self._detect_slashing_epoch_range(validator, cdtp.ref_epoch)
+            for epoch in possible_epochs_with_slashings:
+                per_epoch_buckets[epoch] = per_epoch_buckets.get(epoch, []) + [validator]
```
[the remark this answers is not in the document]
vgorkavenko: done

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+        """
+        Calculates total balance of all active validators in network
+        """
+        total_effective_balance = 0
+
+        for validator in all_validators:
+            if validator.status in [
+                ValidatorStatus.ACTIVE_ONGOING.value,
+                ValidatorStatus.ACTIVE_EXITING.value,
+                ValidatorStatus.ACTIVE_SLASHED.value
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+            ]:
+                validator_effective_balance = validator.validator.effective_balance
+                total_effective_balance += int(validator_effective_balance)
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+    def _is_high_midterm_slashing_penalty(self, cdtp: CommonDataToProcess, cl_rebase: Gwei) -> bool:
+        logger.info({"msg": "Detecting high midterm slashing penalty"})
+        all_slashed_validators = self._not_withdrawn_slashed_validators(cdtp.ref_all_validators, cdtp.ref_epoch)
+        lido_slashed_validators: List[Validator] = [
+            v.validator for v in self.w3.lido_validators.filter_lido_validators(cdtp.lido_keys, all_slashed_validators)
+        ]
+        logger.info({"msg": f"Slashed - all: {len(all_slashed_validators)} lido: {len(lido_slashed_validators)}"})
+
+        # If no one lido in current slashed validators - no need to bunker
+        if len(lido_slashed_validators) == 0:
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+            slashed_keys = [k.validator.pubkey for k in slashed_validators]
+            lido_validators_slashed_in_epoch = [
+                v for v in lido_slashed_validators if v.validator.pubkey in slashed_keys
+            ]
+            if len(lido_validators_slashed_in_epoch) == 0:
+                continue
+            # We should calculate penalties according to bounded slashings in past EPOCHS_PER_SLASHINGS_VECTOR
+            bounded_slashings_balance_sum = 0  # aka state.slashings from spec
+            min_bounded_epoch = max(
+                min(per_epoch_buckets.keys()), epoch - EPOCHS_PER_SLASHINGS_VECTOR
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -1,15 +1,350 @@
...
+            )
+            bounded_epochs = [e for e in per_epoch_buckets if min_bounded_epoch <= e <= epoch]
+            for e in bounded_epochs:
+                bounded_slashings_balance_sum += sum(
+                    [int(v.validator.effective_balance) for v in per_epoch_buckets[e]]
+                )
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -258,21 +258,17 @@ def _calculate_cl_rebase_from(self, cdtp: CommonDataToProcess, prev_blockstamp:
         return cl_rebase
 
     def _calculate_lido_balance_with_vault(
-        self, ref_blockstamp: BlockStamp, lido_validators: List[Validator], vault_balance_at: BlockStamp
+        self, lido_validators: List[LidoValidator], vault_balance_at: BlockStamp
```
skhomuti, line 261: preferably to use a built-in typing for primitives - `list[LidoValidator]`

src/services/bunker.py
```diff
@@ -129,26 +129,26 @@ def _is_high_midterm_slashing_penalty(self, cdtp: CommonDataToProcess, cl_rebase
...
         # Put per epoch buckets into per frame buckets to calculate lido midterm penalties impact in each frame
-        per_frame_buckets: dict[int, int] = {}
-        index = 0
+        per_frame_buckets = defaultdict(int)
         right_border = self.f_conf.initial_epoch + self.f_conf.epochs_per_frame
         # The first bucket should have flexible size. It can be less than 'epochs_per_frame'
         while right_border < min(per_epoch_lido_midterm_penalties.keys()):
             right_border += self.f_conf.epochs_per_frame
-        left_border = min(per_epoch_lido_midterm_penalties.keys())
+        left_border = min_bucket_epoch
```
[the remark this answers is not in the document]
vgorkavenko: Yes, it's an error. Should be `min(per_epoch_lido_midterm_penalties.keys())`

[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
vgorkavenko: done

src/modules/accounting/accounting.py
```diff
@@ -11,6 +13,8 @@
 class Accounting(BaseModule, ConsensusModule):
     def __init__(self, w3: Web3):
         self.report_contract = w3.lido_contracts.accounting_oracle
+        ConsensusModule.report_contract = self.report_contract
+        BunkerService.report_contract = self.report_contract
```
F4ever, line 17: Set global attributes?
If you want to set this attributes, it's better to just pass them into __init__

src/modules/accounting/accounting.py
```diff
@@ -28,4 +32,25 @@ def execute_module(self, blockstamp: BlockStamp):
...
     def build_report(self, blockstamp: BlockStamp):
-        pass
+        member_info = self.get_member_info(blockstamp)
+        frame_config = self.get_frame_config(blockstamp)
+        chain_config = self.get_chain_config(blockstamp)
+
+        all_validators, lido_keys, lido_validators = self.w3.lido_validators.get_lido_validators_with_others(blockstamp)
+        logger.info({"msg": f"Validators - all: {len(all_validators)} lido: {len(lido_validators)}"})
+
+        cdtp = CommonDataToProcess(
```
F4ever, line 42: We are caching almost all requests to nodes, so there is no need to store them like this.
We can just call request data again.
Part of this won't be used in Accounting or in Ejector, but supporting such big structure will take some time and reduce readability

src/modules/accounting/accounting.py
```diff
@@ -28,4 +32,25 @@ def execute_module(self, blockstamp: BlockStamp):
...
+          all_validators,
+          lido_validators,
+          lido_keys,
+          member_info.last_report_ref_slot,
+          member_info.last_report_ref_slot // chain_config.slots_per_epoch,
+          (blockstamp.slot_number - member_info.last_report_ref_slot) * chain_config.seconds_per_slot,
+        )
+
+        b = BunkerService(self.w3, frame_config, chain_config)
+        is_bunker = b.is_bunker_mode(cdtp)
```
F4ever, line 55: Lets provide here
```
blockstamp
member_info
frame_config
chain_config
```
Rest data you can fetch via w3 or calculate inside service

src/modules/accounting/typings.py
```diff
@@ -0,0 +1,22 @@
+from dataclasses import dataclass
+
+from src.providers.consensus.typings import Validator
+from src.providers.keys.typings import LidoKey
+from src.typings import BlockStamp, SlotNumber
+from src.web3_extentions import LidoValidator
+
+Gwei = int
```
F4ever, line 8: Gwei = NewType('Gwei', int)

src/modules/accounting/typings.py
```diff
@@ -0,0 +1,22 @@
...
+Epoch = int
```
F4ever, line 9: There is EpochNumber in typings already

src/modules/accounting/typings.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@dataclass(frozen=True)
+class CommonDataToProcess:
```
F4ever, line 13: This class name is very abstract

src/modules/submodules/consensus.py
```diff
@@ -102,7 +115,7 @@ def _get_member_info(self, blockstamp: BlockStamp) -> MemberInfo:
 
         return MemberInfo(
             is_member=is_member,
-            last_report_ref_slot=last_report_ref_slot,
+            last_report_ref_slot=last_member_report_ref_slot,
```
F4ever, line 118: Let's rename last_report_ref_slot -> last_member_report_ref_slot in MemberInfo

src/modules/submodules/consensus.py
```diff
@@ -128,12 +141,12 @@ def get_blockstamp_for_report(self, blockstamp: BlockStamp) -> Optional[BlockSta
             logger.info({'msg': 'Deadline missed.'})
             return
 
-        return self._get_first_non_missed_slot(blockstamp, member_info.current_ref_slot)
+        return self.get_first_non_missed_slot(blockstamp, member_info.current_ref_slot)
 
-    def _get_first_non_missed_slot(self, blockstamp: BlockStamp, slot: SlotNumber) -> BlockStamp:
-        _, epoch_per_frame = self._get_frame_config(blockstamp)
+    def get_first_non_missed_slot(self, blockstamp: BlockStamp, slot: SlotNumber) -> BlockStamp:
```
F4ever, line 146: Why it's not private anymore?

src/modules/submodules/consensus.py
```diff
@@ -128,12 +141,12 @@ def get_blockstamp_for_report(self, blockstamp: BlockStamp) -> Optional[BlockSta
...
+        frame_config = self.get_frame_config(blockstamp)
 
-        for i in range(slot, slot - epoch_per_frame * 32, -1):
+        for i in range(slot, slot - frame_config.epochs_per_frame * 32, -1):
```
F4ever, line 149: 32 - slots_per_epoch

src/web3_extentions/lido_validators.py
```diff
@@ -30,11 +30,19 @@ def get_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
         lido_keys = self.w3.kac.get_all_lido_keys(blockstamp)
         validators = self.w3.cc.get_validators(blockstamp.state_root)
 
-        return self._merge_validators(lido_keys, validators)
+        return self.filter_lido_validators(lido_keys, validators)
+
+    @lru_cache(maxsize=1)
+    def get_lido_validators_with_others(self, blockstamp: BlockStamp) -> tuple[list[Validator], list[LidoKey], list[LidoValidator]]:
```
F4ever, line 36: Not sure this method should be here
Lets move it to bunker file/folder

src/web3_extentions/lido_validators.py
```diff
@@ -30,11 +30,19 @@ def get_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
...
+        lido_keys = self.w3.kac.get_all_lido_keys(blockstamp)
+        validators = self.w3.cc.get_validators(blockstamp.state_root)
+        lido_validators = self.filter_lido_validators(lido_keys, validators)
+
+        return validators, lido_keys, lido_validators
 
     @staticmethod
-    def _merge_validators(keys: list[LidoKey], validators: list[Validator]) -> list[LidoValidator]:
-        """Merging and filter non-lido validators."""
+    def filter_lido_validators(keys: list[LidoKey], validators: list[Validator]) -> list[LidoValidator]:
```
F4ever, line 44: lets make this private

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+            if nearest_cl_rebase < 0 or far_cl_rebase < 0:
+                return True
+        return False
+
+    def _calculate_normal_cl_rebase(self, cdtp: CommonDataToProcess) -> Gwei:
+        """
+        Calculate normal CL rebase (relative to all validators and the previous Lido frame)
+        for current frame for Lido validators
+        """
+        last_report_blockstamp = self.cm.get_first_non_missed_slot(cdtp.ref_blockstamp, cdtp.last_report_ref_slot)
```
F4ever, line 175: Lets move this method outside conensus class and make it static

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+        )
+
+        cl_rebase = ref_lido_balance_with_vault - corrected_prev_lido_balance_with_vault
+
+        logger.info({"msg": f"CL rebase from {prev_epoch=}: {cl_rebase} Gwei"})
+
+        return cl_rebase
+
+    def _calculate_lido_balance_with_vault(
+        self, lido_validators: list[LidoValidator], vault_balance_at: BlockStamp
```
F4ever, line 260: vault_balance_at - sounds like we expect here address

F4ever: Lets rename somehow

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+    ) -> tuple[Gwei, Gwei]:
+        """
+        Calculate Lido validators balance and balance in withdrawal vault contract in particular block
+        """
+        lido_balance = sum([int(v.validator.balance) for v in lido_validators])
```
F4ever, line 265:
```suggestion
        lido_balance = sum(int(v.validator.balance) for v in lido_validators)
```

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+            "msg": "Simulate 'handleOracleReport' contract function call",
+            "call_args": args,
+            "before_call": before_report_total_pooled_ether,
+            "after_call": after_report_total_pooled_ether,
+        })
+
+        frame_cl_rebase = self.w3.from_wei(after_report_total_pooled_ether - before_report_total_pooled_ether, 'gwei')
+        logger.info({"msg": f"Simulated CL rebase for frame: {frame_cl_rebase} Gwei"})
+
+        return frame_cl_rebase
```
F4ever, line 84:
```suggestion
        return Gwei(frame_cl_rebase)
```

F4ever: Hm, the function return type is gwei, but it returns wei

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+    def _detect_slashing_epoch_range(validator: Validator, ref_epoch: Epoch) -> Sequence[Epoch]:
+        """
+        Detect slashing epoch range for validator
+        If difference between validator's withdrawable epoch and exit epoch is greater enough,
+        then we can be sure that validator was slashed in particular epoch
+        Otherwise, we can only assume that validator was slashed in epochs range
+        due because its exit epoch shifted because of huge exit queue
+        Read more here: https://hackmd.io/@lido/r1Qkkiv3j
+        """
+        assert validator.validator.slashed, "Validator should be slashed to detect slashing epoch range"
```
F4ever, line 333: No asserts. Raise exceptiopns
Asserts could be turned off in production code.

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+        is_known_slashed_epoch = False
 
-    def is_bunker_mode(self) -> bool:
-        pass
+        v = validator.validator
+        if int(v.withdrawable_epoch) - int(v.exit_epoch) > MIN_VALIDATOR_WITHDRAWABILITY_DELAY:
```
F4ever, line 337:
```suggestion
    possible_slashed_epoch = int(v.withdrawable_epoch) - EPOCHS_PER_SLASHINGS_VECTOR
        if int(v.withdrawable_epoch) - int(v.exit_epoch) > MIN_VALIDATOR_WITHDRAWABILITY_DELAY:
            return range(possible_slashed_epoch, possible_slashed_epoch + 1)
```

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+from src.web3_extentions import LidoValidator
 from src.web3_extentions.typings import Web3
 
+from src.modules.accounting.typings import CommonDataToProcess, Gwei, Epoch
+from src.modules.submodules.consensus import ConsensusModule, FrameConfig, ChainConfig
+from src.providers.consensus.typings import ValidatorStatus, Validator
+from src.typings import BlockStamp, SlotNumber
+
+MIN_VALIDATOR_WITHDRAWABILITY_DELAY = 256
+EPOCHS_PER_SLASHINGS_VECTOR = 8192
```
F4ever, line 15: Add some comments about this constants

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+        possible_slashed_epoch = int(v.withdrawable_epoch) - EPOCHS_PER_SLASHINGS_VECTOR
 
-    def _check_slashings(self):
-        pass
+        if is_known_slashed_epoch:
+            return range(possible_slashed_epoch, possible_slashed_epoch + 1)
 
-    def _check_activity_leak(self):
-        pass
+        return range(possible_slashed_epoch, ref_epoch + 1)
```
F4ever, line 344: Why not just ref_epoch here?

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+            for epoch in possible_epochs_with_slashings:
+                per_epoch_buckets[epoch].append(validator)
+
+        min_bucket_epoch = min(per_epoch_buckets.keys())
+
+        # Iterate through per_epoch_buckets and calculate lido midterm penalties for each bucket
+        per_epoch_lido_midterm_penalties = defaultdict(int)
+        # We should calculate total_balance for each bucket, but we do it once for all per_epoch_buckets
+        total_balance = self._calculate_total_effective_balance(cdtp.ref_all_validators)
+        for epoch, slashed_validators in per_epoch_buckets.items():
```
F4ever, line 111: Hope we will never got into bunker mode)

F4ever: I'm afraid it could take a lot of time to calculate everything

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+            slashed_keys = [k.validator.pubkey for k in slashed_validators]
```
F4ever, line 112:
```suggestion
            slashed_keys = (k.validator.pubkey for k in slashed_validators)
```

F4ever: This will increase perf

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+            "before_call": before_report_total_pooled_ether,
+            "after_call": after_report_total_pooled_ether,
+        })
+
+        frame_cl_rebase = self.w3.from_wei(after_report_total_pooled_ether - before_report_total_pooled_ether, 'gwei')
+        logger.info({"msg": f"Simulated CL rebase for frame: {frame_cl_rebase} Gwei"})
+
+        return frame_cl_rebase
+
+    def _is_high_midterm_slashing_penalty(self, cdtp: CommonDataToProcess, cl_rebase: Gwei) -> bool:
```
F4ever, line 86: Lets split this function. It should increase readability and decrease tests difficalty

src/services/bunker.py
```diff
@@ -1,15 +1,344 @@
...
+            slashed_keys = [k.validator.pubkey for k in slashed_validators]
+            lido_validators_slashed_in_epoch = [
+                v for v in lido_slashed_validators if v.validator.pubkey in slashed_keys
+            ]
+            if not lido_validators_slashed_in_epoch:
+                continue
+            # We should calculate penalties according to bounded slashings in past EPOCHS_PER_SLASHINGS_VECTOR
+            min_bounded_epoch = max(min_bucket_epoch, epoch - EPOCHS_PER_SLASHINGS_VECTOR)
+            bounded_slashings_balance_sum = sum([
+                sum(int(v.validator.effective_balance) for v in e_vals)
```
F4ever, line 121: you can use reduce here

src/modules/accounting/accounting.py
vgorkavenko: I'll remove it

src/services/bunker.py
vgorkavenko: No, we convert from wei to gwei above

src/services/bunker.py
vgorkavenko: "big problems require big calculations" :smiling_face_with_tear: 
But actually only requests to states will take a lot of time, I guess

src/services/bunker.py
vgorkavenko: I'm not sure it's necessary. It's just sum of internal values by condition. it seems like we'll make readability worse this way

src/services/bunker.py
vgorkavenko: Yes. The major part should be static. will be done

src/modules/accounting/accounting.py
vgorkavenko: Agree. Will be reforged

src/modules/accounting/accounting.py
vgorkavenko: Done

src/modules/accounting/typings.py
vgorkavenko: done

src/modules/accounting/typings.py
vgorkavenko: removed

src/modules/accounting/typings.py
vgorkavenko: Removed

src/modules/submodules/consensus.py
vgorkavenko: done

src/modules/submodules/consensus.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: fixed

src/services/bunker.py
vgorkavenko: fixed

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: reforged

src/services/bunker.py
```diff
@@ -1,15 +1,395 @@
...
+    def _get_cl_rebase_for_frame(self, blockstamp: BlockStamp) -> Gwei:
+        logger.info({"msg": "Getting CL rebase for frame"})
+        ref_lido_balance, ref_withdrawal_vault_balance = self._calculate_lido_balance_with_vault(
+            self.lido_validators, blockstamp
+        )
+        # Make a static call of 'handleOracleReport' without EL rewards
+        # to simulate report and get total pool ether after that
+        args = {
+            "_reportTimestamp": blockstamp.slot_number * self.c_conf.seconds_per_slot + self.c_conf.genesis_time,
+            "_timeElapsed": (blockstamp.slot_number - self.m_info.last_member_report_ref_slot) * self.c_conf.seconds_per_slot,
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        if frame_cl_rebase < 0:
+            logger.info({"msg": "Bunker ON. CL rebase is negative"})
+            return True
+        if self._is_high_midterm_slashing_penalty(blockstamp, frame_cl_rebase):
+            logger.info({"msg": "Bunker ON. High midterm slashing penalty"})
+            return True
+        if self._is_abnormal_cl_rebase(blockstamp, frame_cl_rebase):
+            logger.info({"msg": "Bunker ON. Abnormal CL rebase"})
+            return True
+
```
avsetsin, line 66: I would suggest adding `return False` at the end of the method

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        for validator in all_slashed_validators:
+            v = validator.validator
+            if not v.slashed:
+                raise Exception("Validator should be slashed to detect slashing epoch range")
+            possible_slashed_epoch = int(v.withdrawable_epoch) - EPOCHS_PER_SLASHINGS_VECTOR
+            if int(v.withdrawable_epoch) - int(v.exit_epoch) > MIN_VALIDATOR_WITHDRAWABILITY_DELAY:
+                per_epoch_buckets[possible_slashed_epoch].append(validator)
+                continue
+            else:
+                for epoch in range(possible_slashed_epoch, ref_epoch + 1):
```
avsetsin, line 337:
```
               <-------36d------>
      [-------?------------------] slashing
      se     pse                 we
====================|=====================> time
                ref_epoch
```
It shows a slashing, the beginning of which cannot be determined by validator state

`se` - slashed epoch
`pse` - possible slashed epoch (we - 36d)
`we` - withdrawable epoch

We are interested in the following range (`x` for mandatory and `o` for optional):
```
               <-------36d------>
      [-------?------------------] slashing
      se     pse                 we
ooooooxxxxxxxxx=====|=====================> time
                ref_epoch
```

avsetsin: I guess the range should be [ref_epoch - 36d, pse]

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+            slashings = sum(bounded_slashed_validators.values())
+            adjusted_total_slashing_balance = min(
+                slashings * PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX,
+                total_balance
+            )
+            for v in lido_validators_slashed_in_epoch:
+                effective_balance = int(v.validator.effective_balance)
+                penalty_numerator = effective_balance // EFFECTIVE_BALANCE_INCREMENT * adjusted_total_slashing_balance
+                penalty = penalty_numerator // total_balance * EFFECTIVE_BALANCE_INCREMENT
+                per_epoch_lido_midterm_penalties[epoch][v.index] = penalty
```
avsetsin, line 375: `epoch` here is an expected slashed epoch, not a midterm epoch.
midterm epoch is calculated as: `validator.withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR // 2`

https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#slashings

src/utils/helpers.py
```diff
@@ -0,0 +1,31 @@
+import logging
+from src.providers.consensus.client import ConsensusClient
+from src.providers.http_provider import NotOkResponse
+from src.typings import BlockStamp, SlotNumber
+
+logger = logging.getLogger(__name__)
+
+
+def get_first_non_missed_slot(cc: ConsensusClient, slot: SlotNumber, max_deep: int) -> BlockStamp:
+    for i in range(slot, slot - max_deep, -1):
```
avsetsin, line 10: max(slot - max_deep, 0)

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        for epoch, validator_penalty in sorted(per_epoch_lido_midterm_penalties.items()):
+            if epoch > right_border:
+                index += 1
+                left_border = right_border + 1
+                right_border += frame_config.epochs_per_frame
+            if left_border <= epoch <= right_border:
+                for val_index, penalty in validator_penalty.items():
+                    if val_index not in per_frame_buckets[index]:
+                        per_frame_buckets[index][val_index] = penalty
+        return [sum(penalties.values()) for penalties in per_frame_buckets.values()]
```
avsetsin, line 403: it seems it can be simplified to

```python
for epoch, validator_penalty in per_epoch_lido_midterm_penalties.items():
  frame = get_frame_by_epoch(epoch)
  
  if not per_frame_buckets[frame]:
    per_frame_buckets[frame] = 0

  per_frame_buckets[frame] += validator_penalty

return per_frame_buckets


def get_frame_by_epoch(epoch):
  return (epoch - initialEpoch) // epochsPerFrame
```

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+            v.validator for v in self.w3.lido_validators.filter_lido_validators(self.lido_keys, all_slashed_validators)
+        ]
+        logger.info({"msg": f"Slashed: {len(all_slashed_validators)=} | {len(lido_slashed_validators)=}"})
+
+        # If no one lido in current slashed validators - no need to bunker
+        if not lido_slashed_validators:
+            return False
+
+        # We should calculate total_balance for each bucket, but we do it once for all per_epoch_buckets
+        total_balance = self._calculate_total_effective_balance(self.all_validators)
```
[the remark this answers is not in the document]
vgorkavenko: Not exactly. It should be the sum of effective balances of only active validators.
Description of `get_total_active_balance` from spec tells us following:
`Return the combined effective balance of the active validators.`

I will add `active` to the function name for clarity

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+            bounded_slashed_validators = {}
+            for e, e_vals in per_epoch_buckets.items():
+                if min_bounded_epoch <= e <= epoch:
+                    for v in e_vals:
+                        if v.validator.pubkey not in bounded_slashed_validators:
+                            bounded_slashed_validators[v.index] = int(v.validator.effective_balance)
+            slashings = sum(bounded_slashed_validators.values())
+            adjusted_total_slashing_balance = min(
+                slashings * PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX,
+                total_balance
```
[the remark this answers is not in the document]
vgorkavenko: Answered above

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+    def _get_withdrawn_from_vault_between(self, prev_blockstamp: BlockStamp, curr_blockstamp: BlockStamp) -> int:
+        """
+        Lookup for ETHDistributed event and sum up all withdrawalsWithdrawn
+        """
+        prev_epoch = prev_blockstamp.slot_number // self.c_conf.slots_per_epoch
+        curr_epoch = curr_blockstamp.slot_number // self.c_conf.slots_per_epoch
+        logger.info({"msg": f"Get withdrawn from vault between {prev_epoch,curr_epoch=}"})
+        vault_withdrawals = 0
+        events = self.w3.lido_contracts.lido.events.ETHDistributed.get_logs(
+            from_block=int(prev_blockstamp.block_number) + 1, to_block=int(curr_blockstamp.block_number)
```
avsetsin, line 263: Could you add a comment on why we are adding 1 block here?

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+
+    def _calculate_lido_balance_with_vault(
+        self, lido_validators: list[LidoValidator], vault_balance_at_blockstamp: BlockStamp
+    ) -> tuple[Gwei, Gwei]:
+        """
+        Calculate Lido validators balance and balance in withdrawal vault contract in particular block
+        """
+        lido_balance = sum(int(v.validator.balance) for v in lido_validators)
+        withdrawal_vault_balance = self.w3.eth.get_balance(
+            self.w3.lido_contracts.withdrawal_vault.address, vault_balance_at_blockstamp.block_hash
```
avsetsin, line 249: 🤔 it might be worth getting the withdrawal_vault address on the ref_block in case of future protocol upgrades

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+
+        nearest_cl_rebase = self._calculate_cl_rebase_between(nearest_blockstamp, blockstamp)
+        far_cl_rebase = self._calculate_cl_rebase_between(far_blockstamp, blockstamp)
+
+        logger.info({"msg": f"CL rebase {nearest_cl_rebase,far_cl_rebase=}"})
+        return nearest_cl_rebase, far_cl_rebase
+
+    def _calculate_cl_rebase_between(self, prev_blockstamp: BlockStamp, curr_ref_blockstamp: BlockStamp) -> Gwei:
+        """
+        Calculate CL rebase from prev_blockstamp to ref_blockstamp given the changes in withdrawal vault
```
avsetsin, line 209: Maybe add a bit more details?

Skimmed validator rewards are sent to 0x01 withdrawal credentials address, which in the case of Lido is a withdrawal vault. To account for all changes in validators' balances, we must take its balance into account

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        logger.info({"msg": f"Normal CL rebase: {normal_cl_rebase} Gwei"})
+        return Gwei(normal_cl_rebase)
+
+    def _calculate_nearest_and_far_cl_rebase(self, blockstamp: BlockStamp) -> tuple[Gwei, Gwei]:
+        """
+        Calculate CL rebase from nearest and far epochs to ref epoch given the changes in withdrawal vault
+        """
+        logger.info({"msg": "Calculating nearest and far CL rebase"})
+        nearest_slot = (blockstamp.slot_number - NEAREST_EPOCH_DISTANCE * self.c_conf.slots_per_epoch)
+        far_slot = (blockstamp.slot_number - FAR_EPOCH_DISTANCE * self.c_conf.slots_per_epoch)
```
avsetsin, line 195: I would suggest adding some sanity checks here:
- nearest_slot >= far_slot
- far_slot > previous_report_ref_slot

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+
+        lookup_max_deep = self.c_conf.slots_per_epoch * self.f_conf.epochs_per_frame
```
avsetsin, line 197: I wonder if a whole day can be missed? 🤔 Let's say in the case of an inactive leak and we are in a separate smaller validator group

avsetsin: I think it's okay to make the assumption that this will never happen, but we could, for example, put it in a separate method and add a description that we thought about it and decided that this level of paranoia is enough

src/services/bunker.py
avsetsin: Do we want to handle cases where one distance = 0, both = 0, X = x?

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        nearest_blockstamp = get_first_non_missed_slot(self.w3.cc, SlotNumber(nearest_slot), lookup_max_deep)
+        far_blockstamp = get_first_non_missed_slot(self.w3.cc, SlotNumber(far_slot), lookup_max_deep)
+
+        nearest_cl_rebase = self._calculate_cl_rebase_between(nearest_blockstamp, blockstamp)
+        far_cl_rebase = self._calculate_cl_rebase_between(far_blockstamp, blockstamp)
```
avsetsin, line 202: Do we want some optimization, like adding a cache for some methods?

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
-        pass
+    @staticmethod
+    def _calculate_normal_cl_rebase(
+        epochs_passed: int, mean_total_lido_effective_balance: int, mean_total_effective_balance: int
+    ) -> Gwei:
+        normal_cl_rebase = int(
+            NORMALIZED_CL_PER_EPOCH * mean_total_lido_effective_balance * epochs_passed
+            / math.sqrt(mean_total_effective_balance) * (1 - MISTAKE_RANGE)
+        )
+        return Gwei(normal_cl_rebase)
```
avsetsin, line 413: I didn't figure out the formulas, would appreciate it if you could tell me on the call

src/services/bunker.py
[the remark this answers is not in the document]
avsetsin: Yeah, you're right, I missed it

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+from src.typings import BlockStamp, SlotNumber, EpochNumber
+
+# Constants from consensus spec
+MIN_VALIDATOR_WITHDRAWABILITY_DELAY = 256
+EPOCHS_PER_SLASHINGS_VECTOR = 8192
+PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX = 3
+EFFECTIVE_BALANCE_INCREMENT = 2 ** 0 * 10 ** 9
+MIN_DEPOSIT_AMOUNT = 32 * 10 ** 9
+
+FIRST_DEPOSIT_EVENT_LOOKUP_MAXIMUM_BLOCKS_DISTANCE = 225 * 7
```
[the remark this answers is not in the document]
vgorkavenko: removed

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+    ) -> tuple[Gwei, Gwei]:
+        """
+        Calculate Lido validators balance and balance in withdrawal vault contract in particular block
+        """
+        lido_balance = sum(int(v.validator.balance) for v in lido_validators)
+        withdrawal_vault_balance = self.w3.eth.get_balance(
+            self.w3.lido_contracts.withdrawal_vault.address, vault_balance_at_blockstamp.block_hash
+        )
+
+        return Gwei(lido_balance), Gwei(self.w3.from_wei(withdrawal_vault_balance, 'gwei'))
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+    def _is_high_midterm_slashing_penalty(self, blockstamp: BlockStamp, cl_rebase: Gwei) -> bool:
+        logger.info({"msg": "Detecting high midterm slashing penalty"})
+        ref_epoch = EpochNumber(blockstamp.slot_number // self.c_conf.slots_per_epoch)
+        all_slashed_validators = self._not_withdrawn_slashed_validators(self.all_validators, ref_epoch)
+        lido_slashed_validators: list[Validator] = [
+            v.validator for v in self.w3.lido_validators.filter_lido_validators(self.lido_keys, all_slashed_validators)
+        ]
+        logger.info({"msg": f"Slashed: {len(all_slashed_validators)=} | {len(lido_slashed_validators)=}"})
+
+        # If no one lido in current slashed validators - no need to bunker
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        all_slashed_validators: list[Validator], ref_epoch: EpochNumber
+    ) -> dict[EpochNumber, list[Validator]]:
+        """
+        Fill per_epoch_buckets by possible slashed epochs
+        It detects slashing epoch range for validator
+        If difference between validator's withdrawable epoch and exit epoch is greater enough,
+        then we can be sure that validator was slashed in particular epoch
+        Otherwise, we can only assume that validator was slashed in epochs range
+        due because its exit epoch shifted because of huge exit queue
+        Read more here: https://hackmd.io/@lido/r1Qkkiv3j
```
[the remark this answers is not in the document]
vgorkavenko: yep. done

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: Yes, sounds like the right approach. Done

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+                continue
+            # We should calculate penalties according to bounded slashings in past EPOCHS_PER_SLASHINGS_VECTOR
+            min_bounded_epoch = max(min_bucket_epoch, EpochNumber(epoch - EPOCHS_PER_SLASHINGS_VECTOR))
+            bounded_slashed_validators = {}
+            for e, e_vals in per_epoch_buckets.items():
+                if min_bounded_epoch <= e <= epoch:
+                    for v in e_vals:
+                        if v.validator.pubkey not in bounded_slashed_validators:
+                            bounded_slashed_validators[v.index] = int(v.validator.effective_balance)
+            slashings = sum(bounded_slashed_validators.values())
```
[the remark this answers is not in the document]
vgorkavenko: done

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+            if not lido_validators_slashed_in_epoch:
+                continue
+            # We should calculate penalties according to bounded slashings in past EPOCHS_PER_SLASHINGS_VECTOR
+            min_bounded_epoch = max(min_bucket_epoch, EpochNumber(epoch - EPOCHS_PER_SLASHINGS_VECTOR))
+            bounded_slashed_validators = {}
+            for e, e_vals in per_epoch_buckets.items():
+                if min_bounded_epoch <= e <= epoch:
+                    for v in e_vals:
+                        if v.validator.pubkey not in bounded_slashed_validators:
+                            bounded_slashed_validators[v.index] = int(v.validator.effective_balance)
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
vgorkavenko: Yes, you are right. Fixed

src/utils/helpers.py
vgorkavenko: fixed

src/services/bunker.py
vgorkavenko: yes. reforged

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        per_epoch_lido_midterm_penalties = self._get_per_epoch_lido_midterm_penalties(
+            per_epoch_buckets, lido_slashed_validators, total_balance
+        )
+        # Calculate lido midterm penalties impact in each frame
+        per_frame_buckets = self._get_per_frame_lido_midterm_penalties(per_epoch_lido_midterm_penalties, self.f_conf)
+
+        # If any midterm penalty sum of lido validators in frame bucket greater than rebase we should trigger bunker
+        max_lido_midterm_penalty = max(per_frame_buckets)
+        logger.info({"msg": f"Max lido midterm penalty: {max_lido_midterm_penalty}"})
+        if max_lido_midterm_penalty > cl_rebase:
```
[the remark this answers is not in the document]
vgorkavenko: done

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+            if nearest_cl_rebase < 0 or far_cl_rebase < 0:
+                return True
+        return False
+
+    def _get_normal_cl_rebase(self, blockstamp: BlockStamp) -> Gwei:
+        """
+        Calculate normal CL rebase (relative to all validators and the previous Lido frame)
+        for current frame for Lido validators
+        """
+        current_ref_epoch = EpochNumber(blockstamp.slot_number // self.c_conf.slots_per_epoch)
```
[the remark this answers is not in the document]
vgorkavenko: will be done, I guess. I only use it in my code for now

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        last_report_blockstamp = get_first_non_missed_slot(
+            self.w3.cc, self.last_report_ref_slot, self.c_conf.slots_per_epoch * self.f_conf.epochs_per_frame
+        )
+        last_member_report_ref_epoch = EpochNumber(last_report_blockstamp.slot_number // self.c_conf.slots_per_epoch)
+
+        total_ref_effective_balance = self._calculate_total_effective_balance(self.all_validators)
+        total_ref_lido_effective_balance = self._calculate_total_effective_balance(
+            [v.validator for v in self.lido_validators]
```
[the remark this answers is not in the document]
vgorkavenko: will be changed with LidoValidator dataclass

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        )
+
+        if last_member_report_ref_epoch <= self.f_conf.initial_epoch:
```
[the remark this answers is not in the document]
vgorkavenko: renamed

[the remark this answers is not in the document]
vgorkavenko: Left `todo` to make a choice

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+    def _get_normal_cl_rebase(self, blockstamp: BlockStamp) -> Gwei:
+        """
+        Calculate normal CL rebase (relative to all validators and the previous Lido frame)
+        for current frame for Lido validators
+        """
+        current_ref_epoch = EpochNumber(blockstamp.slot_number // self.c_conf.slots_per_epoch)
+        last_report_blockstamp = get_first_non_missed_slot(
+            self.w3.cc, self.last_report_ref_slot, self.c_conf.slots_per_epoch * self.f_conf.epochs_per_frame
+        )
+        last_member_report_ref_epoch = EpochNumber(last_report_blockstamp.slot_number // self.c_conf.slots_per_epoch)
```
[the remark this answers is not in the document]
vgorkavenko: done

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+            total_last_lido_effective_balance = self._calculate_total_effective_balance(
+                [v.validator for v in last_lido_validators]
+            )
+
+        mean_total_effective_balance = (total_ref_effective_balance + total_last_effective_balance) // 2
+        mean_total_lido_effective_balance = (
+            (total_ref_lido_effective_balance + total_last_lido_effective_balance) // 2
+        )
+
+        epochs_passed = current_ref_epoch - last_completed_epoch
```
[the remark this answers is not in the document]
vgorkavenko: fixed above

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        lido_balance = sum(int(v.validator.balance) for v in lido_validators)
+        withdrawal_vault_balance = self.w3.eth.get_balance(
+            self.w3.lido_contracts.withdrawal_vault.address, vault_balance_at_blockstamp.block_hash
+        )
+
+        return Gwei(lido_balance), Gwei(self.w3.from_wei(withdrawal_vault_balance, 'gwei'))
+
+    def _get_withdrawn_from_vault_between(self, prev_blockstamp: BlockStamp, curr_blockstamp: BlockStamp) -> int:
+        """
+        Lookup for ETHDistributed event and sum up all withdrawalsWithdrawn
```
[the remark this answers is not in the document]
vgorkavenko: Yes, that makes sense. Done

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: yes. added

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: done. I think `X = x` case no need to handle specially, because we will have caching for `self._calculate_cl_rebase_between`

src/services/bunker.py
vgorkavenko: yes. I added lru

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+    def _get_cl_rebase_for_frame(self, blockstamp: BlockStamp) -> Gwei:
+        logger.info({"msg": "Getting CL rebase for frame"})
+        ref_lido_balance, ref_withdrawal_vault_balance = self._calculate_lido_balance_with_vault(
+            self.lido_validators, blockstamp
+        )
+        # Make a static call of 'handleOracleReport' without EL rewards
+        # to simulate report and get total pool ether after that
+        args = {
+            "_reportTimestamp": blockstamp.slot_number * self.c_conf.seconds_per_slot + self.c_conf.genesis_time,
+            "_timeElapsed": (blockstamp.slot_number - self.last_report_ref_slot) * self.c_conf.seconds_per_slot,
```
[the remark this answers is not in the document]
vgorkavenko: done

src/services/bunker.py
```diff
@@ -1,15 +1,413 @@
...
+        total_balance: Gwei,
+    ) -> dict[EpochNumber, dict[str, Gwei]]:
+        """
+        Iterate through per_epoch_buckets and calculate lido midterm penalties for each bucket
+        """
+        min_bucket_epoch = min(per_epoch_buckets.keys())
+        per_epoch_lido_midterm_penalties: dict[EpochNumber, dict[str, Gwei]] = defaultdict(dict)
+        for epoch, slashed_validators in per_epoch_buckets.items():
+            slashed_indexes = set(k.index for k in slashed_validators)
+            lido_validators_slashed_in_epoch = [v for v in lido_slashed_validators if v.index in slashed_indexes]
```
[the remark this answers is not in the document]
vgorkavenko: reforged to dicts

src/services/bunker.py
```diff
@@ -1,15 +1,480 @@
...
+                penalty_numerator = effective_balance // EFFECTIVE_BALANCE_INCREMENT * adjusted_total_slashing_balance
+                penalty = penalty_numerator // total_balance * EFFECTIVE_BALANCE_INCREMENT
+                midterm_penalty_epoch = EpochNumber(
+                    int(v.validator.withdrawable_epoch) - EPOCHS_PER_SLASHINGS_VECTOR // 2
+                )
+                per_epoch_lido_midterm_penalties[midterm_penalty_epoch][key] = penalty
+        return per_epoch_lido_midterm_penalties
+
+    @staticmethod
+    def _get_bounded_slashed_validators(
```
avsetsin, line 438: documentation of this method is welcome

src/services/bunker.py
```diff
@@ -1,15 +1,480 @@
...
 
-    def is_bunker_mode(self) -> bool:
-        pass
+    @staticmethod
+    def _get_frame_by_epoch(epoch: EpochNumber, frame_config: FrameConfig) -> int:
+        return abs(epoch - frame_config.initial_epoch) // frame_config.epochs_per_frame
 
-    def _check_slashings(self):
-        pass
+    def _calculate_normal_cl_rebase(
```
avsetsin, line 472: documentation of this method is welcome

avsetsin (approved):
lgtm

src/modules/submodules/oracle_module.py
```diff
@@ -60,14 +60,17 @@ def _receive_last_finalized_slot(self) -> BlockStamp:
         slot_details = self.w3.cc.get_block_details(block_root)
 
         state_root = StateRoot(slot_details.message.state_root)
-        slot_number = SlotNumber(int(slot_details.message.slot))
+        ref_slot_number = slot_number = SlotNumber(int(slot_details.message.slot))
+        ref_epoch = EpochNumber(ref_slot_number // 32)  # todo: do better
```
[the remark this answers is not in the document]
vgorkavenko: yep. will be fixed here https://github.com/lidofinance/lido-oracle/pull/187

src/services/bunker.py
```diff
@@ -1,15 +1,467 @@
...
+            return True
+        if self._is_abnormal_cl_rebase(blockstamp, frame_cl_rebase):
+            logger.info({"msg": "Bunker ON. Abnormal CL rebase"})
+            return True
+
+        return False
+
+    def _get_config(self, blockstamp: BlockStamp):
+        contract = self.w3.eth.contract(
+            address=Address(b'0x2663f77bee97abFE76551b47342F5FAD3a9BDDdC'),
```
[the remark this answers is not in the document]
vgorkavenko: fixed

src/services/bunker.py
vgorkavenko: done

src/services/bunker.py
vgorkavenko: done

PR #186, opened 2023-02-12
fix: few things

vgorkavenko:
- `ValidatorState` and  `Validator` typings
- cast `str` to `int` in `_get_with_blockstamp`
- fix `_get_first_non_missed_slot`

### DIFF
--- a/src/modules/submodules/consensus.py
+++ b/src/modules/submodules/consensus.py
@@ -102,7 +102,7 @@ class ConsensusModule(ABC):
 
         return MemberInfo(
             is_member=is_member,
-            last_report_ref_slot=last_member_report_ref_slot,
+            last_report_ref_slot=last_report_ref_slot,
             current_ref_slot=current_ref_slot,
             member_ref_slot=member_ref_slot,
             member_report_for_current_ref_slot=member_report_for_current_ref_slot,
--- a/src/modules/submodules/consensus.py
+++ b/src/modules/submodules/consensus.py
@@ -71,7 +71,7 @@ class ConsensusModule(ABC):
 
         # Defaults for dry mode
         current_ref_slot, deadline_slot = self._get_current_frame(blockstamp)
-        is_member, last_member_report_ref_slot, member_report_for_current_ref_slot = True, 0, b''
+        is_member, last_report_ref_slot, member_report_for_current_ref_slot = True, 0, b''
         member_ref_slot = current_ref_slot
 
         if variables.ACCOUNT:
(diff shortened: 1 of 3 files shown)


src/providers/keys/client.py
```diff
@@ -30,7 +30,7 @@ def _get_with_blockstamp(self, url: str, blockstamp: BlockStamp, params: Optiona
         """
         for i in range(self.RETRY_COUNT):
             data, meta = self._get(url, params)
-            if meta['meta']['elBlockSnapshot']['blockNumber'] >= blockstamp.block_number:
+            if meta['meta']['elBlockSnapshot']['blockNumber'] >= int(blockstamp.block_number):
```
F4ever, line 33: This is already int

src/modules/submodules/consensus.py
```diff
@@ -70,7 +71,7 @@ def _get_member_info(self, blockstamp: BlockStamp) -> MemberInfo:
 
         # Defaults for dry mode
         current_ref_slot, deadline_slot = self._get_current_frame(blockstamp)
-        is_member, last_report_ref_slot, member_report_for_current_ref_slot = True, 0, b''
+        is_member, last_member_report_ref_slot, member_report_for_current_ref_slot = True, 0, b''
```
F4ever, line 74: It changes nothing

src/providers/consensus/typings.py
```diff
@@ -42,16 +42,16 @@ class ValidatorState:
     withdrawal_credentials: str
     effective_balance: str
     slashed: bool
-    activation_eligibility_epoch: bool
+    activation_eligibility_epoch: str
```
F4ever, line 45: :+1:

F4ever, line 45: :+1:

src/modules/submodules/consensus.py
vgorkavenko: in dry mode `last_report_ref_slot` var was not used in any way further. I think it should be initial value for `last_member_report_ref_slot` to init `MemberInfo`

src/modules/submodules/consensus.py
```diff
@@ -101,7 +102,7 @@ def _get_member_info(self, blockstamp: BlockStamp) -> MemberInfo:
 
         return MemberInfo(
             is_member=is_member,
-            last_report_ref_slot=last_report_ref_slot,
+            last_report_ref_slot=last_member_report_ref_slot,
```
F4ever, line 105:
```suggestion
            last_report_ref_slot=last_report_ref_slot,
```

F4ever, line 105:
```suggestion
            last_report_ref_slot=last_report_ref_slot,
```

F4ever (approved):

PR #188, opened 2023-02-13
Add sanity checker contract

skhomuti:
Read OracleReportLimits for extra data processing


src/services/extra_data.py
```diff
@@ -57,15 +57,21 @@ def __init__(self, w3: Web3):
...
 
         extra_data = self.build_extra_data(stucked_payloads, exited_payloads)
         return self.to_bytes(extra_data)
 
+    def _get_oracle_report_limits(self):
+        abi = next(filter(lambda x: x.get('name') == 'getOracleReportLimits',
+                          self.w3.lido_contracts.oracleReportSanityChecker.abi))
+        output_names = list(map(itemgetter('name'), abi['outputs'][0]['components']))
+        result = self.w3.lido_contracts.oracleReportSanityChecker.functions.getOracleReportLimits().call()
+        return OracleReportLimits(**dict(zip(output_names, result)))
```
F4ever, line 73: cool idea
Is it possible to write a function that returns abi to write out separately so that it can be used everywhere?

F4ever: + We need at least one positive test for this feature

src/services/extra_data.py
```diff
@@ -57,15 +57,21 @@ def __init__(self, w3: Web3):
         self.w3 = w3
 
     def collect(self, blockstamp: BlockStamp, exited_validators: list[LidoKey], stucked_validators: list[LidoKey]) -> bytes:
```
F4ever, line 59: Here I made a mistake, most likely we will return here not keys but lists of (module, op_id, keys).
We shouldn't report values if they wasn't changed, so we need to compare them. - Or we could add this functionality here?

src/services/extra_data.py
skhomuti: Yea will do it
I expect that the next web3py version will be able to do it out of the box: https://github.com/ethereum/web3.py/pull/2799.

src/services/extra_data.py
skhomuti: Yes, we could. Share with me some knowledge about this

src/services/extra_data.py
skhomuti: done

F4ever (approved):

PR #190, opened 2023-02-14
Added reading env vars from .env file if present

skhomuti:
Just a simple improvement


F4ever (approved):

F4ever:
@skhomuti Add .env.example here

PR #192, opened 2023-02-16
Actualize consensus tests

skhomuti:


src/modules/submodules/consensus.py
```diff
@@ -300,6 +300,7 @@ def _send_report_hash(self, blockstamp: BlockStamp, report_hash: bytes, consensu
 
         tx = consensus_contract.functions.submitReport(blockstamp.slot_number, report_hash, consensus_version)
 
+        # TODO is there should be an error if transaction is not valid?
```
skhomuti, line 303: @F4ever look at this

skhomuti:
@F4ever there appear lots of requests when the consensus module instantiated (like eth_getBlockByHash) that didn't exist before
not sure if it's a problem, I just highlight it

src/modules/submodules/consensus.py
F4ever: There are a lot of cases when transaction is not valid for example, report already sent.
But the result of tx shouldn't have any effect on next oracle flow, so we do need to throw here an exception

PR #193, opened 2023-02-16
Feature/withdrawal finalization

rkolpakov:


src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+        if is_slashed_epoch_undetectable:
+            return None
+
+        return withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR
+
+    def find_earliest_slashed_epoch(self, validators, ref_slot):
+        pubkeys = get_validators_pubkeys(validators)
+        withdrawable_epoch = min(get_validators_withdrawable_epochs(validators))
+
+        start_slot = 0
```
avsetsin, line 90: Let's think about optimization here, requests for validator states are heavy. 
We had the following thoughts here that might help:
- There is no point in going any deeper than the last finalized request
- We can use `activation_epoch` of the validator since it could not slash before
- We can iterate through slots with a multiple of 8192 where possible (at high values). To make it easier for the node to get states. Read more about it here: https://lighthouse-book.sigmaprime.io/advanced_database.html

src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+        return min(
+            negative_rebase_border_epoch,
+            associated_slashings_border_epoch
+        )
+
+    def get_new_requests_border_epoch(self, blockstamp: BlockStamp):
+        return get_epoch_by_slot(blockstamp.slot_number) - NEW_REQUESTS_BORDER
+
+    def get_negative_rebase_border_epoch(self, blockstamp: BlockStamp):  
+        bunker_start_timestamp = self.get_bunker_mode_start_timestamp(blockstamp)
```
avsetsin, line 33: need to handle a scenario when the bunker was disabled before (for the first report when is_bunker: true). `bunkerModeSinceTimestamp()` will return maxuint256 in this case

So we need to take the timestamp of the last **successful** report before the current one

src/services/finalizer.py
```diff
@@ -0,0 +1,52 @@
...
+
+SLOTS_PER_EPOCH = 2**5
+SLOT_TIME = 12
+
+class Finalizer:
+    def __init__(self, w3: Web3) -> None:
+        self.w3 = w3
+        self.safe_border_service = SafeBorder(w3)
+
+    def get_withdrawable_requests(self, is_bunker_mode: bool, share_rate: int, blockstamp: BlockStamp):
```
rkolpakov, line 13: `get next last finalizable id`

src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+EPOCHS_PER_SLASHINGS_VECTOR = 2**13
+SLOTS_PER_EPOCH = 2**5
+
+class SafeBorder:
+    def __init__(self, w3: Web3) -> None:
+        self.w3 = w3
+        self.lido_contracts = w3.lido_contracts
+
+    def get_safe_border_epoch(self, is_bunker: bool, blockstamp: BlockStamp):
+        if is_bunker is False:
```
rkolpakov, line 18: if not bunker

src/services/finalizer.py
```diff
@@ -0,0 +1,52 @@
...
+        (first_request_id, last_request_id) = self._get_unfinalized_withdrawal_ids(blockstamp)
+
+        if last_request_id - first_request_id == 0:
+            return None
+
+        withdrawable_until_epoch = self.safe_border_service.get_safe_border_epoch(is_bunker_mode, blockstamp)
```
rkolpakov, line 19: double check if we should use start or end of this epoch

src/services/finalizer.py
```diff
@@ -0,0 +1,52 @@
...
+        available_eth = self._fetch_locked_ether_amount(blockstamp)
+
+        last_finalizable_request_id_by_timestamp = self._fetch_last_finalizable_request_id_by_timestamp(withdrawable_until_timestamp, first_request_id, last_request_id, blockstamp)
+        last_finalizable_request_id_by_budget = self._fetch_last_finalizable_request_id_by_budget(available_eth, share_rate, first_request_id, last_request_id, blockstamp)
+        last_request_id = min(last_finalizable_request_id_by_timestamp, last_finalizable_request_id_by_budget)
+
+        if last_request_id - first_request_id <= 0:
+            return None
+        
+        return list(range(first_request_id, last_request_id + 1))
```
rkolpakov, line 31: double check this request id (timestamp, etc.)

src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+        if earliest_slashed_epoch is None:
+            return get_epoch_by_slot(blockstamp.slot_number) - NEW_REQUESTS_BORDER
+        
+        # should NEW_REQUESTS_BORDER be here?
+        return earliest_slashed_epoch - NEW_REQUESTS_BORDER
+
+    def get_earliest_slashed_epoch_among_incomplete_slashings(self, blockstamp: BlockStamp):
+        validators = self.get_lido_validators(blockstamp)       
+        validators_slashed = filter_slashed_validators(validators)
+        validators_slashed_non_withdrawn = filter_non_withdrawable_validators(validators_slashed, blockstamp.slot_number)
```
rkolpakov, line 53: slashed_withdrawable_epoch_lt_ref

rkolpakov: validators_with_incompleted_impact

src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+    return slot_number // SLOTS_PER_EPOCH
+
+def get_epoch_by_timestamp(timestamp: str):
+    return timestamp // (SLOTS_PER_EPOCH * 12)
+                
+
+# TODO: stop converting back to list
+def filter_slashed_validators(validators):
+    return list(filter(lambda validator: validator.validator.validator.slashed, validators))
+def filter_non_withdrawable_validators(validators, epoch):
```
rkolpakov, line 127: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#helpers
get_eligible_validator_indices

src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+        validators_with_earliest_exit_epoch = self.get_validators_with_earliest_exit_epoch(validators_slashed_non_withdrawn)
+        first_validator_with_earliest_exit_epoch = validators_with_earliest_exit_epoch[0]
+        earliest_slashed_epoch = self.calc_validator_slashed_epoch_from_state(first_validator_with_earliest_exit_epoch)
+        if earliest_slashed_epoch is not None:
+            return earliest_slashed_epoch
+        
+        earliest_slashed_epoch = self.find_earliest_slashed_epoch(validators_with_earliest_exit_epoch, blockstamp.slot_number)
+        return earliest_slashed_epoch
+
+    def get_validators_with_earliest_exit_epoch(self, validators):
```
rkolpakov, line 67: Add comment about assumptions and separate it semantically

src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+        validators_slashed = filter_slashed_validators(validators)
+        validators_slashed_non_withdrawn = filter_non_withdrawable_validators(validators_slashed, blockstamp.slot_number)
+
+        if len(validators_slashed_non_withdrawn) == 0:
+            return None
+
+        validators_with_earliest_exit_epoch = self.get_validators_with_earliest_exit_epoch(validators_slashed_non_withdrawn)
+        first_validator_with_earliest_exit_epoch = validators_with_earliest_exit_epoch[0]
+        earliest_slashed_epoch = self.calc_validator_slashed_epoch_from_state(first_validator_with_earliest_exit_epoch)
+        if earliest_slashed_epoch is not None:
```
rkolpakov, line 61: Do reverse of this if

src/services/safe_border.py
rkolpakov, line 90: max(slot(last_finalized_request_timestamp), slot(activation_epoch))

src/services/safe_border.py
```diff
@@ -0,0 +1,134 @@
...
+        return withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR
+
+    def find_earliest_slashed_epoch(self, validators, ref_slot):
+        pubkeys = get_validators_pubkeys(validators)
+        withdrawable_epoch = min(get_validators_withdrawable_epochs(validators))
+
+        start_slot = 0
+        end_slot = min(ref_slot, get_epoch_first_slot(withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR))
+
+        while start_slot + 1 < end_slot:
```
rkolpakov, line 93: Find a frame where slashing is existing not a slot

skhomuti (changes requested):

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
+from src.web3py.typings import Web3
+from src.typings import BlockStamp
+from src.web3py.extentions.lido_validators import LidoValidator
+from src.modules.submodules.consensus import ChainConfig
+
+DEFAULT_SHIFT = 8 # epochs ~50 min
```
F4ever, line 6: Move to Oracle Onchain configs

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+    def _get_new_requests_border_epoch(self, blockstamp: BlockStamp):
+        return self.get_epoch_by_slot(blockstamp.ref_slot) - DEFAULT_SHIFT
+
+    def _get_negative_rebase_border_epoch(self, blockstamp: BlockStamp):  
+        bunker_start_timestamp = self._get_bunker_mode_start_timestamp(blockstamp)
+        bunker_start_slot = self.get_slot_by_timestamp(bunker_start_timestamp)
+
+        last_withdrawable_epoch = self.get_epoch_by_slot(bunker_start_slot)
+
+        if bunker_start_slot > blockstamp.ref_slot:
```
F4ever, line 40: How it could be True?

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+
+        return min(
+            negative_rebase_border_epoch,
+            associated_slashings_border_epoch
+        )
+
+    def _get_new_requests_border_epoch(self, blockstamp: BlockStamp):
+        return self.get_epoch_by_slot(blockstamp.ref_slot) - DEFAULT_SHIFT
+
+    def _get_negative_rebase_border_epoch(self, blockstamp: BlockStamp):
```
F4ever, line 34: Add return types to all functions

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _get_validators_earliest_activation_epoch(self, validators):
+        if len(validators) == 0:
+            return 0
+
+        sorted_validators = sorted(validators, key = lambda validator: (int(validator.validator.validator.activation_epoch)))
+        return sorted_validators[0].validator.validator.activation_epoch
+
+    def _get_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
+        return self.w3.cc.get_validators(blockstamp.ref_slot)
```
F4ever, line 125: get_validators - Returns Validator type, not LidoValidator.
Use self.w3.cc.lido_validators.get_lido_validators

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _get_archive_lido_validators_by_keys(self, ref_slot, pubkeys):
+        return self.w3.cc.get_validators(ref_slot, tuple(pubkeys))
```
F4ever, line 128: Update unnotation in get_validators that it can recieve tuple here

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+        # Here we filter not by exit_epoch but by withdrawable_epoch because exited operators can still be slashed.
+        # See more here https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#helpers
+        # at `get_eligible_validator_indices` method.
+        validators_slashed_non_withdrawable = filter_non_withdrawable_validators(validators_slashed, blockstamp.ref_slot)
+
+        if len(validators_slashed_non_withdrawable) == 0:
+            return None
+
+        validators_with_earliest_exit_epoch = self._filter_validators_with_earliest_exit_epoch(validators_slashed_non_withdrawable)
+        earliest_slashed_epoch_predicted = self._predict_earliest_slashed_epoch(validators_with_earliest_exit_epoch[0])
```
F4ever, line 69: If at least one of these validators would have predictable epoch, there is no need to find it with binary search
Maybe check all of them, not only the first one?

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+
+        return withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR
+
+    def _find_earliest_slashed_epoch(self, validators, blockstamp: BlockStamp):
+        pubkeys = get_validators_pubkeys(validators)
+        withdrawable_epoch = min(get_validators_withdrawable_epochs(validators))
+
+        last_finalized_request_id_slot = self._get_last_finalized_withdrawal_request_slot(blockstamp)
+
+        start_slot = max(last_finalized_request_id_slot, self._get_validators_earliest_activation_epoch(validators))
```
F4ever, line 94: bug
last_finalized_request_id_slot - slot
self._get_validators_earliest_activation_epoch(validators) -epcoh

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+        end_slot = min(blockstamp.ref_slot, self.get_epoch_first_slot(withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR))
+
+        while start_slot + 1 < end_slot:
+            mid_slot = (end_slot + start_slot) // 2
+            validators = self._get_archive_lido_validators_by_keys(mid_slot, pubkeys)
+            slashed_validators = filter_slashed_validators(validators)
+
+            if len(slashed_validators) > 0:
+                end_slot = mid_slot - 1
```
F4ever, line 103: Some optimization.
We can use slashed_validators only for next requests

src/services/safe_border.py
```diff
@@ -0,0 +1,164 @@
...
+FAR_FUTURE_EPOCH = 2**64 - 1
+MIN_VALIDATOR_WITHDRAWABILITY_DELAY = 2**8
+EPOCHS_PER_SLASHINGS_VECTOR = 2**13
+
+class SafeBorder:
+    def __init__(self, w3: Web3) -> None:
+        self.w3 = w3
+        self.lido_contracts = w3.lido_contracts
+
+    def get_safe_border_epoch(self, is_bunker: bool, blockstamp: BlockStamp, chain_config: ChainConfig):
```
F4ever, line 17: In each method we use DEFAULT_SHIFT
Lets move it here in one place

src/services/safe_border.py
rkolpakov: https://github.com/lidofinance/lido-dao/blob/feature/shapella-upgrade/contracts/0.8.9/WithdrawalQueue.sol#L41

src/services/safe_border.py
rkolpakov: That would not be correct because calculation depends on the value with shift at least here
https://github.com/lidofinance/lido-oracle/blob/3167a4e3d4f0fc3a47c8b78a19139bf471280f39/src/services/safe_border.py#L43

src/services/safe_border.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def _get_archive_lido_validators_by_keys(self, ref_slot, pubkeys) -> list[Validator]:
+        return self.w3.cc.get_validators(ref_slot, tuple(pubkeys))
+
+    def _get_bunker_mode_start_timestamp(self, blockstamp: BlockStamp) -> int:
+        return self.w3.lido_contracts.withdrawal_queue.functions.bunkerModeSinceTimestamp().call(block_identifier=blockstamp.block_hash)
+
+    def _get_last_finalized_withdrawal_request_slot(self, blockstamp: BlockStamp) -> int:
+        last_finalized_request_id = self.w3.lido_contracts.withdrawal_queue.functions.getLastFinalizedRequestId().call(block_identifier=blockstamp.block_hash)
+        last_finalized_request_data = self.w3.lido_contracts.withdrawal_queue.functions.getWithdrawalRequestStatus(last_finalized_request_id).call(block_identifier=blockstamp.block_hash)
```
avsetsin, line 142: need to handle the case when `last_finalized_request_id == 0` since `getWithdrawalRequestStatus` will revert

src/services/safe_border.py
```diff
@@ -0,0 +1,179 @@
...
+            return 0
+
+        sorted_validators = sorted(validators, key = lambda validator: (int(validator.validator.validator.activation_epoch)))
+        return sorted_validators[0].validator.validator.activation_epoch
+
+    def _get_lido_validators(self, blockstamp: BlockStamp) -> list[Validator]:
+        return [lv.validator for lv in self.w3.lido_validators.get_lido_validators(blockstamp.ref_slot)]
+
+    def _get_archive_lido_validators_by_keys(self, ref_slot, pubkeys) -> list[Validator]:
+        return self.w3.cc.get_validators(ref_slot, pubkeys)
```
avsetsin, line 135: not ref_slot, just slot

avsetsin: need to use uncached query

src/modules/accounting/accounting.py
```diff
@@ -134,7 +137,18 @@ def _get_el_vault_balance(self, blockstamp: BlockStamp) -> Wei:
         ))
 
     def _get_last_withdrawal_request_to_finalize(self, blockstamp: BlockStamp) -> int:
-        return 0
+        is_bunker = self._is_bunker(blockstamp)
+        withdrawal_vault_balance = self._get_withdrawal_balance(blockstamp)
+        el_rewards_vault_balance = self._get_el_vault_balance(blockstamp)
+        finalization_share_rate = self._get_finalization_shares_rate(blockstamp)
```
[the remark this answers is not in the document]
rkolpakov: I suppose so, to be more convenient it's easier to call it as a class function not as a param.

PR #194, opened 2023-02-17
Update extra data collection according to the new input args and cont…

skhomuti:
…racts changes


src/modules/accounting/extra_data.py
```diff
@@ -17,22 +16,28 @@
...
 
 class ItemType(Enum):
     EXTRA_DATA_TYPE_STUCK_VALIDATORS = 0
     EXTRA_DATA_TYPE_EXITED_VALIDATORS = 1
 
 
+class FormatList(Enum):
+    EXTRA_DATA_FORMAT_LIST_EMPTY = 0
+    # TODO old contracts have 0, new contracts have 1, don't forget to change it
+    EXTRA_DATA_FORMAT_LIST_NON_EMPTY = 0
```
F4ever, line 38: For now this should be 1?

PR #204, opened 2023-02-24
Add accounting tests

skhomuti:
Fix typing error

### DIFF
--- a/src/modules/accounting/validator_state.py
+++ b/src/modules/accounting/validator_state.py
@@ -108,7 +108,7 @@ class LidoValidatorStateService:
         for module in staking_modules:
             node_operators_ids_in_module = list(map(lambda op: op.id, filter(lambda operator: operator.staking_module.id == module.id, node_operators)))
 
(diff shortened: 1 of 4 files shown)


src/modules/accounting/validator_state.py
```diff
@@ -151,3 +148,9 @@ def _get_oracle_report_limits(self, blockstamp: BlockStamp) -> OracleReportLimit
         orl = named_tuple_to_dataclass(result, OracleReportLimits)
         logger.info({'msg': 'Fetch oracle sanity checks.', 'value': orl})
         return orl
+
+    def _get_last_requested_validator_indices(self, module: StakingModule, node_operators_ids_in_module: list[int]) -> list[int]:
+        return self.w3.lido_contracts.validators_exit_bus_oracle.functions.getLastRequestedValidatorIndices(
+            module.id,
+            node_operators_ids_in_module,
+        ).call()
```
F4ever, line 156: We need to call this function with block_hash

skhomuti: fixed

PR #207, opened 2023-02-25
Feat/ejector

F4ever:


src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class Ejector(BaseModule, ConsensusModule):
+    """
+    1. Get withdrawals amount
+    2. Get exit prediction
+    Loop:
+     a. Get validator
+     b. Remove withdrawal amount
+     c. Increase order
```
TheDZhon, line 35: The comment isn't clear

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+     d. Check new withdrawals epoches
+     e. If withdrawals ok - exit
```
TheDZhon, line 37: Maybe provide more explanations

src/modules/ejector/data_encode.py
```diff
@@ -0,0 +1,29 @@
...
+
+    for (module_id, op_id), validator in validators:
+        result += module_id.to_bytes(MODULE_ID_LENGTH)
+        result += op_id.to_bytes(NODE_OPERATOR_ID_LENGTH)
+        result += int(validator.validator.index).to_bytes(VALIDATOR_INDEX_LENGTH)
+
+        if len(validator.validator.validator.pubkey) != VALIDATOR_PUB_KEY_LENGTH:
+            raise ValueError(f'Unexpected size of validator pub key. Pub key size: {len(validator.validator.validator.pubkey)}')
+
+        result += int(validator.validator.validator.pubkey)
```
TheDZhon, line 27: are we sure that `to_bytes` should be omitted?

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+
+        rewards_speed_per_epoch = self.prediction_service.get_rewards_per_epoch(blockstamp, chain_config)
+        epochs_to_sweep = self._get_sweep_delay_in_epochs(blockstamp)
+
+        to_withdraw_amount = self.get_total_unfinalized_withdrawal_requests_amount(blockstamp)
+        total_current_balance = self._get_total_balance(blockstamp)
+
+        validators_to_eject = []
+        validator_to_eject_balance_sum = 0
+        # ToDo replace it with exit order
```
TheDZhon, line 83: ?

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+        exit_order: list[tuple[NodeOperatorIndex, LidoValidator]] = []
+        for no_index, validator in exit_order:
```
TheDZhon, line 85: Since `exit_order` is empty, `for` loop wouldn't iterate at all

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+            withdrawal_epoch = self._get_withdrawal_epoch_for_latest_validator(blockstamp, len(validators_to_eject) + 1)
+            future_rewards = (blockstamp.ref_slot - (withdrawal_epoch + epochs_to_sweep)) * rewards_speed_per_epoch
```
TheDZhon, line 87: we can't sum `slot` and `epoch` together (they have different dimensions)
more general consideration, we ought to check every 'to be exited validator' to accumulate future rewards
maybe there is a logic error also regarding prediction tracking

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+
+            if future_rewards + total_current_balance + validator_to_eject_balance_sum >= to_withdraw_amount:
+                return validators_to_eject
+
+            validators_to_eject.append((no_index, validator))
+            validator_to_eject_balance_sum += int(validator.validator.validator.effective_balance)
```
TheDZhon, line 93: too much 'validator' %)

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+
+        to_withdraw_amount = self.get_total_unfinalized_withdrawal_requests_amount(blockstamp)
+        total_current_balance = self._get_total_balance(blockstamp)
+
+        validators_to_eject = []
+        validator_to_eject_balance_sum = 0
+        # ToDo replace it with exit order
+        exit_order: list[tuple[NodeOperatorIndex, LidoValidator]] = []
+        for no_index, validator in exit_order:
+            withdrawal_epoch = self._get_withdrawal_epoch_for_latest_validator(blockstamp, len(validators_to_eject) + 1)
```
TheDZhon, line 86: the method has a bit misleading name, could you possibly improve it and add some docs/explanations

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+
+    def get_total_unfinalized_withdrawal_requests_amount(self, blockstamp: BlockStamp) -> Wei:
+        steth_to_finalize = self.w3.lido_contracts.withdrawal_queue_nft.functions.unfinalizedStETH().call(
+            block_identifier=blockstamp.block_hash,
+        )
+        logger.info({'msg': 'Wei to finalize.'})
+        return steth_to_finalize
+
+    def _get_withdrawal_epoch_for_latest_validator(self, blockstamp: BlockStamp, validators_to_eject_count: int) -> EpochNumber:
```
TheDZhon, line 135: maybe we have too large computational complexity, please check out the original spec it it's suitable
https://hackmd.io/PbUr-orDTAqH63F4-n8-pw?view#Exit-epochs

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+                continue
+
+            if val_exit_epoch == max_exit_epoch_number:
+                latest_to_exit_validators_count += 1
+
+            elif val_exit_epoch > max_exit_epoch_number:
+                max_exit_epoch_number = val_exit_epoch
+                latest_to_exit_validators_count = 1
+
+        churn_limit = self._get_churn_limit(blockstamp)
```
TheDZhon, line 152: we call this one too often maybe
because this method is called in the context of `for loop` inside `get_validators_to_eject`
therefore, we need to iterate over all validators each time

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+
+        free_slots_in_current_epoch = churn_limit - latest_to_exit_validators_count
+        need_to_exit_all_epochs = (validators_to_eject_count - free_slots_in_current_epoch) // churn_limit + 1
+
+        return EpochNumber(max_exit_epoch_number + need_to_exit_all_epochs + MIN_VALIDATOR_WITHDRAWABILITY_DELAY)
+
+    def _get_sweep_delay_in_epochs(self, blockstamp: BlockStamp):
+        validators = self.w3.cc.get_validators(blockstamp.state_root)
```
TheDZhon, line 160: can cache all validators in the top-level method

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+
+        def if_validators_balance_withdrawable(validator: Validator):
```
TheDZhon, line 162: maybe we can roll back to the spec's definitions to have a good reference for readers?

TheDZhon: suppose we have a validator with a balance '31 ETH', and '0x01' creds, and `exit_epoch == FAR_FUTURE_EPOCH`
this func would treat this validator as `withdrawable`, while following the genuine spec of CL, this validator has no available skimmable rewards

src/modules/ejector/ejector.py
```diff
@@ -0,0 +1,198 @@
...
+        if report_blockstamp:
+            self.process_report(report_blockstamp)
+
+    @lru_cache(maxsize=1)
+    def build_report(self, blockstamp: BlockStamp) -> tuple:
+        validators = self.get_validators_to_eject(blockstamp)
+
+        data, data_format = encode_data(validators)
+
+        return ReportData(
```
TheDZhon, line 64: maybe we should coordinate and handle the sanity check limits regarding data elements max count (on-chain)

src/modules/ejector/prediction.py
```diff
@@ -0,0 +1,72 @@
...
+
+        events = self._group_events_by_transaction_hash(token_rebase_events, eth_distributed_events)
+
+        if not events:
+            return Wei(0)
+
+        total_rewards = 0
+        time_spent = 0
+        for event in events:
+            total_rewards += event['withdrawalsWithdrawn'] + event['executionLayerRewardsWithdrawn']
```
TheDZhon, line 54: CL rewards?

src/modules/ejector/prediction.py
```diff
@@ -0,0 +1,72 @@
...
+        return Wei(total_rewards * chain_configs.seconds_per_slot * chain_configs.slots_per_epoch // time_spent)
+
+    @staticmethod
+    def _group_events_by_transaction_hash(event_type_1: list[EventData], event_type_2: list[EventData]):
+        result_event_data = []
+
+        for event_1 in event_type_1:
+            for event_2 in event_type_2:
+                if event_2['transactionHash'] == event_1['transactionHash']:
+                    result_event_data.append({
```
TheDZhon, line 66: maybe add sanity checks (their timestamps must be equal), and smth else

PR #214, opened 2023-02-26
Feat/latest oracle v3 review fixes

vgorkavenko:


src/modules/accounting/accounting.py
```diff
@@ -167,8 +167,7 @@ def _get_finalization_shares_rate(self, blockstamp: BlockStamp) -> int:
         simulation = self.get_rebase_after_report(blockstamp)
         return simulation.post_total_pooled_ether * SHARE_RATE_PRECISION_E27 // simulation.post_total_shares
 
-    @lru_cache(maxsize=1)
-    def get_rebase_after_report(self, blockstamp: BlockStamp) -> LidoReportRebase:
+    def get_rebase_after_report(self, blockstamp: BlockStamp, cl_only=False) -> LidoReportRebase:
```
avsetsin, line 170: I would suggest do not have a flag that changes the method behaviour. It could be refactored to smth like this:

```python
def get_full_rebase_after_report(blockstamp):
  wc_balance = self._get_withdrawal_balance(blockstamp)
  el_balance = self._get_el_vault_balance(blockstamp)
  return simulate_oracle_report(wc_balance, el_balance)

def get_cl_rebase_after_report(blockstamp):
  wc_balance = self._get_withdrawal_balance(blockstamp)
  return simulate_oracle_report(wc_balance, el_balance=0)

def simulate_oracle_report(wc_balance, el_balance):
  # ...
  # calls self.w3.lido_contracts.lido.functions.handleOracleReport
```

src/modules/accounting/bunker.py
```diff
@@ -82,18 +80,25 @@ def is_bunker_mode(
...
 
-        self.all_validators, self.lido_keys, self.lido_validators = self._get_lido_validators_with_others(blockstamp)
+        self.all_validators = {
+            v.validator.pubkey: v for v in self.w3.cc.get_validators_no_cache(blockstamp.slot_number)
+        }
+        self.lido_keys = {str(k.key): k for k in self.w3.kac.get_all_lido_keys(blockstamp)}
+        self.lido_validators = {
+            v.validator.pubkey: v
+            for v in self.w3.lido_validators.get_lido_validators(blockstamp)
+        }
```
avsetsin, line 90: since we often reuse different structs with keys or validators, like dicts, list, sets and so on, I would suggest separating them to some readable helpers and reuse them everywhere, respecting the naming convention

avsetsin (approved):
lgtm 👍

PR #216, opened 2023-02-27
feat: tech metrics

vgorkavenko:


src/providers/http_provider.py
```diff
@@ -12,6 +13,11 @@
 logger = logging.getLogger(__name__)
 
 
+def _url_to_request_name_label(url: str) -> str:
+    """Remove all params from url and replace them with {param}"""
+    return '/'.join(['{param}' if ('0x' in part or part.isdigit()) else part for part in url.split('?')[0].split('/')])
```
skhomuti, line 18: I'm not sure about it
URL path may have some credentials like api keys
I think we should add these metrics to the implementations, not a base class

skhomuti (changes requested):

src/providers/http_provider.py
vgorkavenko: `url` arg already contains only API endpoints, without any credentials. Currently, we have only `KeysAPI` and `CL API `providers, which parent is `HttpProvider` and both of them have the same logic for working with `url`

src/metrics/prometheus/basic.py
```diff
@@ -1,76 +1,100 @@
...
 )
 
-ETH2_REQUESTS = Counter(
-    'eth2_rpc_requests',
-    'Total count of requests to ETH2 API',
-    ['method', 'code', 'domain'],
+CL_REQUESTS_DURATION = Histogram(
+    'cl_requests_duration',
+    'Duration of requests to CL API',
+    ['name'],
```
F4ever, line 64:
```suggestion
    ['name', 'code', 'domain'],
```

src/metrics/prometheus/basic.py
```diff
@@ -1,76 +1,72 @@
...
-BUILD_INFO = Gauge(
-    'build_info',
+class Status(Enum):
+    SUCCESS = 'success'
+    FAILURE = 'failure'
+
+
+BUILD_INFO = Info(
+    'build',
     'Build info',
```
F4ever, line 15: ToDo: add build info data

src/metrics/prometheus/task.py
```diff
@@ -0,0 +1,32 @@
+import logging
+from functools import wraps
+from time import perf_counter
+from typing import Callable
+
+from src.metrics.prometheus.basic import TASKS_DURATION, Status
+
```
F4ever, line 7: two lines

src/metrics/prometheus/task.py
```diff
@@ -0,0 +1,32 @@
...
+logger = logging.getLogger(__name__)
+
+
+def task(name: str):
```
F4ever, line 11: duration_meter maybe?
Task is something that you run separately from main process or in another thread

src/metrics/prometheus/task.py
```diff
@@ -0,0 +1,32 @@
...
+                    result = func(*args, **kwargs)
+                    t.labels(name=name, status=Status.SUCCESS)
+                    return result
+                except Exception as e:
+                    t.labels(name=name, status=Status.FAILURE)
+                    raise e
+                finally:
+                    stop = perf_counter()
+                    logger.debug({
+                        "msg": f"Task '{name}' finished", "duration (sec)": max(0, stop - t._start)  # pylint: disable=protected-access
```
F4ever, line 27: Can it be negative?

src/modules/submodules/typings.py
```diff
@@ -8,6 +9,7 @@ class MemberInfo:
     is_report_member: bool
     is_submit_member: bool
     is_fast_lane: bool
+    last_report_ref_slot: Optional[SlotNumber]
```
F4ever, line 12: Not sure it could be None
this is uint64

src/utils/build.py
```diff
@@ -0,0 +1,12 @@
+import json
+
+UNKNOWN_BUILD_INFO = {"version": "unknown", "branch": "unknown", "commit": "unknown"}
+
+
+def get_build_info() -> dict:
+    try:
+        with open("../../build-info.json", "r") as f:
```
F4ever, line 8: It won't work. Because we are running Oracle from root folder.
I suppose smth like this should work
```suggestion
        with open("./build-info.json", "r") as f:
```

src/providers/http_provider.py
```diff
@@ -46,31 +45,35 @@ def _get(self, url: str, params: Optional[dict] = None) -> Tuple[dict | list, di
...
+                t.labels(
+                    name=request_name,
+                    code=response.status_code,
+                    domain=urlparse(self.host).netloc,
+                )
 
         return data, json_response
+
+    @abstractmethod
+    def _url_to_request_name_label(self, url: str) -> str:
```
F4ever, line 78: Maybe just pass original url to _get function
This method wont work for finalized and head

src/web3py/middleware.py
```diff
@@ -29,31 +29,31 @@ def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
...
+
+            # https://www.jsonrpc.org/specification#error_object
+            # https://eth.wiki/json-rpc/json-rpc-error-codes-improvement-proposal
+            error = response.get("error")
+            code: int = 0
+            if isinstance(error, dict):
+                code = error.get("code") or code
+
+            t.labels(
+                name=method,
```
F4ever, line 52: 99% of method_names would be eth_call.
Lets pass here some of params like "to" (address) and "method" method in bytes or name (if it could be done)

src/providers/http_provider.py
vgorkavenko: It's expected behaviour

src/metrics/prometheus/task.py
vgorkavenko: I don't think so, but I prefer to put 0 instead of a negative value, just in case

F4ever: Let's remove this. There is no case it could be negative, so this check is pointless.

src/providers/http_provider.py
vgorkavenko: We can pass to `_get` two separated args - endpoint and params, but I'm not sure that it's a good decision

src/metrics/prometheus/task.py
F4ever: It could be negative if there will be a bug and in this case you will hide this bug

src/modules/accounting/accounting.py
```diff
@@ -7,6 +7,7 @@
 from src.constants import SHARE_RATE_PRECISION_E27
 from src.modules.accounting.typings import ReportData, AccountingProcessingState, LidoReportRebase, \
     SharesRequestedToBurn
```
F4ever, line 9:
```suggestion
from src.modules.accounting.typings import (
    ReportData, 
    AccountingProcessingState, 
    LidoReportRebase,
    SharesRequestedToBurn,
)
```

vgorkavenko: fixed

src/web3py/middleware.py
vgorkavenko: fixed

src/providers/http_provider.py
vgorkavenko: fixed

src/utils/build.py
vgorkavenko: fixed

src/modules/submodules/typings.py
vgorkavenko: fixed

src/metrics/prometheus/task.py
vgorkavenko: fixed

src/metrics/prometheus/task.py
vgorkavenko: fixed

src/metrics/prometheus/task.py
vgorkavenko: fixed

src/web3py/middleware.py
```diff
@@ -18,42 +18,56 @@ def metrics_collector(
...
-                method=method,
-                code=failed.status_code,
+        call_method = ''
+        call_to = ''
+        if hasattr(w3, 'lido_contracts'):
+            if endpoint_name == 'eth_call':
+                args = params[0]
+                call_to = args['to']
+                if contract := w3.lido_contracts.contracts_dict.get(call_to, ''):
+                    call_method = contract.get_function_by_selector(args['data']).fn_name
```
F4ever, line 38: if fn_name is '' send args['data']

src/web3py/middleware.py
```diff
@@ -18,42 +18,56 @@ def metrics_collector(
     """
     Works correctly with MultiProvider and vanilla Providers.
 
-    ETH_RPC_REQUESTS_DURATION - HISTOGRAM with requests time.
-    ETH_RPC_REQUESTS - Counter with requests count, response codes and request domain.
+    EL_REQUESTS_DURATION - HISTOGRAM with requests time, count, response codes and request domain.
     """
 
-    def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
+    def middleware(endpoint_name: RPCEndpoint, params: Any) -> RPCResponse:
```
F4ever, line 24:
```suggestion
    def middleware(method: RPCEndpoint, params: Any) -> RPCResponse:
```

src/utils/build.py
```diff
@@ -0,0 +1,14 @@
...
+import os
+
+UNKNOWN_BUILD_INFO = {"version": "unknown", "branch": "unknown", "commit": "unknown"}
+
+
+def get_build_info() -> dict:
+    try:
+        path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "./build-info.json")
+        with open(path, "r") as f:
+            build_info = json.load(f)
```
F4ever, line 11: except on JSONDecodError

src/web3py/middleware.py
```diff
@@ -18,42 +20,67 @@ def metrics_collector(
...
-    ETH_RPC_REQUESTS_DURATION - HISTOGRAM with requests time.
-    ETH_RPC_REQUESTS - Counter with requests count, response codes and request domain.
+    EL_REQUESTS_DURATION - HISTOGRAM with requests time, count, response codes and request domain.
     """
 
+    contracts = []
+
+    abi_dir = './assets/'
+    for filename in os.listdir(abi_dir):
+        with open(os.path.join(abi_dir, filename), 'r') as f:
```
F4ever, line 30: try except JSONDecodError

src/web3py/middleware.py
```diff
@@ -18,42 +20,67 @@ def metrics_collector(
...
+                    endpoint=method,
+                    call_method=call_method,
+                    call_to=call_to,
+                    code=failed.status_code,
+                    domain=domain,
+                )
+                raise
+
+            # https://www.jsonrpc.org/specification#error_object
+            # https://eth.wiki/json-rpc/json-rpc-error-codes-improvement-proposal
```
F4ever, line 70: Remove this url

tests/providers.py
```diff
@@ -128,12 +129,13 @@ def __init__(self, mock_path: Path, host: str, w3: Web3):
         self.responses = []
         self.from_file = ResponseFromFileHTTPProvider(mock_path, w3)
 
-    def _get(self, url: str, params: Optional[dict] = None) -> dict | list:
+    def _get(self, endpoint: str, path_params: Optional[Sequence[str | int]] = None, query_params: Optional[dict] = None) -> dict | list:
+        url = endpoint.format(*path_params) if path_params else endpoint
         try:
-            response = self.from_file._get(url, params)  # pylint: disable=protected-access
+            response = self.from_file._get(url, query_params)  # pylint: disable=protected-access
```
F4ever, line 135:
```suggestion
            response = self.from_file._get(url, query_params=query_params)  # pylint: disable=protected-access
```

tests/providers.py
```diff
@@ -128,12 +129,13 @@ def __init__(self, mock_path: Path, host: str, w3: Web3):
...
         except NoMockException:
-            response = super()._get(url, params)
-        self.responses.append({"url": url, "params": params, "response": response})
+            response = super()._get(url, query_params)
```
F4ever, line 137:
```suggestion
            response = super()._get(url, query_params=query_params)
```

PR #217, opened 2023-02-27
refactor: move responses to fixtures, clean up conftest

rkolpakov:


src/modules/accounting/typings.py
```diff
@@ -73,3 +74,8 @@ class LidoReportRebase:
     post_total_shares: int
     withdrawals: int
     el_reward: int
+
+@dataclass
+class Account:
+    address: Address
+    _private_key: HexBytes
```
F4ever, line 81: eol

F4ever (approved):

PR #218, opened 2023-02-27
feat: get cl config from node

vgorkavenko:


src/providers/consensus/client.py
```diff
@@ -33,9 +34,33 @@ class ConsensusClient(HTTPProvider):
     API_GET_BLOCK_HEADER = '/eth/v1/beacon/headers/{}'
     API_GET_BLOCK_DETAILS = 'eth/v2/beacon/blocks/{}'
     API_GET_VALIDATORS = 'eth/v1/beacon/states/{}/validators'
+    _API_GET_CONSENSUS_SPEC = 'eth/v1/config/spec'
```
avsetsin, line 37: `_API_GET_CONSENSUS_SPEC` -> `API_GET_CONSENSUS_SPEC`?

src/providers/consensus/client.py
```diff
@@ -33,9 +34,33 @@ class ConsensusClient(HTTPProvider):
...
+
+        We don't use response dataclass init 'as-is' here because new fields may appear and brake struct.
+        We should only take what we need
+
+        No cache because this method is using without any args
+        """
+        data, _ = self._get(self._API_GET_CONSENSUS_SPEC)
+        return ConsensusSpecRequiredOnly(
+            **{k: v for k, v in data.items() if k in ConsensusSpecRequiredOnly.__annotations__}
+        )
```
avsetsin, line 62: - can we cover to `int` here?
- what will happen if one of the required keys does not exist?

tests/modules/accounting/test_bunker.py
```diff
@@ -184,6 +184,17 @@ def bunker(web3, lido_validators) -> BunkerService:
...
+    #     MIN_VALIDATOR_WITHDRAWABILITY_DELAY="256",
+    #     EPOCHS_PER_SLASHINGS_VECTOR="8192",
+    #     PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX="3",
+    #     EFFECTIVE_BALANCE_INCREMENT="1000000000",
+    #     MAX_EFFECTIVE_BALANCE="32000000000",
+    #     MIN_PER_EPOCH_CHURN_LIMIT="4",
+    #     CHURN_LIMIT_QUOTIENT="65536",
+    #     ETH1_ADDRESS_WITHDRAWAL_PREFIX="0x01",
+    #     MAX_WITHDRAWALS_PER_PAYLOAD="16",
+    # )
```
avsetsin, line 197: ?

src/providers/consensus/client.py
vgorkavenko: 1. not all values are numbers. not sure that we need implement some convert logic by types in dataclass
2. exception when creating dataclass object

tests/modules/accounting/test_bunker.py
vgorkavenko: will be fixed when we approve the concept because it brings a lot of issues in tests

src/providers/consensus/client.py
vgorkavenko: it isn't  provided that this method and its endpoint will be called directly from outside, so it is a private attribute of the class

PR #236, opened 2023-03-03
feat: account withdrawal vault balance for reporting

vgorkavenko:


README.md
```diff
@@ -41,6 +41,7 @@ export MEMBER_PRIV_KEY=$ORACLE_PRIVATE_KEY_0X_PREFIXED
 export POOL_CONTRACT=0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84
 export STETH_PRICE_ORACLE_CONTRACT=0x3a6bd15abf19581e411621d669b6a2bbe741ffd6
 export STETH_CURVE_POOL_CONTRACT=0xDC24316b9AE028F1497c275EB9192a3Ea0f67022
+export WITHDRAWAL_VAULT_CONTRACT=0xB9D7934878B5FB9610B3fE8A5e441e8fad7E293f
```
avsetsin, line 44: I would suggest getting the address from Lido contract -> withdrawal credentials

app/metrics.py
```diff
@@ -80,6 +80,10 @@ def get_full_current_metrics(w3, beacon, beacon_spec, partial_metrics) -> PoolMe
     logging.info(
         f'Lido validators\' sum. balance on Beacon: {full_metrics.beaconBalance} wei or {full_metrics.beaconBalance/1e18} ETH'
     )
+
+    block_number = beacon.get_block_by_beacon_slot(slot)
+    partial_metrics.withdrawalVaultBalance = w3.eth.get_balance(withdrawal_vault_address, block_identifier=block_number)
```
avsetsin, line 85: it's better to use blockhash. I guess the issue here, is that we need to include a workaround for web3 for this, right?

README.md
vgorkavenko: fixed

app/metrics.py
vgorkavenko: I would prefer to make minimal changes to the current oracle, and this trade-off (don't patch web3) looks reasonable

avsetsin (approved):

app/metrics.py
```diff
@@ -77,6 +80,14 @@ def get_full_current_metrics(w3, beacon, beacon_spec, partial_metrics) -> PoolMe
         full_metrics.beaconValidators,
         full_metrics.activeValidatorBalance,
     ) = beacon.get_balances(slot, validators_keys)
+
+    block_number = beacon.get_block_by_beacon_slot(slot)
```
TheDZhon, line 84: don't we need to run smth like `get_first_non_missed_slot` from the `oracle-v3` branch in case of a missed block?

TheDZhon (commented):
LGTM, though have one question to double-check and one minor consideration regarding logging.

app/metrics.py
avsetsin: We would not want to make many changes to the current version right now. The current version does not know how to find the previous slot on the CL and collect the report on it, e mean this applies not only to EL, but also to CL blocks, so the current version can't build a report if a block in the ref slot is missed. 

This is an old design bug that will be fixed in version 3

app/metrics.py
```diff
@@ -77,6 +80,14 @@ def get_full_current_metrics(w3, beacon, beacon_spec, partial_metrics) -> PoolMe
...
+    withdrawal_credentials = w3.toHex(pool.functions.getWithdrawalCredentials().call(block_identifier=block_number))
+    full_metrics.beaconBalance += w3.eth.get_balance(
```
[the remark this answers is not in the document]
vgorkavenko: done

vgorkavenko:
I've added a separate metric for withdrawals vault balance. @avsetsin pls, re-review it

F4ever (approved):

avsetsin (approved):

PR #242, opened 2023-03-07
Now extra fields in responses will be ignored when dataclass created

skhomuti:


vgorkavenko (approved):

src/utils/dataclass.py
```diff
@@ -21,26 +21,45 @@ def __post_init__(self):
...
 
-def list_of_dataclasses(_dataclass):
+@dataclass
+class FromResponse:
+    """
+    Class for extending dataclass with custom from_response method, ignored extra fields
+    """
+
+    @classmethod
+    def from_response(cls, **kwargs):
```
F4ever, line 46:
```suggestion
    def from_response(cls, **kwargs) -> 'FromResponse':
```

tests/utils/test_dataclass.py
```diff
@@ -71,3 +72,32 @@ def test_dataclasses_utils_fail_on_unexpected_key():
...
+class Hooman(Nested, FromResponse):
+    favourite_pet: Pet
+    pets: list[Pet]
+
+
+def test_dataclass_nested_with_extra_fields():
+    hooman_response = dict(favourite_pet={"name": "Bob", "age": 5, "extra": "field"},
+                           pets=[{"name": "Bob", "age": 5, "extra": "field"}])
+    hooman = Hooman.from_response(**hooman_response)
+    assert hooman == Hooman(favourite_pet=Pet(name="Bob", age=5), pets=[Pet(name="Bob", age=5)])
```
F4ever, line 103: eol

src/utils/dataclass.py
```diff
@@ -21,26 +21,45 @@ def __post_init__(self):
...
+                setattr(self, field.name, factory(**getattr(self, field.name)))
 
+    @staticmethod
+    def __get_dataclass_factory(field_type):
+        if issubclass(field_type, FromResponse):
+            return field_type.from_response
+        return field_type
 
-def list_of_dataclasses(_dataclass):
+@dataclass
```
F4ever, line 39: two lines between classes

src/utils/abi.py
```diff
@@ -6,11 +6,11 @@ def camel_to_snake(name):
     return re.sub('([a-z0-9])([A-Z])', r'\1_\2', name).lower()
 
 
-def named_tuple_to_dataclass(response, dataclass):
+def named_tuple_to_dataclass(response, dataclass_factory):
```
F4ever, line 9: Why factory? This is still just dataclass

src/utils/dataclass.py
```diff
@@ -21,26 +21,45 @@ def __post_init__(self):
...
+class FromResponse:
+    """
+    Class for extending dataclass with custom from_response method, ignored extra fields
+    """
+
+    @classmethod
+    def from_response(cls, **kwargs):
+        class_field_names = [field.name for field in fields(cls)]
+        return cls(**{k: v for k, v in kwargs.items() if k in class_field_names})
```
F4ever, line 48: This won't work if key will be integer. Maybe parse to str here or ignore int values

src/utils/dataclass.py
```diff
@@ -21,26 +21,45 @@ def __post_init__(self):
...
-def list_of_dataclasses(_dataclass):
+@dataclass
+class FromResponse:
+    """
+    Class for extending dataclass with custom from_response method, ignored extra fields
+    """
+
+    @classmethod
+    def from_response(cls, **kwargs):
+        class_field_names = [field.name for field in fields(cls)]
```
F4ever, line 47: you can use generator here

F4ever:
```suggestion
        class_field_names = (field.name for field in fields(cls))
```

src/utils/abi.py
F4ever, line 9: Add annotations here

src/utils/dataclass.py
skhomuti: not quite

src/utils/dataclass.py
skhomuti: JSON responses cannot contain integer keys. Therefore I assume that we don't receive integer keys here because casting json to dict already parses int keys to str

src/utils/dataclass.py
skhomuti: image
a generator will be empty or lost some items on the next dict comprehension

src/utils/abi.py
skhomuti: added

src/utils/dataclass.py
skhomuti: Added

PR #243, opened 2023-03-07
feat: finalizer checks if WQ contract is on pause

rkolpakov:


pyproject.toml
```diff
@@ -22,6 +22,7 @@ prometheus-client = "^0.16.0"
 web3-multi-provider = "^0.5.0"
 timeout-decorator = "^0.5.0"
 poetry = "^1.3.2"
+hypothesis = "^6.68.2"
```
F4ever, line 25: Lets add it to dev deps

src/services/withdrawal.py
```diff
@@ -28,20 +28,23 @@ def get_next_last_finalizable_id(
         withdrawal_vault_balance: Wei,
         el_rewards_vault_balance: Wei
     ) -> int:
-        if not self._has_unfinalized_requests():
+        is_paused = self._fetch_is_paused()
+        if is_paused or not self._has_unfinalized_requests():
```
F4ever, line 32: Maybe just split it into two checks and add logs for each

tests/modules/accounting/test_withdrawal_unit.py
```diff
@@ -1,10 +1,16 @@
...
 
+from hypothesis import given, strategies, settings, HealthCheck
+from dataclasses import dataclass
 from unittest.mock import Mock
 from src.services.withdrawal import Withdrawal
 from src.modules.submodules.consensus import ChainConfig, FrameConfig
 from tests.conftest import get_blockstamp_by_state
 
+@dataclass
+class Limit:
```
F4ever, line 11: Where we use it?

PR #244, opened 2023-03-07
fix: wf prediction earliest slash epoch wrong condition

rkolpakov:


src/services/safe_border.py
```diff
@@ -111,14 +111,19 @@ def _get_earliest_slashed_epoch_among_incomplete_slashings(self) -> Optional[Epo
...
         exit_epoch = int(validator.validator.exit_epoch)
         withdrawable_epoch = int(validator.validator.withdrawable_epoch)
 
         exited_period = withdrawable_epoch - exit_epoch
-        is_slashed_epoch_undetectable = exited_period > MIN_VALIDATOR_WITHDRAWABILITY_DELAY
+
+        if exited_period < MIN_VALIDATOR_WITHDRAWABILITY_DELAY:
+            raise WrongExitPeriod("exit_epoch and withdrawable_epoch are too close")
+
+        is_slashed_epoch_undetectable = exited_period == MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```
F4ever, line 126:
```suggestion
        is_slashed_epoch_undetectable = exited_period != MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```

F4ever: If I understand correctly tests should fail here, but they passed.
Add to tests this case

rkolpakov: It should be equal so the condition is correct
see [discussion](https://discord.com/channels/761182643269795850/1072446086423257088/1082294765493694514)

F4ever (approved):

PR #248, opened 2023-03-08
Contract call function fix

F4ever:


src/web3py/contract_tweak.py
```diff
@@ -1,14 +1,102 @@
...
+
+    # Fast tweak. Don't need to decode args
+    if decode_tuples and fn_abi["outputs"]:
+        decoded = named_tree(fn_abi["outputs"], normalized_data)
+        normalized_data = recursive_dict_to_namedtuple(decoded)
+
+    if len(normalized_data) == 1:  # pylint: disable=no-else-return
+        return normalized_data[0]
+    else:
+        return normalized_data
```
vgorkavenko, line 99:
```suggestion
    if len(normalized_data) == 1:
        return normalized_data[0]
    return normalized_data
```

skhomuti (approved):

src/web3py/contract_tweak.py
F4ever: This is copy-paste from web3py lib, so I don't want to change here anything, event minor things

F4ever: In future, we just remove this whole code

PR #254, opened 2023-03-09
refactor: fix mypy issues

madlabman:


src/providers/keys/client.py
```diff
@@ -40,6 +40,6 @@ def _get_with_blockstamp(self, url: str, blockstamp: BlockStamp, params: Optiona
 
     @lru_cache(maxsize=1)
     @list_of_dataclasses(LidoKey.from_response)
-    def get_all_lido_keys(self, blockstamp: BlockStamp) -> list[LidoKey]:
+    def get_all_lido_keys(self, blockstamp: BlockStamp) -> list[dict]:
```
skhomuti, line 43: `list_of_dataclasses` affects the return type, the current variant is correct

src/providers/consensus/client.py
```diff
@@ -67,15 +72,19 @@ def get_validators(self, blockstamp: BlockStamp) -> list[Validator]:
         return self.get_validators_no_cache(blockstamp)
 
     @list_of_dataclasses(Validator.from_response)
-    def get_validators_no_cache(self, blockstamp: BlockStamp, pub_keys: Optional[str | tuple] = None) -> list[Validator]:
+    def get_validators_no_cache(self, blockstamp: BlockStamp, pub_keys: Optional[str | tuple] = None) -> list[dict]:
```
skhomuti, line 75: see `list_of_dataclasses`

src/web3py/contract_tweak.py
```diff
@@ -1,20 +1,30 @@
 import itertools
```
skhomuti, line 1: Probably it should be ignored because mostly it's just a copypaste from `web3py`

skhomuti (changes requested):

src/modules/submodules/consensus.py
```diff
@@ -330,7 +328,7 @@ def _get_slot_delay_before_data_submit(self, blockstamp: BlockStamp) -> int:
 
         members, _ = consensus_contract.functions.getMembers().call(block_identifier=blockstamp.block_hash)
 
-        mem_position = members.index(variables.ACCOUNT.address)
+        mem_position = members.index(variables.ACCOUNT.address)  # doesn't supposed to be used without private key provided?
```
F4ever, line 331: Yeah, seems this is bug.
Definitely we should check if ACCOUNT is not None here.

src/providers/keys/client.py
madlabman: `get_all_lido_keys` and similar functions return mappings, which is expected by the `list_of_dataclasses`. `list_of_dataclasses` is a separate function and its return type is inferred from its own typing.

src/web3py/contract_tweak.py
madlabman: Done

src/modules/submodules/consensus.py
madlabman: Done

F4ever (approved):

F4ever:
Just fix mypy before merge

skhomuti (approved):

PR #260, opened 2023-03-10
Accounting refactoring + tests

F4ever:


src/modules/accounting/accounting.py
```diff
@@ -224,11 +225,28 @@ def get_rebase_after_report(self, blockstamp: ReferenceBlockStamp, cl_only=False
...
+
+    def _get_processing_ref_slot(self, blockstamp: BlockStamp):
+        return self.report_contract.functions.getLastProcessingRefSlot().call(block_identifier=blockstamp.block_hash)
+
     @lru_cache(maxsize=1)
     def _is_bunker(self, blockstamp: ReferenceBlockStamp) -> bool:
         frame_config = self.get_frame_config(blockstamp)
         chain_config = self.get_chain_config(blockstamp)
-        cl_rebase_report = self.get_rebase_after_report(blockstamp, cl_only=True)
+        cl_rebase_report = self.get_rebase_after_report(blockstamp, ignore_execution_rewards=True)
```
avsetsin, line 249: A flag that changes the behavior of a function is not a good practice. I would suggest splitting it into several, like: `get_full_rebase_after_report` and `get_cl_rebase_after_report`

src/modules/accounting/accounting.py
```diff
@@ -183,39 +190,33 @@ def _get_finalization_shares_rate(self, blockstamp: ReferenceBlockStamp) -> int:
...
+        blockstamp: ReferenceBlockStamp,
+        ignore_execution_rewards: bool = False,
+    ) -> LidoReportRebase:
+        """
+        To calculate how much withdrawal request protocol can finalize - needs finalization share rate after this report
+        """
         validators_count, cl_balance = self._get_consensus_lido_state(blockstamp)
 
+        chain_conf = self.get_chain_config(blockstamp)
         timestamp = chain_conf.genesis_time + blockstamp.ref_slot * chain_conf.seconds_per_slot
```
avsetsin, line 204: maybe it's better to introduce a method for this `get_timestamp_by_slot(slot, chain_conf)`

src/modules/accounting/accounting.py
```diff
@@ -224,11 +225,28 @@ def get_rebase_after_report(self, blockstamp: ReferenceBlockStamp, cl_only=False
...
+
+        if last_ref_slot:
+            slots_elapsed = blockstamp.ref_slot - last_ref_slot
+        else:
+            slots_elapsed = blockstamp.ref_slot - frame_config.initial_epoch * chain_conf.slots_per_epoch
+
+        return slots_elapsed
+
+    def _get_processing_ref_slot(self, blockstamp: BlockStamp):
+        return self.report_contract.functions.getLastProcessingRefSlot().call(block_identifier=blockstamp.block_hash)
```
avsetsin, line 243: it's also used in bunker and save_border services, better to have is one place

src/modules/accounting/accounting.py
F4ever: These two functions inside will call the same function with the same flag. Not sure if we need such abstraction.
Otherwise, we violate DRY.

src/modules/accounting/accounting.py
F4ever: Fixed

src/modules/accounting/accounting.py
F4ever: fixed

tests/factory/blockstamp.py
```diff
@@ -0,0 +1,30 @@
...
+
+    block_root: BlockRoot = BlockRoot(HexStr('0x8cae2ea12fb6b488225277929e8905b533e3b09491b15d9948949ced9119c6da'))
+    state_root: StateRoot = StateRoot(HexStr('0x623801c28526c1923f14e1bb5258e40a194059c42e280ee61c7189bf2fdbe05e'))
+    slot_number: SlotNumber = SlotNumber(113500)
+    block_hash: BlockHash = BlockHash(HexStr('0x4372578a683ba1c85c259a42492efbe0de9a28b1ac050b5e61065499ab80b0ca'))
+    block_number: BlockNumber = BlockNumber(108006)
+    block_timestamp: Timestamp = Timestamp(0)
+
+    ref_slot: SlotNumber = SlotNumber(113500)
+    ref_epoch: EpochNumber = EpochNumber(113500//12)
```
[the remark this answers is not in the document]
F4ever: fixed

PR #240, opened 2023-03-06
Oracle v3 fixes

avsetsin:
- #235
- #241 
- #238 
- #242 
- #245 
- #255
- #256 
- #259
- #252 
- #257 
- #260
- #263 
- #266
- #270
- #258
- #265
- #267
- #261 
- #279
- #273
- #274
- #290

PR #262, opened 2023-03-10
feat: account shares to burn while calculating shares rate

rkolpakov:


src/modules/accounting/accounting.py
```diff
@@ -179,7 +179,16 @@ def _get_last_withdrawal_request_to_finalize(self, blockstamp: ReferenceBlockSta
     @lru_cache(maxsize=1)
     def _get_finalization_shares_rate(self, blockstamp: ReferenceBlockStamp) -> int:
         simulation = self.get_rebase_after_report(blockstamp)
-        shares_rate = simulation.post_total_pooled_ether * SHARE_RATE_PRECISION_E27 // simulation.post_total_shares
+        cover_shares, non_cover_shares = self.w3.lido_contracts.burner.functions.getSharesRequestedToBurn().call(
```
F4ever, line 182: You should pass _sharesRequestedToBurn on the ref slot to handleOracleReport
(https://github.com/lidofinance/lido-oracle/pull/240#discussion_r1129047356)

F4ever: + handleOraclereport returns now array
https://github.com/lidofinance/lido-oracle/pull/240#discussion_r1129062958

PR #261, opened 2023-03-10
[bunker] chore: refactoring abnormal case

vgorkavenko:

PR #264, opened 2023-03-11
Oracle upgrade to contracts V2

F4ever:
- #258
- #265
- #267
- #261 
- #279


src/services/validator_state.py
```diff
@@ -42,7 +42,7 @@ def get_extra_data(self, blockstamp: ReferenceBlockStamp, chain_config: ChainCon
             stuck_validators=stuck_validators,
             exited_validators=exited_validators,
             max_items_in_payload_count=orl.max_accounting_extra_data_list_items_count,
-            max_items_count=orl.max_accounting_extra_data_list_items_count,
+            max_items_count=orl.max_node_operators_per_extra_data_item_count,
```
avsetsin, line 45: pls double check it, it looks like we need to swap them

F4ever: Fixed

PR #266, opened 2023-03-12
[ejector] check if is paused + tests

madlabman:
Check if paused implemented via withdrawals NFT contract call.


tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+    def test_should_not_report_on_no_withdraw_requests(
+        self,
+        ejector: Ejector,
+        ref_blockstamp: ReferenceBlockStamp,
+        validator_to_exit_it: ValidatorToExitIterator,
+        chain_config: ChainConfig,
+    ) -> None:
+        ejector.get_chain_config = Mock(return_value=chain_config)
+        ejector.get_total_unfinalized_withdrawal_requests_amount = Mock(return_value=0)
+        validator_to_exit_it.v_conf.max_validators_to_exit = 100
```
skhomuti, line 142: It is the default value already

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+        ejector._get_sweep_delay_in_epochs = Mock(return_value=1)
+        ejector._get_total_balance = Mock(return_value=50)
+        ejector.validators_state_service.get_recently_requested_but_not_exited_validators = Mock(
+            return_value=[]
+        )
+
+        result = ejector.get_validators_to_eject(ref_blockstamp)
+        assert result == [], "Unexpected validators to eject"
+
+    # NOTE: not sure if this test makes sense
```
skhomuti, line 171: Why? should it be removed?

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+@pytest.mark.unit
+def test_compute_activation_exit_epoch(
+    ejector: Ejector,
+    ref_blockstamp: ReferenceBlockStamp,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    with monkeypatch.context() as m:
+        m.setattr(ejector_module, "MAX_SEED_LOOKAHEAD", 17)
+        result = ejector.compute_activation_exit_epoch(ref_blockstamp)
+        assert result == 3546 + 17 + 1, "Unexpected activation exit epoch"
```
skhomuti, line 227: Set `ref_epoch` for blockstamp explicitly

skhomuti: and in other tests too if it needed

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+
+
+@pytest.mark.unit
+def test_is_main_data_submitted(ejector: Ejector, blockstamp: BlockStamp) -> None:
+    ejector._get_processing_state = Mock(return_value=Mock(data_submitted=True))
+    assert (
+        ejector.is_main_data_submitted(blockstamp) == True
```
skhomuti, line 234: `is True` everywhere

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+@pytest.mark.usefixtures("consensus_client")
+def test_get_sweep_delay_in_epochs(
+    ejector: Ejector,
+    ref_blockstamp: ReferenceBlockStamp,
+    chain_config: ChainConfig,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    ejector.w3.cc.get_validators = Mock(
+        return_value=[
+            LidoValidatorFactory.build(),
```
skhomuti, line 313: `LidoValidatorFactory.batch(1024)`

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+                "is_active_validator",
+                Mock(side_effect=lambda v, _: bool(v)),
+            )
+            yield
+
+    @pytest.mark.unit
+    @pytest.mark.usefixtures("consensus_client")
+    def test_get_churn_limit_no_validators(self, ejector: Ejector) -> None:
+        ejector.w3.cc.get_validators = Mock(return_value=[])
+        result = ejector._get_churn_limit(ref_blockstamp)
```
skhomuti, line 391: `ref_blockstamp` is a fixture but you don't call it

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+
+class TestChurnLimit:
+    """_get_churn_limit tests"""
+
+    @pytest.fixture(autouse=True)
+    def mock_is_active_validator(self, monkeypatch: pytest.MonkeyPatch) -> Iterable:
+        with monkeypatch.context() as m:
+            m.setattr(
+                ejector_module,
+                "is_active_validator",
```
skhomuti, line 382: Suggest do not mock simple functions like this and create `active_validator` and `non-active_validator` fixtures with overrides activation_epoch and exit_epoch

tests/modules/ejector/test_ejector.py
skhomuti: Because `ejector.w3.cc.get_validators = Mock(return_value=[1, 1, 0]) ` looks strange - the numbers are not a validators

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+        return_value=[
+            Mock(validator=Mock(exit_epoch=999)),
+            Mock(validator=Mock(exit_epoch=42)),
+            Mock(validator=Mock(exit_epoch=42)),
+            Mock(validator=Mock(exit_epoch=1)),
+        ]
+    )
+
+    with monkeypatch.context() as m:
+        m.setattr(ejector_module, "FAR_FUTURE_EPOCH", 999)
```
skhomuti, line 453: You can pass FAR_FUTURE_EPOCH as an exit_epoch for the first validator. Its more obvious than attribute mocking

skhomuti (changes requested):

tests/modules/ejector/test_ejector.py
skhomuti: Because ejector.w3.cc.get_validators = Mock(return_value=[1, 1, 0])  looks strange - the numbers are not a validators

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+
+
+@pytest.fixture()
+def ref_blockstamp() -> ReferenceBlockStamp:
+    return cast(ReferenceBlockStamp, ReferenceBlockStampFactory.build())
+
+
+@pytest.fixture()
+def ejector(web3: Web3, contracts: LidoContracts) -> Ejector:
+    mod = object.__new__(Ejector)
```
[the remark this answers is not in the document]
madlabman: I avoid instantiating underlying services.

tests/modules/ejector/test_ejector.py
madlabman: `ref_blockstamp` is passed as is, and test is not failed on attribute access since no validators are returned.

tests/modules/ejector/test_ejector.py
madlabman: Compute intensive, but may accepted.

tests/modules/ejector/test_ejector.py
madlabman: Defined at fixture level.

tests/modules/ejector/test_ejector.py
```diff
@@ -0,0 +1,460 @@
...
+    return mod
+
+
+@pytest.fixture()
+def validator_to_exit_it(
+    web3,
+    ref_blockstamp: ReferenceBlockStamp,
+    chain_config: ChainConfig,
+) -> ValidatorToExitIterator:
+    it = Mock(spec=ValidatorToExitIterator)
```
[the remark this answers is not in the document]
madlabman: It calls the CL. Don't want it.

tests/modules/ejector/test_ejector.py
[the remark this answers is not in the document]
skhomuti: All of them do nothing except w3 assigning.
Creating with `__new__` looks so fragile, on my view, all classes should be instantiated simply and don't make many things in it

tests/modules/ejector/test_ejector.py
skhomuti: But it looks like a test error. In this case, pass None or mock that causes an Exception if they called. Passing invalid object looks strange

tests/modules/ejector/test_ejector.py
skhomuti: Not defined at this moment. I think we shouldn't expect constant values from global-scoped objects. If the test requires the number, it should be passed here

tests/modules/ejector/test_data_encode.py
```diff
@@ -0,0 +1,225 @@
...
+        length = VALIDATOR_PUB_KEY_LENGTH * 2  # 2 hex digits per byte
+        return HexStr("0x" + "".join(random.choice(symbols) for _ in range(length)))
+
+    return _factory
+
+
+@pytest.mark.unit
+def test_encode_data(pubkey_factory: Callable[[], HexStr]) -> None:
+    def _lido_validator(index: int):
+        return Mock(index=index, validator=Mock(pubkey=pubkey_factory()))
```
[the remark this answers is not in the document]
madlabman: Added new factory with fixed field type.

tests/modules/ejector/test_data_encode.py
```diff
@@ -0,0 +1,225 @@
...
+    ]
+)
+
+
+@pytest.fixture()
+def pubkey_factory() -> Callable:
+    def _factory():
+        symbols = string.hexdigits
+        length = VALIDATOR_PUB_KEY_LENGTH * 2  # 2 hex digits per byte
+        return HexStr("0x" + "".join(random.choice(symbols) for _ in range(length)))
```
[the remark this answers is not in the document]
madlabman: Same as prev.

tests/modules/ejector/test_ejector.py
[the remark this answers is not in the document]
madlabman: Implementation changed, no CL request atm, removed.

tests/modules/ejector/test_ejector.py
madlabman: I described why it works, it is a mistake, apparently.

tests/modules/ejector/test_ejector.py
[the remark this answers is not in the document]
madlabman: There are some runtime checks, but looks acceptable.

tests/modules/ejector/test_ejector.py
madlabman: Moved to the test body.

tests/modules/ejector/test_ejector.py
madlabman: Probably. @rkolpakov wdyt?

tests/modules/ejector/test_ejector.py
madlabman: Nice catch, fixed.

tests/modules/ejector/test_ejector.py
madlabman: > Suggest do not mock simple functions like this and create `active_validator` and `non-active_validator` fixtures with overrides activation_epoch and exit_epoch

`is_active_validator` and other filter-like functions sit in a separate module, so we abstract them.

madlabman: > Because ejector.w3.cc.get_validators = Mock(return_value=[1, 1, 0]) looks strange - the numbers are not a validators

We're interested in the count of the "validator" objects only. I suggest keeping it as is atm.

skhomuti (approved):

src/modules/ejector/ejector.py
```diff
@@ -130,6 +137,12 @@ def get_validators_to_eject(self, blockstamp: ReferenceBlockStamp) -> list[tuple
 
         return validators_to_eject
 
+    @lru_cache(maxsize=1)
+    def _is_paused(self, blockstamp: ReferenceBlockStamp) -> bool:
```
F4ever, line 141: Should it be called with latest blockstamp?
Should we stop sending transactions in this case?

madlabman: `build_report` is run over the given block stamp and cached, so I preserve this behavior.

avsetsin: We should not exit validators if on the reference slot the oracle contract was on pause. We may skip the check on the latest block since a tx will be reverted, but if we can check it and log the result it's better to do it

src/modules/submodules/oracle_module.py
```diff
@@ -20,6 +21,13 @@
 logger = logging.getLogger(__name__)
 
 
+class RetVal(Enum):
```
F4ever, line 24:
```suggestion
class MouduleExecuteDelay(Enum):
```

src/modules/submodules/oracle_module.py
```diff
@@ -20,6 +21,13 @@
...
+    """Signals from execute_module method"""
+
+    WAIT_NEXT_SLOT = 0
+    WAIT_NEXT_FINALIZED_EPOCH = 1
```
F4ever, line 28:
```suggestion
    NEXT_SLOT = 0
    NEXT_FINALIZED_EPOCH = 1
```

src/modules/ejector/ejector.py
```diff
@@ -62,12 +62,15 @@ def __init__(self, w3: Web3):
...
         report_blockstamp = self.get_blockstamp_for_report(last_finalized_blockstamp)
         if report_blockstamp:
+            if self._is_paused(report_blockstamp):
+                logger.info({'msg': 'Ejector is paused. Skip report.'})
+                return RetVal.WAIT_NEXT_FINALIZED_EPOCH
             self.process_report(report_blockstamp)
-            return True
-        return False
+            return RetVal.WAIT_NEXT_SLOT
+        return RetVal.WAIT_NEXT_FINALIZED_EPOCH
```
F4ever, line 73: Let's do wmth like
```python
if no report_blockstamp:
    return RetVal.WAIT_NEXT_FINALIZED_EPOCH
if self._is_paused():
    logger.info({'msg': 'Ejector is paused. Skip report.'})
    return RetVal.WAIT_NEXT_FINALIZED_EPOCH

self.process_report(report_blockstamp)
return RetVal.WAIT_NEXT_SLOT
```

src/modules/ejector/ejector.py
F4ever: Yeah, maybe let's check it via latest_blockstamp, or just do not provide block identifier to call
and remove cache.
Otherwise, we will have in logs a lot of reverted transaction and grafana will be a little bit broken

madlabman: Changed to a call to `isPaused` with no block identifier at the top of `execute_module` method.

PR #268, opened 2023-03-13
feat: business metrics

vgorkavenko:


src/modules/accounting/accounting.py
```diff
@@ -125,6 +134,11 @@ def _calculate_report(self, blockstamp: ReferenceBlockStamp) -> ReportData:
             extra_data_items_count=extra_data.items_count,
         )
 
+        ACCOUNTING_IS_BUNKER.set(report_data.is_bunker)
```
F4ever, line 137: Maybe have separate function where all metrics from report_data will be set?
like
```python
def set_mertrics(data: ReportData):
```

src/modules/accounting/extra_data.py
```diff
@@ -74,10 +75,16 @@ def collect(
         extra_data = self.build_extra_data(stuck_payloads, exited_payloads)
         extra_data_bytes = self.to_bytes(extra_data)
 
-        data_format = FormatList.EXTRA_DATA_FORMAT_LIST_NON_EMPTY if extra_data else FormatList.EXTRA_DATA_FORMAT_LIST_EMPTY
+        if extra_data:
```
F4ever, line 78: pull updates to contracts-v2

src/modules/ejector/ejector.py
```diff
@@ -120,8 +131,11 @@ def get_validators_to_eject(self, blockstamp: ReferenceBlockStamp) -> list[tuple
             c_conf=chain_config,
         )
 
+        withdrawal_epoch = self._get_predicted_withdrawable_epoch(
```
F4ever, line 134: len(validators_to_eject) will increase in each cycle

F4ever: This will break current logic

PR #271, opened 2023-03-15
feat: take max batches count from contract

rkolpakov:


avsetsin (approved):
lgtm

src/services/withdrawal.py
```diff
@@ -31,10 +31,10 @@ def get_finalization_batches(
         withdrawal_vault_balance: Wei,
         el_rewards_vault_balance: Wei
     ) -> list[int]:
-        is_paused = self._fetch_is_paused()
-        has_unfinalized_requests = self._has_unfinalized_requests()
+        if self._fetch_is_paused():
```
F4ever, line 34: Let's add the comment that we are not stopping txs here, we just don't eject

F4ever: I remember that we already discussed, but I don't remember this results of discussion

F4ever: Or we decide to stop tx?

F4ever:
Check tests and there are some conflics

src/services/withdrawal.py
rkolpakov: It's withdrawal finalization not ejecting. In this case we r not stopping tx just finalizing zero requests

PR #274, opened 2023-03-15
[ejector] chore: refactor exit order

vgorkavenko:


src/services/validator_state.py
```diff
@@ -237,7 +242,16 @@ def get_recently_requests_to_exit_pubkeys(
 
         logger.info({'msg': f'Fetch exit events. Got {len(events)} events.'})
 
-        return set(bytes_to_hex_str(event['args']['validatorPubkey']) for event in events)
+        # Initialize dict with empty sets for operators which validators were not contained in any event
```
F4ever, line 245: Why do we need this complication

src/utils/validator_state.py
```diff
@@ -27,6 +28,19 @@ def is_on_exit(validator: Validator) -> bool:
     return int(validator.validator.exit_epoch) != FAR_FUTURE_EPOCH
 
 
+def is_exitable(validator: LidoValidator, last_requested_to_exit_index: int) -> bool:
```
F4ever, line 31: I'm not sure this belongs here.
last_requested_to_exit_index - it's our internal thing, this is business logic.
but here we store utils that can be used anywhere
Also we should mention one important thing that last_requested_to_exit_index works only if this validator is holding by node operator which last_requested_to_exit_index is.

vgorkavenko: That's why here is `LidoValidator` type for `validator` arg

src/services/validator_state.py
vgorkavenko: for the universality of uses

src/services/validator_state.py
```diff
@@ -222,10 +225,12 @@ def non_exited_validators(validator: LidoValidator) -> bool:
 
         return validators_recently_requested_to_exit
 
-    def get_recently_requests_to_exit_pubkeys(
-            self, blockstamp: ReferenceBlockStamp,
-            chain_config: ChainConfig,
-    ) -> set[HexStr]:
+    def get_recently_requests_to_exit_indexes(
```
F4ever, line 228: Rename

F4ever (approved):

PR #292, opened 2023-03-18
Fix: Various important fixes for Oracle

F4ever:
1. Healthcheck port in Docker
2. Contract simulate rebase call
3. TX_GAS_USED to variable


avsetsin (approved):
lgtm

PR #296, opened 2023-03-20
Check keys from Keys API

choooze:
According to the task (https://www.notion.so/ebbe9321e1544043a04679d991a4b3ec?v=fae8607f93fc4299b1fdf0182142fb5c&p=20cbfa76d0104ba98d135b77dfb2e0c6&pm=s) checking that number of keys fetched from the Keys API is >= of `total_deposited_validators` fetched from StakingRouter.
```
    "name": "getAllNodeOperatorDigests",
    "outputs": [
...
              {
                "internalType": "uint256",
                "name": "totalDepositedValidators",
                "type": "uint256"
              },
...
```

Didn't add the same check to `./src/services/bunker_cases/abnormal_cl_rebase.py` where `self.lido_keys = self.w3.kac.get_all_lido_keys(blockstamp)` used.


src/web3py/extensions/lido_validators.py
```diff
@@ -107,6 +111,22 @@ def get_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
         lido_keys = self.w3.kac.get_all_lido_keys(blockstamp)
         validators = self.w3.cc.get_validators(blockstamp)
 
+        no_operators = self.get_lido_node_operators(blockstamp)
+
+        # Make sure that used keys fetched from Keys API >= total amount of total deposited validators from Staking Router
+        total_used_keys = 0
+        for key in lido_keys:
+            if key.used:
+                total_used_keys += 1
```
avsetsin, line 120: Why don't we query only used keys from KAPI instead? We don't use the other ones anywhere

F4ever: Yeah, good point

choooze: Does it mean that we will never get `user: false` key from Keys API or it means that we have to filter keys on provider side (`src/providers/keys/client.py`)?

F4ever: Yeah, pass ?used=false to keys service API

choooze: > Yeah, pass ?used=false to keys service API

`?used=true` I suppose.
OK, got it.

choooze: Done

src/web3py/extensions/lido_validators.py
```diff
@@ -96,6 +96,10 @@ class LidoValidator(Validator):
     lido_id: LidoKey
 
 
+class CountOfKeysDiffersException(Exception):
```
F4ever, line 99: Maybe throw InconsistentData here?

F4ever (approved):

PR #297, opened 2023-03-21
Reload contracts on every slot

skhomuti:
Clear cache if contracts has changed


F4ever (approved):

src/utils/cache.py
```diff
@@ -0,0 +1,5 @@
+
+def clear_object_lru_cache(obj: object):
+    wrappers = [a for a in dir(obj) if hasattr(getattr(obj, a), 'cache_clear')]
```
F4ever, line 3: validator_state.py service, clear cache also

F4ever (approved):

PR #300, opened 2023-03-22
fix: http provider handle timeout error

rkolpakov:


src/providers/http_provider.py
```diff
@@ -54,13 +54,13 @@ def _get(
...
         """
         with self.PROMETHEUS_HISTOGRAM.time() as t:
-            response = self.session.get(
-                self._urljoin(self.host, endpoint.format(*path_params) if path_params else endpoint),
-                params=query_params,
-                timeout=self.REQUEST_TIMEOUT,
-            )
-
             try:
+                response = self.session.get(
```
F4ever, line 58: This won't work, because in this exception there is no handle for Timeouterror

skhomuti (approved):

src/providers/http_provider.py
```diff
@@ -54,11 +54,16 @@ def _get(
...
-                params=query_params,
-                timeout=self.REQUEST_TIMEOUT,
-            )
+            try:
+                response = self.session.get(
+                    self._urljoin(self.host, endpoint.format(*path_params) if path_params else endpoint),
+                    params=query_params,
+                    timeout=self.REQUEST_TIMEOUT,
+                )
+            except Timeout as error:
```
F4ever, line 63: We need to add handle for this exception to BaseModule

skhomuti: common `TimeoutError` re-raised from here and will be handled in base module

F4ever (approved):

PR #303, opened 2023-03-23
[bunker] fix: first round review

vgorkavenko:


src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -156,8 +156,17 @@ def _calculate_cl_rebase_between_blocks(
         To account for all changes in validators' balances, we must account
         withdrawn events from WithdrawalVault contract.
         Check for these events is enough to account for all withdrawals since the protocol assumes that
-        the vault can only be withdrawn at the time of the Oracle report
+        the vault can only be withdrawn at the time of the Oracle report between reference slots.
         """
+        if prev_blockstamp.block_number == ref_blockstamp.block_number:
```
F4ever, line 161: This is private function, all checks we do in _is_negative_specific_cl_rebase and in is_abnormal_cl_rebase
Why do we need this and next one?

vgorkavenko: Because of missing blocks, it may happen that the slots may be equal to

F4ever (approved):

PR #306, opened 2023-03-24
feat: fallbacks for HTTPProvider

vgorkavenko:


src/providers/http_provider.py
```diff
@@ -51,18 +55,52 @@ def _get(
         self, endpoint: str, path_params: Optional[Sequence[str | int]] = None, query_params: Optional[dict] = None
```
F4ever, line 55: Lets rename this to def get

src/providers/http_provider.py
```diff
@@ -51,18 +55,52 @@ def _get(
...
+                        "error": str(e),
+                        "provider": urlparse(host).netloc,
+                    }
+                )
+                error = e
+        msg = f"No active host available for {self.__class__.__name__}"
+        logger.error({"msg": msg})
+        raise NoActiveProviderError(msg) from error
+
+    def _simple_get(
```
F4ever, line 78: And ths to _get

src/providers/http_provider.py
```diff
@@ -12,6 +12,10 @@
 logger = logging.getLogger(__name__)
 
 
+class NoActiveProviderError(Exception):
```
F4ever, line 15: This won't work, NoActiveProviderError in lib and here are different exceptions. This exception wont be catched in BaseOracleModule
maybe just import NoActiveProviderError from multi_http_provider?

PR #309, opened 2023-03-27
[bunker] chore: add docstrings

vgorkavenko:


src/services/bunker.py
```diff
@@ -30,6 +39,12 @@ def is_bunker_mode(
         chain_config: ChainConfig,
         simulated_cl_rebase: LidoReportRebase,
     ) -> bool:
+        """
+        Triggers in 'OR' relation:
```
F4ever, line 43: Triggers if at least one of the following statements is True

src/services/bunker.py
```diff
@@ -30,6 +39,12 @@ def is_bunker_mode(
...
+          - if any of the checks is True, then bunker mode is ON
+          - if all checks are False, then bunker mode is OFF
+        If it is the very first run, we don't check bunker mode
```
F4ever, line 46: Move to code comment

src/services/bunker.py
F4ever, line 43: Check all three cases and one of those are True, bunker mode is On

src/services/bunker.py
vgorkavenko: fixed

F4ever (approved):

PR #314, opened 2023-03-29
Skip reporting if bunker mode is active

skhomuti:


src/modules/accounting/accounting.py
```diff
@@ -122,15 +122,13 @@ def is_extra_data_submitted(self, blockstamp: BlockStamp) -> bool:
         return processing_state.extra_data_submitted
 
     def is_reporting_allowed(self, blockstamp: ReferenceBlockStamp) -> bool:
```
F4ever, line 124: Maybe do this check in Accounting module?
In `def execute_module`?

skhomuti: why? we need to collect maximum report data as we can

src/modules/ejector/ejector.py
```diff
@@ -177,6 +171,15 @@ def get_validators_to_eject(self, blockstamp: ReferenceBlockStamp) -> list[tuple
     def _is_paused(self, blockstamp: ReferenceBlockStamp) -> bool:
         return self.report_contract.functions.isPaused().call(block_identifier=blockstamp.block_hash)
 
+    def is_reporting_allowed(self, blockstamp: ReferenceBlockStamp) -> bool:
+        on_pause = self._is_paused(blockstamp)
+        CONTRACT_ON_PAUSE.set(on_pause)
+        if on_pause:
```
F4ever, line 177: Maybe
`logger.info({'msg': 'Fetch on pause', 'value': on_pause})`
`return not on_pause`

tests/modules/ejector/test_ejector.py
```diff
@@ -70,11 +68,12 @@ def test_ejector_execute_module_on_pause(
     ejector: Ejector, blockstamp: BlockStamp
 ) -> None:
     ejector.get_blockstamp_for_report = Mock(return_value=blockstamp)
+    ejector.build_report = Mock(return_value=(1, 294271, 0, 1, b''))
     ejector._is_paused = Mock(return_value=True)
     assert (
         ejector.execute_module(last_finalized_blockstamp=blockstamp)
-        is ModuleExecuteDelay.NEXT_FINALIZED_EPOCH
-    ), "execute_module should wait for the next finalized epoch"
```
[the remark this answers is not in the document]
F4ever: I would prefer to see Sleep until new finalized block, but we cache everything so should be ok

PR #318, opened 2023-03-30
fix: exception chaining

vgorkavenko:


src/providers/http_provider.py
```diff
@@ -56,21 +55,20 @@ def _get(
         Returns (data, meta) or raises exception
         """
         error = None
-        for host in self.hosts:
+        for i, host in enumerate(self.hosts):
             try:
                 return self._get_without_fallbacks(host, endpoint, path_params, query_params)
             except Exception as e:  # pylint: disable=W0703
```
F4ever, line 61: Not Exception bu ConnectionError

src/providers/http_provider.py
```diff
@@ -55,22 +56,22 @@ def _get(
...
                 logger.warning(
                     {
-                        "msg": "Host not responding.",
+                        "msg": f"Host [{urlparse(host).netloc}] responded with error",
                         "error": str(e),
                         "provider": urlparse(host).netloc,
                     }
                 )
                 error = e
```
F4ever, line 71: NotOkResponse

src/providers/http_provider.py
F4ever: instead of Exception pass all 4 errors

PR #327, opened 2023-04-01
feat(oracle): add a new env var called `MEMBER_PRIV_KEY_FILE`

selfuryon:
Hello!
This is a small PR to allow reading `member_privkey` from a file, which is a standard way to read secrets in Kubernetes.


F4ever:
We are going to release Oracle v3 and dropping support for this version.
Thanks for PR!

PR #328, opened 2023-04-01
feat(oracle-v3): add a new env variable `MEMBER_PRIV_KEY_FILE`

selfuryon:
The same as https://github.com/lidofinance/lido-oracle/pull/327, but for oracle-v3

### DIFF
--- a/src/variables.py
+++ b/src/variables.py
@@ -10,6 +10,12 @@ KEYS_API_URI = os.getenv('KEYS_API_URI', '').split(',')
 # - Account -
 ACCOUNT = None
 MEMBER_PRIV_KEY = os.getenv('MEMBER_PRIV_KEY')
+
+MEMBER_PRIV_KEY_FILE = os.getenv('MEMBER_PRIV_KEY_FILE')
+if MEMBER_PRIV_KEY_FILE is not None:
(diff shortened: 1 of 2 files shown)


src/variables.py
```diff
@@ -10,6 +10,12 @@
 # - Account -
 ACCOUNT = None
 MEMBER_PRIV_KEY = os.getenv('MEMBER_PRIV_KEY')
+
+MEMBER_PRIV_KEY_FILE = os.getenv('MEMBER_PRIV_KEY_FILE')
+if MEMBER_PRIV_KEY_FILE is not None:
```
skhomuti, line 15:
```
if not os.path.exists(MEMBER_PRIV_KEY_FILE): 
    raise ValueError(f"File {MEMBER_PRIV_KEY_FILE} is not found")
```

F4ever: Yeah, let's add more details, like "provided path is not valid" or smth similar

F4ever:
Thanks for PR! We will add this feature to the next pre-release.

src/variables.py
selfuryon: Ok, I'll add today that

PR #334, opened 2023-04-04
Added global lru cache

skhomuti:


F4ever:
Check all files with cache
There are cache in lido_validator.py
Should we clear cache from http clients and web3 also?

skhomuti:
@F4ever I don't think so about the web3 client. It uses a cache with only safe requests that didn't change response in any case

the other http clients already use global_lru_cache

F4ever:
Yeah, that's what I mean. Responses from http CL are safe too, they doesn't depend on contract

PR #340, opened 2023-04-06
Feat: Make contract version check soft

F4ever:


src/modules/submodules/consensus.py
```diff
@@ -196,7 +196,12 @@ def get_blockstamp_for_report(self, last_finalized_blockstamp: BlockStamp) -> Op
         """
         latest_blockstamp = self._get_latest_blockstamp()
 
-        self._check_contract_versions(latest_blockstamp)
+        if self._check_contract_versions(latest_blockstamp):
+            logger.info({
+                'msg': 'Oracle\'s version is higher than contract and/or consensus version. '
```
skhomuti, line 201: it would be great to see versions in logs

F4ever: Done

skhomuti (approved):

PR #341, opened 2023-04-06
Fix: Withdrawal border _get_withdrawal_request_status calls now getWithdrawalStatus method

F4ever:


src/web3py/extensions/tx_utils.py
skhomuti, line 1: it should be in a separate PR

F4ever: Fixed

PR #348, opened 2023-04-08
Handle missed slot correctly

F4ever:


vgorkavenko:
It looks like we should handle the same for `/validators` request

app/beacon.py
```diff
@@ -70,11 +78,26 @@ def __init__(self, url, slots_per_epoch):
...
+
         try:
             return int(response.json()['data']['message']['body']['execution_payload']['block_number'])
         except KeyError as error:
             logging.error(f'Response [{response.status_code}] with text: {str(response.text)} was returned.')
-            raise KeyError from error
+            raise error
+
+    def get_slot_for_report(self, ref_slot: int, epochs_per_frame: int, slots_per_epoch: int):
+        for slot_num in range(ref_slot, ref_slot - epochs_per_frame * slots_per_epoch + 1, -1):
```
avsetsin, line 92:
```suggestion
        for slot_num in range(ref_slot, ref_slot - epochs_per_frame * slots_per_epoch - 1, -1):
```
as it's a decreasing range

F4ever: - 1 would be ours, previous report

avsetsin (approved):
lgtm

app/beacon.py
F4ever: Removed +1 at all

vgorkavenko (approved):

infloop (approved):

PR #352, opened 2023-04-10
Feat/oracle v3 member priv key file

F4ever:


src/variables.py
```diff
@@ -10,6 +10,16 @@
 # - Account -
 ACCOUNT = None
 MEMBER_PRIV_KEY = os.getenv('MEMBER_PRIV_KEY')
+
+MEMBER_PRIV_KEY_FILE = os.getenv('MEMBER_PRIV_KEY_FILE')
+if MEMBER_PRIV_KEY_FILE is not None:
```
avsetsin, line 15: it might be better to allow an empty string. Can be useful in infra (leave a blank value on staging for example)

```suggestion
if not MEMBER_PRIV_KEY_FILE:
```

avsetsin (approved):

avsetsin (approved):

PR #356, opened 2023-04-12
tx no funds fix

F4ever:


src/web3py/extensions/tx_utils.py
```diff
@@ -42,6 +42,9 @@ def _check_transaction(transaction, params: TxParams) -> bool:
         except ContractLogicError as error:
             logger.warning({"msg": "Transaction reverted.", "error": str(error)})
             return False
+        except ValueError as error:
+            logger.error({"msg": "Not enough funds.", "error": str(error)})
```
avsetsin, line 46: is this the only possible cause of the ValueError?

src/web3py/extensions/tx_utils.py
```diff
@@ -79,7 +82,10 @@ def _estimate_gas(transaction: ContractFunction, account: LocalAccount) -> Optio
         try:
             gas = transaction.estimate_gas({'from': account.address})
         except ContractLogicError as error:
-            logger.warning({'msg': 'Contract logic error', 'error': str(error)})
+            logger.warning({'msg': 'Contract logic error.', 'error': str(error)})
+            return None
+        except ValueError as error:
+            logger.warning({'msg': 'Execution reverted.', 'error': str(error)})
```
avsetsin, line 88: is this the only possible cause of the ValueError?

src/web3py/extensions/tx_utils.py
F4ever: We will see details in "error" and if it is another exception we will see it
But for now as I know yes

src/web3py/extensions/tx_utils.py
F4ever: We will see details in "error" and if it is another exception we will see it
But for now as I know yes

PR #374, opened 2023-05-14
fix prediction test

F4ever:


avsetsin (approved):

tests/modules/ejector/test_prediction.py
```diff
@@ -262,9 +262,9 @@ def test_get_rewards_no_matching_events(web3, contracts):
     ]
 
     p = RewardsPredictionService(web3)
-    rewards = p.get_rewards_per_epoch(bp, cc)
 
-    assert rewards == Wei(0)
+    with pytest.raises(ValueError):
```
skhomuti, line 266: now it is an inconsistent events case. We already have tests on it below
In this test case I suppose we should just mock event calls with empty results and receive the same result as expected before (zero wei)

skhomuti (changes requested):

skhomuti (approved):

PR #375, opened 2023-05-14
Match event optimization

F4ever:
Optimize event match (O(n)) and improve consistency check

### DIFF
--- a/src/services/prediction.py
+++ b/src/services/prediction.py
@@ -11,6 +11,10 @@ from src.web3py.typings import Web3
 logger = logging.getLogger(__name__)
 
 
+class InconsistentEvents(Exception):
+    pass
+
+
 class RewardsPredictionService:
     """
     Based on events predicts amount of eth that protocol will earn per epoch.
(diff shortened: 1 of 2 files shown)


avsetsin (approved):
lgtm

src/services/prediction.py
```diff
@@ -63,23 +63,33 @@ def get_rewards_per_epoch(
...
 
-        if len(event_type_1) == len(event_type_2) == len(result_event_data):
-            return result_event_data
+            event_1 = event_type_1_dict.pop(event_2['transactionHash'], None)
 
-        raise ValueError(
-            f"Events are inconsistent: {len(event_type_1)=}, {len(event_type_2)=}, {len(result_event_data)=}"
-        )
+            if not event_1:
+                raise ValueError(f'Events are inconsistent: no events from type 1 with {tx_hash=}.')
```
skhomuti, line 82: suggesting to make this func less common and add some business logic specific to avoid confusing exceptions like this
in general we'd have no idea what the event type 1 or 2 here if they will be raised

skhomuti: and we have literally one call of this func

skhomuti (changes requested):

skhomuti (approved):

avsetsin (approved):

PR #376, opened 2023-05-14
Fix wrong exit validators count

F4ever:
In the function _get_predicted_withdrawable_epoch in the file ejector.py the
maximum value of exit_epoch among validators ( max_exit_epoch_number ) can be less
than blockstamp.ref_epoch + 1 + MAX_SEED_LOOKAHEAD (according to
compute_activation_exit_epoch function). This can happen when all validators have
reached their exit_epochs and none of the validators have initiated an exit yet.
In this case max_exit_epoch_number sets to blockstamp.ref_epoch + 1 +
MAX_SEED_LOOKAHEAD , but at the same time latest_to_exit_validators_count is left
untouched with a non-zero value It leads to the reduction of the capacity of the exit queue.

### DIFF
--- a/src/modules/ejector/ejector.py
+++ b/src/modules/ejector/ejector.py
@@ -250,10 +250,11 @@ class Ejector(BaseModule, ConsensusModule):
         """
         max_exit_epoch_number, latest_to_exit_validators_count = self._get_latest_exit_epoch(blockstamp)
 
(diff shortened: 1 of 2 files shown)


avsetsin (approved):
lgtm

skhomuti (approved):

PR #377, opened 2023-05-14
Tweak prometheus buckets

F4ever:
Some functions work more than 5 minutes, so we don't see accurate prometheus histogram, so we increased buckets for requests and functions.


avsetsin (approved):
lgtm

skhomuti (approved):

PR #397, opened 2023-10-03
Feat: Upgrade oracle's deps

F4ever:


src/web3py/extensions/contracts.py
```diff
@@ -17,6 +17,8 @@
 
 
 class LidoContracts(Module):
+    w3: Web3
```
F4ever, line 20: For linters.
New version of web3 modules contains smth like
w3: Unify[Web3, AsyncWeb3]

vgorkavenko (approved):

skhomuti (approved):

Issue #389, opened 2023-06-09
Unhandled exception in oracle transaction submission


tinchoabbate:
Hey folks, I think there's a case of an unhandled exception that might lead to unexpectedly halting the oracle service. It doesn't look worrisome, so I'm reporting it publicly here for you to take a closer look.

Every time an oracle sends a tx (for example to submit report data), execution goes through [`check_and_send_transaction`](https://github.com/lidofinance/lido-oracle/blob/44678954915b8291c949904c63de5e4e4983b427/src/web3py/extensions/tx_utils.py#L18). This function then calls [`_sign_and_send_transaction`](https://github.com/lidofinance/lido-oracle/blob/44678954915b8291c949904c63de5e4e4983b427/src/web3py/extensions/tx_utils.py#L29). Which, after [sending](https://github.com/lidofinance/lido-oracle/blob/44678954915b8291c949904c63de5e4e4983b427/src/web3py/extensions/tx_utils.py#L115) the raw transaction, waits for the transaction receipt [calling](https://github.com/lidofinance/lido-oracle/blob/44678954915b8291c949904c63de5e4e4983b427/src/web3py/extensions/tx_utils.py#L118) `wait_for_transaction_receipt`.

The issue is that, as far as I understand from its [documentation](https://web3py.readthedocs.io/en/v6.0.0/web3.eth.html#web3.eth.Eth.wait_for_transaction_receipt) and [code](https://github.com/ethereum/web3.py/blob/v6.0.0-beta.10/web3/eth/eth.py#L456), `wait_for_transaction_receipt` throws a `TimeExhausted` exception after 120 seconds of polling for the receipt. This exception is not handled anywhere in the code of the oracle. My impression is that the oracle's code _attempts_ to handle the case of not finding a receipt with [this if statement](https://github.com/lidofinance/lido-oracle/blob/44678954915b8291c949904c63de5e4e4983b427/src/web3py/extensions/tx_utils.py#L120-L123), but I don't think that's the correct way of doing it, because it won't catch the exception.

If I'm following this correctly, then this would happen in the unlikely scenario of oracles trying to submit their txs when the network's is congested. In any case, feels like this exception could be handled in a safer way to make the code more robust.

I might be missing something here, so of course any thoughts welcome!

PR #399, opened 2023-10-03
Fix: Transaction submission unhandled exception

F4ever:
https://github.com/lidofinance/lido-oracle/issues/389

### DIFF
--- a/src/web3py/extensions/tx_utils.py
+++ b/src/web3py/extensions/tx_utils.py
@@ -3,14 +3,16 @@ from typing import Optional
 
 from eth_account.signers.local import LocalAccount
 from web3 import Web3
+from hexbytes import HexBytes
 from web3.contract.contract import ContractFunction
-from web3.exceptions import ContractLogicError
+from web3.exceptions import ContractLogicError, TimeExhausted
 from web3.module import Module
 from web3.types import TxReceipt, Wei, TxParams, BlockData
 
 from src import variables, constants
 from src.metrics.prometheus.basic import TRANSACTIONS_COUNT, Status
 from src.utils.input import prompt
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/web3py/extensions/tx_utils.py
```diff
@@ -118,9 +120,12 @@ def _sign_and_send_transaction(
         tx_hash = self.w3.eth.send_raw_transaction(signed_tx.rawTransaction)
         logger.info({"msg": "Transaction sent.", "value": tx_hash.hex()})
 
-        tx_receipt = self.w3.eth.wait_for_transaction_receipt(tx_hash)
+        return self._handle_sent_transaction(tx_hash)
 
-        if not tx_receipt:
+    def _handle_sent_transaction(self, transaction_hash: HexBytes) -> Optional[TxReceipt]:
+        try:
+            tx_receipt = self.w3.eth.wait_for_transaction_receipt(transaction_hash)
```
vgorkavenko, line 127: `tx_receipt` can still be`None` (by the library code). Let's handle this as well

vgorkavenko (changes requested):

src/web3py/extensions/tx_utils.py
F4ever: As I see it can't
```
    def wait_for_transaction_receipt(
        self, transaction_hash: _Hash32, timeout: float = 120, poll_latency: float = 0.1
    ) -> TxReceipt:
        try:
            with Timeout(timeout) as _timeout:
                while True:
                    try:
                        tx_receipt = self._transaction_receipt(transaction_hash)
                    except TransactionNotFound:
                        tx_receipt = None
                    if tx_receipt is not None:
                        break
                    _timeout.sleep(poll_latency)
            return tx_receipt

        except Timeout:
            raise TimeExhausted(
                f"Transaction {HexBytes(transaction_hash) !r} is not in the chain "
                f"after {timeout} seconds"
            )
``

vgorkavenko (approved):

skhomuti (approved):

PR #400, opened 2023-10-03
Fix: unhealthy container

F4ever:


vgorkavenko (changes requested):
It shouldn't be like this. `Pulse` is needed for using by docker healthcheck because it is like a simple approach to ensure that everything is fine with application, so it shouldn't be requested inside a code (otherwise we obtain a recursive dependence).

Let's fix `Dockerfile` and nothing else:
From:
```
CMD curl -f http://localhost:$HEALTHCHECK_SERVER_PORT/healthcheck || exit 1
```
To:
```
CMD curl -f http://localhost:$HEALTHCHECK_SERVER_PORT/pulse || exit 1
```

F4ever:
@vgorkavenko 
No pulse function just do `/pulse/` request to server that contains last pulse datetime.
All rest requests to this server just returns is last_puls_datetime expired or not

vgorkavenko (approved):

vgorkavenko:
> @vgorkavenko No pulse function just do `/pulse/` request to server that contains last pulse datetime. All rest requests to this server just returns is last_puls_datetime expired or not

Yes. Thanks for the explanation!

src/modules/submodules/oracle_module.py
```diff
@@ -110,6 +111,10 @@ def run_cycle(self, blockstamp: BlockStamp) -> ModuleExecuteDelay:
             logger.error({'msg': 'Keys API service returned incorrect number of keys.', 'error': str(error)})
         except ValueError as error:
             logger.error({'msg': 'Unexpected error.', 'error': str(error)})
+        else:
+            # Do pulse only if no exceptions only
```
skhomuti, line 115: double "only" typo

skhomuti (approved):

src/modules/submodules/oracle_module.py
F4ever: fixed

PR #401, opened 2023-10-03
Feat: Add more logs to ejector module

F4ever:


vgorkavenko (dismissed):

src/modules/ejector/ejector.py
```diff
@@ -296,6 +295,12 @@ def _get_latest_exit_epoch(self, blockstamp: BlockStamp) -> tuple[EpochNumber, i
                 max_exit_epoch_number = val_exit_epoch
                 latest_to_exit_validators_count = 1
 
+        logger.info({
+            'msg': 'Calculate latest exit epoch',
```
skhomuti, line 299: in the `_get_predicted_withdrawable_epoch` this value can be changed to 0, and this message will be misleading

suggest to put it later up in the `_get_predicted_withdrawable_epoch` function

skhomuti (changes requested):

src/modules/ejector/ejector.py
F4ever: It's true, but `_get_predicted_withdrawable_epoch` called many time in cycle. And `_get_latest_exit_epoch` called once and actual for current blockchain state. We always can calculate `compute_exit_epoch`, because it is `ref_slot + some_constant`

src/modules/ejector/ejector.py
```diff
@@ -261,8 +261,7 @@ def _get_predicted_withdrawable_epoch(
 
         churn_limit = self._get_churn_limit(blockstamp)
 
-        remain_exits_capacity_for_epoch = churn_limit - latest_to_exit_validators_count
-        epochs_required_to_exit_validators = (validators_to_eject_count - remain_exits_capacity_for_epoch) // churn_limit + 1
+        epochs_required_to_exit_validators = (validators_to_eject_count + latest_to_exit_validators_count) // churn_limit
```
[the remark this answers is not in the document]
F4ever: I just opened the brackets and simplified the expression)

skhomuti (approved):

PR #413, opened 2023-10-23
Feat: add more logs to code and log validators request in ejector module

F4ever:


src/modules/ejector/ejector.py
```diff
@@ -105,6 +106,19 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
...
+                f.write(
+                    json.dumps([{
+                        **v.__dict__,
+                        'validator': v.validator.__dict__,
+                    } for v in self.w3.cc.get_validators(blockstamp)])
+                )
+        except Exception as error:  # pylint: disable=W0703
+            logger.error({'msg': 'Failed to save get_validators response.', 'error': str(error)})
+        else:
+            logger.info({'msg': 'Response get_validators from Consensus Cline wrote to file validators_response.json.'})
```
skhomuti, line 120:
```suggestion
            logger.info({'msg': 'Response get_validators from Consensus Client written to validators_response.json.'})
```

src/modules/ejector/ejector.py
```diff
@@ -105,6 +106,19 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
 
         EJECTOR_VALIDATORS_COUNT_TO_EJECT.set(report_data.requests_count)
 
+        try:
+            with open('validators_response.json', 'w') as f:
+                f.write(
+                    json.dumps([{
+                        **v.__dict__,
+                        'validator': v.validator.__dict__,
```
skhomuti, line 114: it works like a charm including nested dataclasses too
```
from dataclasses import asdict 

asdict(v)
```

src/modules/ejector/ejector.py
```diff
@@ -261,8 +275,7 @@ def _get_predicted_withdrawable_epoch(
 
         churn_limit = self._get_churn_limit(blockstamp)
 
-        remain_exits_capacity_for_epoch = churn_limit - latest_to_exit_validators_count
-        epochs_required_to_exit_validators = (validators_to_eject_count - remain_exits_capacity_for_epoch) // churn_limit + 1
+        epochs_required_to_exit_validators = (validators_to_eject_count + latest_to_exit_validators_count) // churn_limit
```
skhomuti, line 278: still not sure whether it really needs to change

skhomuti (changes requested):

src/modules/ejector/ejector.py
F4ever: The result is the same, I just simplified the expression

F4ever: you can do it on paper and check it

src/modules/ejector/ejector.py
F4ever: lol, it's very slow.
My method works 0.6107 seconds
asdict works 9.8773 seconds
with same input data (500 MB of validators)

skhomuti (dismissed):

vgorkavenko (dismissed):

src/modules/ejector/ejector.py
infloop: I'm here against any changes in logic, even simplification of expressions

vgorkavenko (approved):

infloop (approved):

PR #426, opened 2024-04-11
feat: attestation and committees endpoints

vgorkavenko:


src/providers/consensus/typings.py
```diff
@@ -108,3 +108,31 @@ class BlockDetailsResponse(Nested, FromResponse):
     # https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockV2
     message: BlockMessage
     signature: str
+
+
+@dataclass
+class AttestationSourceTargetData(FromResponse):
```
madlabman, line 114: Why not Checkpoint as of spec?

src/providers/http_provider.py
```diff
@@ -110,8 +111,9 @@ def _get_without_fallbacks(
         host: str,
         endpoint: str,
         path_params: Optional[Sequence[str | int]] = None,
-        query_params: Optional[dict] = None
-    ) -> Tuple[dict | list, dict]:
+        query_params: Optional[dict] = None,
+        stream: bool = False
```
madlabman, line 115: `json = True` maybe?

vgorkavenko: just to be aligned with `requests` lib

src/providers/consensus/typings.py
vgorkavenko: hz, I'll change it

vgorkavenko: fixed

src/providers/consensus/client.py
```diff
@@ -1,5 +1,8 @@
 from http import HTTPStatus
-from typing import Literal, Optional, Union
+from typing import Literal, Optional, Union, Iterator, cast, Tuple
+
+import json_stream.requests  # type: ignore
```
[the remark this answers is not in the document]
vgorkavenko: it was auto-import. will change that if it's mandatory

src/providers/consensus/client.py
```diff
@@ -74,11 +80,11 @@ def get_block_root(self, state_id: Union[SlotNumber, BlockRoot, LiteralState]) -
     @lru_cache(maxsize=1)
     def get_block_header(self, state_id: Union[SlotNumber, BlockRoot]) -> BlockHeaderFullResponse:
         """Spec: https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockHeader"""
-        data, meta_data = self._get(
+        data, meta_data = cast(Tuple[dict, dict], self._get(
```
skhomuti, line 83: you probably may use native types like `cast(tuple[dict, dict])`
but why this cast is here at all?

src/providers/consensus/client.py
```diff
@@ -96,6 +102,56 @@ def get_block_details(self, state_id: Union[SlotNumber, BlockRoot]) -> BlockDeta
...
+        for att in data:
+            yield BlockAttestation.from_response(**att)
+
+    @lru_cache(maxsize=1)
+    def get_attestation_committees(
+        self,
+        blockstamp: BlockStamp,
+        epoch: Optional[EpochNumber] = None,
+        index: Optional[int] = None,
+        slot: Optional[SlotNumber] = None
```
skhomuti, line 124: modern syntax is like `epoch: EpochNumber | None = None`

madlabman (approved):

src/providers/http_provider.py
```diff
@@ -110,8 +111,9 @@ def _get_without_fallbacks(
...
+    ) -> Tuple[dict | list, dict] | Response:
```
skhomuti, line 116: this return value like from different abstraction level
I'd prefer to process streaming here as it is with usual json response

skhomuti: you also don't need cast everywhere in that case

skhomuti (changes requested):

src/providers/consensus/client.py
vgorkavenko: because of mypy issue https://github.com/python/mypy/issues/11692#issuecomment-994746538

src/providers/http_provider.py
vgorkavenko: I would prefer to give the ability of handling the stream to the child implementations

PR #289, opened 2023-03-17
Feat/business metrics

vgorkavenko:

PR #428, opened 2024-04-11
feat: add CSM performance oracle module

madlabman:
Just initialization of the module and some blueprint code.


src/modules/csm/csm.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+
+        for mod in modules:
+            if mod.name == "": # FIXME
+                return mod
+
+        raise ValueError("No CSM module found")
+
+    @lru_cache(maxsize=1)
+    def module_validators_by_node_operators(self, blockstamp: BlockStamp) -> ValidatorsByNodeOperator:
```
vgorkavenko, line 101: `TODO: should be better`

src/modules/csm/typings.py
```diff
@@ -0,0 +1,37 @@
...
+@dataclass
+class AttestationsAggregate:
+    assigned: int
+    included: int
+
+@dataclass
+class PerformanceCache:
+    """Data structure to store required data for performance calculation."""
+
+    slot_to_duty: dict[SlotNumber, tuple[ValidatorIndex, CommitteeIndex, int]]
```
vgorkavenko, line 33: is it the current duty under processing?

src/web3py/extensions/contracts.py
```diff
@@ -154,10 +173,20 @@ def get_accounting_last_processing_ref_slot(self, blockstamp: BlockStamp) -> Slo
         FRAME_PREV_REPORT_REF_SLOT.labels('accounting').set(result)
         return result
 
+    # TODO: Why here's no cache applied?
```
vgorkavenko, line 176: it was done explicitly 🤔  https://github.com/lidofinance/lido-oracle/pull/289#discussion_r1148337913

vgorkavenko: might be because it is used only in one particular place rather than the accounting thing

.env.example
```diff
@@ -2,4 +2,6 @@ CONSENSUS_CLIENT_URI=http://...
 EXECUTION_CLIENT_URI=http://...
 KEYS_API_URI=https://...
 LIDO_LOCATOR_ADDRESS=0x1...
-MEMBER_PRIV_KEY=aaa...
\ No newline at end of file
+CSM_ORACLE_ADDRESS=...
```
F4ever, line 5: It won't be in Locator?

src/web3py/extensions/contracts.py
F4ever: 🤔

F4ever: Yeah, it's called once. Only to collect metrics.

vgorkavenko (approved):

PR #436, opened 2024-04-23
feat: add dummy IPFS provider

madlabman:


src/providers/ipfs/dummy.py
```diff
@@ -0,0 +1,33 @@
...
+
+from web3.module import Module
+
+from .types import CIDv0, CIDv1, IPFSProvider, NotFound
+
+
+class DummyIPFSProvider(IPFSProvider, Module):
+    """Dummy IPFS provider which using the local filesystem as a backend"""
+
+    mempool: dict[CIDv0 | CIDv1, bytes]
```
vgorkavenko, line 11: `cache`? mislead a little at first glance

madlabman: It's in-memory pool of content for this provider, that's why the name.

PR #437, opened 2024-04-23
feat: build CSM oracle report

madlabman:


src/modules/csm/csm.py
```diff
@@ -61,22 +61,41 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
...
+
+            distribution[no_id] = share
+            total += share
 
         # Calculate share of each CSM node operator.
-        _ = self._to_distribute(blockstamp)
-        shares: tuple[tuple[NodeOperatorId, int]] = tuple()  # type: ignore
+        to_distribute = self.w3.csm.fee_distributor.pending_to_distribute(blockstamp.block_hash)
+        shares: list[tuple[NodeOperatorId, int]] = []
+        for no_id, share in distribution.items():
```
vgorkavenko, line 93: share of distribution by performance intersects with stETH shares. but I can't come up with better names here. wdyt?

src/modules/csm/csm.py
```diff
@@ -61,22 +61,41 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
...
+            shares.append((no_id, to_distribute * share // total))
+
+        distributed = sum((s for (_, s) in shares))
```
vgorkavenko, line 96: `total_distributed` and having increment in the cycle above?

src/modules/csm/csm.py
```diff
@@ -61,22 +61,41 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
...
+        if not distributed:
+            ...  # TODO: The code expects the report built, but it doesn't make sense.
```
vgorkavenko, line 98: it is possible if we have the first report, but the validators are still pending to activate. they will have zero performance and due to that nothing to distribute for them. so it is legit case for the report. we can logging it like a warn I guess

src/providers/execution/contracts/CSFeeOracle.py
```diff
@@ -33,6 +33,19 @@ def fee_distributor(self, block: BlockIdentifier = "latest") -> Address:
...
+
+        resp = self.functions.perfThresholdBP().call(block_identifier=block)
+        logger.debug(
+            {
+                "msg": "Call to perfThresholdBP()",
+                "value": resp,
+                "block_identifier": repr(block),
+            }
+        )
+        return resp / 10_000  # Convert from basis points
```
vgorkavenko, line 47: take `MAX_BP` from the contract as well

src/web3py/extensions/lido_validators.py
```diff
@@ -174,7 +174,9 @@ def get_lido_validators_by_node_operators(self, blockstamp: BlockStamp) -> Valid
     @lru_cache(maxsize=1)
     def get_module_validators_by_node_operators(self, module_id: StakingModuleId, blockstamp: BlockStamp) -> ValidatorsByNodeOperator:
         """Get module validators by querying the KeysAPI for the module keys"""
-        raise NotImplementedError()
+        # TODO: Re-evaluate the implementation, sub-optimal solution.
```
vgorkavenko, line 177: very sub-optimal, I'd say. we have 1% CSM validators and 99% from other modules, even on testnet. so I'd suggest to do the plain request(s) to KeysAPI here to get the validators + operators by the module address

src/providers/execution/contracts/CSFeeOracle.py
madlabman: It's internal and well known constant.

src/modules/csm/csm.py
madlabman: Yeah, just a first approach.

src/modules/csm/csm.py
madlabman: Looks better in this way.

vgorkavenko (approved):

PR #445, opened 2024-05-07
feat: `epochs_to_process`

vgorkavenko:


src/modules/csm/state.py
```diff
@@ -75,13 +94,50 @@ def status(self) -> None:
...
+            self.clear()
+            self.commit()
+
+        try:
+            self.validate(l_epoch, r_epoch)
+        except InvalidStateLeftBorder:
+            _discard()
+        except InvalidStateRightBorder:
+            if self.r_epoch < r_epoch:
+                _discard()
```
madlabman, line 126: It should be `self.r_epoch > r_epoch` instead.

src/modules/csm/state.py
```diff
@@ -75,13 +94,50 @@ def status(self) -> None:
...
+            else:
+                logger.warning({"msg": "The last report was missed. Reuse the state with the new right border"})
+                # expand the state to the new right border
+                self.epochs_to_process.update(sequence(self.r_epoch - 1, r_epoch))
```
madlabman, line 130: `self.r_epoch` as well as `self.r_epoch - 1` could be already processed, so `self.r_epoch + 1`, I think.

src/modules/csm/state.py
```diff
@@ -75,13 +94,50 @@ def status(self) -> None:
...
+
+        if self.r_epoch > 0 and self.r_epoch != r_epoch:
+            logger.warning({"msg": f"Invalid right state border: should be {r_epoch=} but {self.r_epoch=}"})
+            raise InvalidStateRightBorder()
+
+    def validate_and_adjust(self, l_epoch: EpochNumber, r_epoch: EpochNumber):
+
+        def _discard():
+            logger.warning({"msg": "Discarding invalidated state cache"})
+            self.clear()
```
madlabman, line 117: `assert is_empty` inside the `clear` method would be nice.

src/modules/csm/state.py
```diff
@@ -26,9 +39,10 @@ def perf(self) -> float:
 @dataclass
 class State(UserDict[ValidatorIndex, AttestationsAggregate]):
     """Tracks processing state of CSM performance oracle frame"""
-
+    l_epoch: EpochNumber = field(default_factory=int)
+    r_epoch: EpochNumber = field(default_factory=int)
     data: dict[ValidatorIndex, AttestationsAggregate] = field(default_factory=dict)
-    processed_epochs: set[EpochNumber] = field(default_factory=set)
+    epochs_to_process: set[EpochNumber] = field(default_factory=set)
```
madlabman, line 45: The thing I'm concerned the most is that we have no metric to check the state as it was using `processed_epochs`.

src/modules/csm/state.py
```diff
@@ -75,13 +94,50 @@ def status(self) -> None:
...
+    def validate(self, l_epoch: EpochNumber, r_epoch: EpochNumber) -> None:
+        if self.l_epoch > 0 and self.l_epoch != l_epoch:
+            logger.warning({"msg": f"Invalid left state border: should be {l_epoch=} but {self.l_epoch=}"})
+            raise InvalidStateLeftBorder()
+
+        if self.r_epoch > 0 and self.r_epoch != r_epoch:
+            logger.warning({"msg": f"Invalid right state border: should be {r_epoch=} but {self.r_epoch=}"})
+            raise InvalidStateRightBorder()
+
+    def validate_and_adjust(self, l_epoch: EpochNumber, r_epoch: EpochNumber):
```
madlabman, line 113: `migrate` or something similar?

madlabman (changes requested):

src/modules/csm/state.py
vgorkavenko: not sure. left as is for a while

PR #452, opened 2024-05-20
feat: fetch stuck events for the given frame only

madlabman:


src/providers/execution/contracts/CSModule.py
```diff
@@ -1,72 +1,92 @@
...
         super().__init__(address)
 
-    def get_stuck_node_operators(self, l_block: BlockIdentifier, r_block: BlockIdentifier) -> Iterable:
+    # TODO: Consider adding cache.
+    def current_stuck_operators_ids(self, block: BlockIdentifier = "latest") -> Iterable[NodeOperatorId]:
+        # TODO: Check performance on a large amount of node operators in a module.
+        if self.is_deployed(block):
+            for no_id in self.node_operators_ids(block):
+                if self.node_operator_summary(no_id, block).stuckValidatorsCount > 0:
+                    yield no_id
```
vgorkavenko, line 45:
```
if not self.is_deployed(block):
   return
```
looks better I think

src/providers/execution/contracts/CSModule.py
```diff
@@ -1,72 +1,92 @@
...
 
     # TODO: Inherit from the base class.
     def __init__(self, address: ChecksumAddress | None = None) -> None:
         with open(self.abi_path, encoding="utf-8") as f:
             self.abi = json.load(f)
         super().__init__(address)
 
-    def get_stuck_node_operators(self, l_block: BlockIdentifier, r_block: BlockIdentifier) -> Iterable:
+    # TODO: Consider adding cache.
+    def current_stuck_operators_ids(self, block: BlockIdentifier = "latest") -> Iterable[NodeOperatorId]:
```
vgorkavenko, line 40: `current...` sounds a little misleading. can it be just `get_stuck_operators_ids` ?

src/providers/execution/contracts/CSModule.py
```diff
@@ -1,72 +1,92 @@
...
-        events = sorted(self.get_stuck_keys_events(r_block_number), key=by_no_id)
+        events = sorted(self.get_stuck_keys_events(l_block_number, r_block_number), key=by_no_id)
         for no_id, group in groupby(events, key=by_no_id):
-            last_event = sorted(tuple(group), key=lambda e: e["blockNumber"])[-1]
+            last_event = sorted(tuple(group), key=by_block)[-1]
             # Operators unstucked at the very beginning of the frame are fine.
             if (
-                last_event["args"]["stuckValidatorsCount"] == 0 and
+                last_event["args"]["stuckKeysCount"] == 0 and
                 last_event["blockNumber"] <= l_block_number
```
vgorkavenko, line 63: `last_event["blockNumber"] == l_block_number`
should be enough, I guess

vgorkavenko: btw the case is so rare that I think we don't need this branch in code at all

madlabman: No, we have to check for the case. But, I've changed the flow a little, so this check is swallowed here.

vgorkavenko (approved):

PR #456, opened 2024-05-22
Close some TODOs

madlabman:


src/modules/csm/csm.py
```diff
@@ -227,10 +227,14 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
 
         start = time.time()
         for checkpoint in checkpoints:
-            # TODO: Check that we still need to check these checkpoints.
+            if self.current_frame_range(self._receive_last_finalized_slot()) != (l_ref_slot, r_ref_slot):
+                logger.info({"msg": "Checkpoints were prepared for an outdated frame, stop proccessing"})
+                break
```
vgorkavenko, line 232: it will lead to `All epochs were processed` log on the 242 line. can we `return False`?

madlabman: I did add a raise.

vgorkavenko (approved):

PR #457, opened 2024-05-22
feat: Support dynamic versions in base Oracle

F4ever:


src/modules/submodules/consensus.py
```diff
@@ -209,29 +205,29 @@ def get_blockstamp_for_report(self, last_finalized_blockstamp: BlockStamp) -> Op
...
-                'msg': f'Oracle\'s version higher than contract supports. '
-                       f'Expected contract version {contract_version} got {self.CONTRACT_VERSION}. '
-                       f'Expected consensus version {consensus_version} got {self.CONSENSUS_VERSION}.'
-            })
-
-        return compatibility
+            raise IncompatibleOracleVersion(
+                f'Incompatible Oracle version. '
+                f'Expected contract version {contract_version} got {self.COMPATIBLE_CONTRACT_VERSIONS}. '
+                f'Expected consensus version {consensus_version} got {self.COMPATIBLE_CONSENSUS_VERSIONS}.'
```
vgorkavenko, line 229:
```suggestion
                f'Expected oracle version {contract_version} got {self.COMPATIBLE_CONTRACT_VERSIONS}. '
                f'Expected oracle version {consensus_version} got {self.COMPATIBLE_CONSENSUS_VERSIONS}.'
```

src/modules/ejector/ejector.py
```diff
@@ -98,7 +98,7 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
         data, data_format = encode_data(validators)
 
         report_data = ReportData(
-            self.CONSENSUS_VERSION,
+            self.report_contract.get_consensus_version(blockstamp.block_hash),
```
vgorkavenko, line 101: is it integer?

src/modules/submodules/consensus.py
```diff
@@ -209,29 +205,29 @@ def get_blockstamp_for_report(self, last_finalized_blockstamp: BlockStamp) -> Op
...
-            logger.warning({
-                'msg': f'Oracle\'s version higher than contract supports. '
-                       f'Expected contract version {contract_version} got {self.CONTRACT_VERSION}. '
-                       f'Expected consensus version {consensus_version} got {self.CONSENSUS_VERSION}.'
-            })
-
-        return compatibility
+            raise IncompatibleOracleVersion(
+                f'Incompatible Oracle version. '
+                f'Expected contract version {contract_version} got {self.COMPATIBLE_CONTRACT_VERSIONS}. '
```
madlabman, line 228: I would swap the variables, in the way that we expect `COMPATIBLE_CONTRACT_VERSIONS` but got `contract_version` (because it's what we get from the network, right?).

src/modules/ejector/ejector.py
F4ever: Yes
`def get_consensus_version(self, block_identifier: BlockIdentifier = 'latest') -> int:`

src/modules/submodules/consensus.py
F4ever: fixed

src/modules/submodules/consensus.py
F4ever: fixed

vgorkavenko (approved):

madlabman (approved):

PR #462, opened 2024-06-05
Feat: AO multi third phase support

F4ever:


src/modules/accounting/third_phase/extra_data_v2.py
```diff
@@ -0,0 +1,146 @@
...
+
+                index += 1
+
+            result.append(tx)
+
+        return index, result
+
+    @staticmethod
+    def add_hashes_to_transactions(txs_data: list[bytes]) -> tuple[bytes, list[bytes]]:
+        txs_data.reverse()
```
vgorkavenko, line 134: why we need reverse here?

src/modules/accounting/types.py
```diff
@@ -70,6 +72,13 @@ class OracleReportLimits:
     max_node_operators_per_extra_data_item_count: int
     request_timestamp_margin: int
     max_positive_token_rebase: int
+    appeared_validators_per_day_limit: Optional[int] = None
+
+    @classmethod
+    def from_response(cls, **kwargs) -> Self:
+        # churn_validators_per_day_limit was renamed in new version
```
vgorkavenko, line 79: need a link to the evidence where `churn_validators_per_day_limit` is changed

src/modules/accounting/third_phase/extra_data_v2.py
madlabman: We construct a linked list of items bound via `nextHash` field. So we start hashing from the very end to get the sequence.

src/modules/accounting/types.py
```diff
@@ -61,7 +63,7 @@ class AccountingProcessingState:
 
 @dataclass
 class OracleReportLimits:
-    churn_validators_per_day_limit: int
+    exited_validators_per_day_limit: int
```
madlabman, line 66: Is this change backward compatible?

F4ever: Yes, because
```
    @classmethod
    def from_response(cls, **kwargs) -> Self:
        # Compatability breaking rename
        # churn_validators_per_day_limit -> exited_validators_per_day_limit
        # Unpack structure by order
        return cls(*kwargs.values())  # pylint: disable=no-value-for-parameter
```

madlabman (approved):

PR #464, opened 2024-06-12
Fix: backward report build

F4ever:


src/providers/execution/contracts/staking_router.py
```diff
@@ -58,3 +58,30 @@ def get_all_node_operator_digests(self, module: StakingModule, block_identifier:
 
 class StakingRouterContractV2(StakingRouterContract):
     abi_path = './assets/StakingRouterV2.json'
+
+    @lru_cache(maxsize=1)
```
vgorkavenko, line 62: not sure that it is good idea to have cache and `block_identifier = 'latest'` by default at the same time

F4ever: This is a special cache, that ignores 'latest' and default values and caches only hashes and numbers

src/providers/execution/contracts/staking_router.py
```diff
@@ -58,3 +58,30 @@ def get_all_node_operator_digests(self, module: StakingModule, block_identifier:
...
+    def get_staking_modules(self, block_identifier: BlockIdentifier = 'latest') -> list[StakingModule]:
+        """
+        Returns all registered staking modules
+        """
+        contract_version = self.get_contract_version(block_identifier)
+
+        if contract_version == 1:
+            # Backward compatibility in case if new oracle have to build report for old protocol version
+            # But latest contracts has new version
+            staking_router = self.w3.eth.contract(
```
vgorkavenko, line 73: let's log that we call an old abi. here or instead of logging at the end of function

F4ever: fixed

vgorkavenko (approved):

PR #465, opened 2024-06-12
fix: `CSFeeOracle` with no initial epoch

vgorkavenko:


src/modules/csm/csm.py
```diff
@@ -253,12 +256,15 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
         return self.state.is_fulfilled
 
     @lru_cache(maxsize=1)
-    def current_frame_range(self, blockstamp: BlockStamp) -> tuple[SlotNumber, SlotNumber]:
+    def current_frame_range(self, converter: Web3Converter, blockstamp: BlockStamp) -> tuple[SlotNumber, SlotNumber]:
```
madlabman, line 259: Why not to `converter = self.converter(blockstamp)`?

vgorkavenko: to not to do it twice

PR #469, opened 2024-06-13
Fix report

madlabman:


tests/modules/csm/test_csm_module.py
```diff
@@ -0,0 +1,65 @@
...
+            ValidatorIndex(1): AttestationsAggregate(included=1000, assigned=1000),
+            ValidatorIndex(2): AttestationsAggregate(included=1000, assigned=1000),
+            ValidatorIndex(3): AttestationsAggregate(included=999, assigned=1000),
+            ValidatorIndex(4): AttestationsAggregate(included=900, assigned=1000),
+            ValidatorIndex(5): AttestationsAggregate(included=500, assigned=1000),  # underperforming
+            ValidatorIndex(5): AttestationsAggregate(included=500, assigned=1000),  # underperforming
+            ValidatorIndex(6): AttestationsAggregate(included=0, assigned=0),  # underperforming
+            ValidatorIndex(7): AttestationsAggregate(included=900, assigned=1000),
+            ValidatorIndex(8): AttestationsAggregate(included=500, assigned=1000),  # underperforming
+            # ValidatorIndex(9): AttestationsAggregate(included=0, assigned=0),  # missing in state
```
vgorkavenko, line 55: uncomment

tests/modules/csm/test_state.py
```diff
@@ -0,0 +1,127 @@
...
+    assert state.avg_perf == 0
+
+    state = State(
+        {
+            ValidatorIndex(0): AttestationsAggregate(included=0, assigned=0),
+            ValidatorIndex(1): AttestationsAggregate(included=0, assigned=0),
+        }
+    )
+
+    # assert state.avg_perf == 0
```
vgorkavenko, line 38: uncomment

src/modules/csm/csm.py
```diff
@@ -255,8 +211,69 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
...
+
+        assert self.state
+
+        threshold = self.state.avg_perf - self.w3.csm.oracle.perf_leeway(blockstamp.block_hash)
+        operators_to_validators = self.module_validators_by_node_operators(blockstamp)
+
+        # Build the map of the current distribution operators.
+        distribution: dict[NodeOperatorId, int] = {}
+        for (_, no_id), validators in operators_to_validators.items():
+            if no_id in self.stuck_operators(blockstamp):
```
vgorkavenko, line 225: `self.stuck_operators(blockstamp)` should be got outside `for`

src/modules/csm/csm.py
```diff
@@ -255,8 +211,69 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
...
+            no_share = 0
+
+            for v in validators:
+                try:
+                    aggr = self.state.data[ValidatorIndex(int(v.index))]
+                except KeyError:
+                    # It's possible that the validator is not assigned to any duty, hence it's performance
+                    # is not presented in the aggregates (e.g. exited, pending for activation etc).
+                    continue
+                else:
```
vgorkavenko, line 237: `else` doesn't make sense here, as I see

src/modules/csm/csm.py
```diff
@@ -255,8 +211,69 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
...
+                    if aggr.perf > threshold:
+                        # Count of assigned attestations used as a metrics of time
+                        # the validator was active in the current frame.
+                        no_share += aggr.assigned
+
+            if no_share:
+                distribution[no_id] = no_share
```
vgorkavenko, line 244:
```suggestion
            distribution[no_id] = no_share
```
wdyt? the code will become easier to read

src/modules/csm/csm.py
```diff
@@ -255,8 +211,69 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
...
+
+        # Calculate share of each CSM node operator.
+        to_distribute = self.w3.csm.fee_distributor.shares_to_distribute(blockstamp.block_hash)
```
vgorkavenko, line 247: makes sense to get only if `total > 0`

src/modules/csm/csm.py
```diff
@@ -255,8 +211,69 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
...
+        shares: dict[NodeOperatorId, int] = defaultdict(int)
+        total = sum(p for p in distribution.values())
+        if total > 0:
+            for no_id, portion in distribution.items():
```
vgorkavenko, line 251: `no_share` becomes `portion` here. let's make the same name for readability

src/modules/csm/csm.py
```diff
@@ -255,8 +211,69 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
...
+        l_ref_slot, _ = self.current_frame_range(blockstamp)
+        # NOTE: r_block is guaranteed to be <= ref_slot, and the check
+        # in the inner frames assures the  l_block <= r_block.
+        return self.w3.csm.get_csm_stuck_node_operators(
+            get_first_non_missed_slot(
+                self.w3.cc,
+                l_ref_slot,
+                blockstamp.slot_number,
+                direction='forward',
+            ).message.body.execution_payload.block_hash,
```
vgorkavenko, line 269: put in var for readability

tests/modules/csm/test_csm_module.py
madlabman: It's done intentionally, this validator doesn't exist from the State's perspective.

src/modules/csm/csm.py
madlabman: We don't need empty entries in a tree.

src/modules/csm/csm.py
madlabman: It's done to clearly separate the case we catch and the rest.

src/modules/csm/csm.py
madlabman: It's cached, so no need for an extra variable.

src/modules/csm/csm.py
vgorkavenko: but you work with `shares` for tree

src/modules/csm/csm.py
vgorkavenko: there is `continue`, so it won't reach the code below

src/modules/csm/csm.py
vgorkavenko: as a reader it is not clear that there is caching

src/modules/csm/csm.py
madlabman: So basically it's either here, either in the shares' calculation loop.

src/modules/csm/csm.py
madlabman: Yeah, but it's cheap and saves for readability.

src/modules/csm/csm.py
madlabman: It's already readable, nah?

vgorkavenko: all in a pile

src/modules/csm/csm.py
madlabman: Bad reader.

src/modules/csm/csm.py
madlabman: Moved because of early return now.

vgorkavenko (approved):

PR #472, opened 2024-06-17
fix: `minimum checkpoint step is not reached` is not error

vgorkavenko:


madlabman:
It will blow up the whole oracle now?

PR #474, opened 2024-06-18
feat: block roots from state, events

vgorkavenko:


src/modules/checks/suites/execution_node.py
```diff
@@ -48,13 +48,25 @@ def check_events_range_availability(web3, blockstamp, deposit_contract):
     list(get_events_in_range(deposit_contract.events.DepositEvent, blockstamp.block_number, latest_block.number))
 
 
-def check_events_week_range_availability(web3, deposit_contract):
-    """Check that execution-client able to get event logs a week ago"""
+@pytest.mark.parametrize('range_', [8 * 225 * 32, 32 * 225 * 32], ids=['week', 'month'])
+def check_events_week_range_availability(web3, deposit_contract, range_):
```
madlabman, line 52: Does it make sense to define a `DepositContract` in the provider module?

src/modules/checks/suites/execution_node.py
```diff
@@ -48,13 +48,25 @@ def check_events_range_availability(web3, blockstamp, deposit_contract):
...
-            l_block=BlockNumber(max(0, latest_block.number - 8 * 225 * 32)),  # 8 days
-            r_block=latest_block.number,
+            l_block=BlockNumber(from_block),
+            r_block=to_block,
         )
     )
+    deposits_count_before = _get_deposit_count(from_block)
+    deposits_count_now = _get_deposit_count(to_block)
+    if deposits_count_now - deposits_count_before > 0:
+        assert events, "No events found"
```
madlabman, line 72: len should be equal to the diff?

src/modules/checks/suites/conftest.py
```diff
@@ -40,6 +41,8 @@ def web3():
     })
     if variables.LIDO_LOCATOR_ADDRESS:
         web3.attach_modules({'lido_contracts': LidoContracts})
+    if variables.CSM_ORACLE_ADDRESS and variables.CSM_MODULE_ADDRESS:
+        web3.attach_modules({'csm': LazyCSM})
```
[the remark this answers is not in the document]
madlabman: How we treat this behaviour in the CSM submodule?

src/modules/checks/suites/execution_node.py
vgorkavenko: make sense

src/modules/checks/suites/consensus_node.py
```diff
@@ -4,20 +4,24 @@
...
 
 def check_block_header_provided(web3: Web3, blockstamp):
     """Check that consensus-client able to provide block header"""
-    web3.cc.get_block_header(blockstamp.slot_number)
+    assert web3.cc.get_block_header(blockstamp.slot_number), "consensus-client provide no block header"
+
+
+def check_block_roots_from_state_provided(web3: Web3, blockstamp):
+    """Check that consensus-client able to provide block roots from state"""
+    assert web3.cc.get_state_block_roots(blockstamp.slot_number), "consensus-client provide no block roots from state"
```
madlabman, line 27: What blockstamp is being used here?

vgorkavenko: either `finalized` or `finalized - frame size`

src/modules/checks/suites/conftest.py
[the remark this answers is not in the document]
vgorkavenko: fixed

src/modules/checks/suites/consensus_node.py
madlabman: Does it make sense? I think we should to step back for CL tests to make sure we have historical state?

madlabman: frame_size from what?

src/modules/checks/suites/execution_node.py
vgorkavenko: it is used only for this file and only in testing purpose. not sure about changing the provider module for this

src/modules/checks/suites/consensus_node.py
vgorkavenko: > Does it make sense? I think we should to step back for CL tests to make sure we have historical state?

I changed nothing here. The test will be run with finalized slot and with slot with -270 epochs back. see `blockstamp_frame_ago` fixture

madlabman (approved):

PR #477, opened 2024-06-19
fix: iterator conditions

vgorkavenko:


src/modules/csm/checkpoint.py
```diff
@@ -50,48 +55,46 @@ def __init__(
...
         self.l_epoch = l_epoch
-        self.r_epoch = self._get_adjusted_r_epoch(r_epoch, finalized_epoch)
+        self.r_epoch = r_epoch
 
-    def __iter__(self):
+        self.max_available_epoch_to_check = min(
+            self.r_epoch, EpochNumber(finalized_epoch - self.CHECKPOINT_SLOT_DELAY_EPOCHS)
+        )
 
+        if self.r_epoch > self.max_available_epoch_to_check and not self._is_min_step_reached(finalized_epoch):
```
madlabman, line 64:
```suggestion
        if self.r_epoch > self.max_available_epoch_to_check and not self._is_min_step_reached(self.max_available_epoch_to_check):
```

src/modules/csm/checkpoint.py
```diff
@@ -50,48 +55,46 @@ def __init__(
...
+                checkpoint_slot = self.converter.get_epoch_first_slot(
+                    EpochNumber(epoch + self.CHECKPOINT_SLOT_DELAY_EPOCHS)
+                )
                 logger.info(
                     {"msg": f"Checkpoint slot {checkpoint_slot} with {len(checkpoint_epochs)} duty epochs is prepared"}
                 )
                 yield Checkpoint(checkpoint_slot, checkpoint_epochs)
                 checkpoint_epochs = []
+            if epoch == self.max_available_epoch_to_check:
+                break
```
madlabman, line 83: Since we have python 3.12 we can significantly simplify this function.

```python
    def __iter__(self):
        for checkpoint_epochs in batched(
            sequence(self.l_epoch, self.max_available_epoch_to_check),
            self.MAX_CHECKPOINT_STEP,
        ):
            checkpoint_slot = self.converter.get_epoch_first_slot(
                EpochNumber(max(checkpoint_epochs) + self.CHECKPOINT_SLOT_DELAY_EPOCHS)
            )
            logger.info(
                {"msg": f"Checkpoint slot {checkpoint_slot} with {len(checkpoint_epochs)} duty epochs is prepared"}
            )
            yield Checkpoint(checkpoint_slot, checkpoint_epochs)
```

madlabman (commented):
Let's try the following code for `_get_block_roots`:

```python
    def _get_block_roots(self, checkpoint_slot: SlotNumber):
        logger.info({"msg": f"Get block roots for slot {checkpoint_slot}"})
        # checkpoint for us like a time point, that's why we use slot, not root
        pivot_index = checkpoint_slot % SLOTS_PER_HISTORICAL_ROOT
        br = self.cc.get_state_block_roots(checkpoint_slot)
        # replace duplicated roots to None to mark missed slots
        return [br[i] if i == pivot_index or br[i] != br[i - 1] else None for i in range(len(br))]
```

madlabman (approved):

F4ever (approved):

PR #491, opened 2024-07-03
feat: custom errors + kapi check

F4ever:


src/main.py
```diff
@@ -137,10 +144,12 @@ def check_providers_chain_ids(web3: Web3, cc: ConsensusClientModule, kac: KeysAP
     if execution_chain_id == consensus_chain_id == keys_api_chain_id:
         return
 
-    raise ValueError('Different chain ids detected:\n'
-                     f'Execution chain id: {execution_chain_id}\n'
-                     f'Consensus chain id: {consensus_chain_id}\n'
-                     f'Keys API chain id: {keys_api_chain_id}\n')
+    raise IncompatibleVersionException(
```
madlabman, line 147: Version?

F4ever: fixed

src/providers/http_provider.py
```diff
@@ -139,7 +141,7 @@ def _get_without_fallbacks(
                     code=0,
                     domain=urlparse(host).netloc,
                 )
-                raise error
+                raise self.PROVIDER_EXCEPTION(status=0, text='Response error.') from error
```
vgorkavenko, line 144: does the final message in log include text from `error`?

F4ever: Will be something like this
```
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in func
ConnectionError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
RuntimeError: Failed to open database
```

vgorkavenko: Yes, but it will be caught on the higher level in try\catch and wrapped by logger
image

F4ever: Yeah, thats an issue

vgorkavenko (approved):

vgorkavenko (approved):

PR #499, opened 2024-08-06
fix: get_first_non_missed_slot

vgorkavenko:


src/utils/slot.py
```diff
@@ -23,14 +22,13 @@ class SlotNotFinalized(Exception):
     pass
 
 
-def get_first_non_missed_slot(
+def _get_closest_non_missed_headers(
```
F4ever, line 25: Not the closest?

src/utils/slot.py
```diff
@@ -90,25 +83,49 @@ def get_first_non_missed_slot(
...
-            existing_header = parent_header
+def get_prev_non_missed_slot(
+    cc: ConsensusClient,
+    slot: SlotNumber,
+    last_finalized_slot_number: SlotNumber,
+) -> BlockDetailsResponse:
+    """
+    Get non-missed slot data. In case of missed slot, we take parent root and get parent slot data.
+    """
+    slot_is_missing, existing_header = _get_closest_non_missed_headers(cc, slot, last_finalized_slot_number)
```
F4ever, line 98:
```suggestion
    is_slot_missing, existing_header = _get_closest_non_missed_headers(cc, slot, last_finalized_slot_number)
```

src/utils/slot.py
F4ever, line 25: not the closest

src/utils/slot.py
```diff
@@ -23,14 +22,13 @@ class SlotNotFinalized(Exception):
...
+def _get_closest_non_missed_headers(
     cc: ConsensusClient,
     slot: SlotNumber,
     last_finalized_slot_number: SlotNumber,
-    direction: Literal['back'] | Literal['forward'] = 'back'
-) -> BlockDetailsResponse:
+) -> tuple[bool, BlockHeaderFullResponse]:
     """
-    Get past/next closest non-missed slot and returns its details.
+    Get closest non-missed slot and returns its header.
```
F4ever, line 31: Also, not the closest

src/utils/slot.py
```diff
@@ -90,25 +83,49 @@ def get_first_non_missed_slot(
...
-    return slot_details
+    if not slot_is_missing:
+        return cc.get_block_details(existing_header.data.root)
+
+    non_missed_header_parent_root = existing_header.data.header.message.parent_root
+    parent_header = cc.get_block_header(non_missed_header_parent_root)
+
+    if (
+        int(parent_header.data.header.message.slot) >= slot
+        or int(existing_header.data.header.message.slot) - int(parent_header.data.header.message.slot) < 1
```
F4ever, line 108: in another words:
existing_header.data.header.message.slot == parent_header.data.header.message.slot

F4ever: possibly we can check that
block_number between existing and parent is 1

vgorkavenko: for some reason existing_header slot can be less than parent_header slot from node perspective, so with the current solution we will catch this

madlabman (approved):

F4ever (approved):

PR #500, opened 2024-08-08
tests: current_frame_range tests for CSM module

madlabman:


src/modules/csm/csm.py
```diff
@@ -273,11 +273,14 @@ def current_frame_range(self, blockstamp: BlockStamp) -> tuple[EpochNumber, Epoc
         if converter.frame_config.initial_epoch == far_future_initial_epoch:
             raise ValueError("CSM oracle initial epoch is not set yet")
 
-        l_ref_slot = self.w3.csm.get_csm_last_processing_ref_slot(blockstamp)
+        l_ref_slot = last_processing_ref_slot = self.w3.csm.get_csm_last_processing_ref_slot(blockstamp)
         r_ref_slot = initial_ref_slot = self.get_initial_ref_slot(blockstamp)
 
+        if last_processing_ref_slot > blockstamp.slot_number:
+            raise ValueError(f"{last_processing_ref_slot=} > {blockstamp.slot_number=}")
```
vgorkavenko, line 280: let's add a bit details here. like " X more than Y. probably, expected slot isn't reached yet"

src/modules/csm/csm.py
```diff
@@ -293,6 +296,11 @@ def current_frame_range(self, blockstamp: BlockStamp) -> tuple[EpochNumber, Epoc
                 EpochNumber(converter.get_epoch_by_slot(l_ref_slot) + converter.frame_config.epochs_per_frame)
             )
 
+        if last_processing_ref_slot > r_ref_slot:
+            raise CSMError(f"{last_processing_ref_slot=} > {r_ref_slot=}")
```
vgorkavenko, line 300: and here as well. it isn't legit as I understand

tests/modules/csm/test_csm_module.py
```diff
@@ -63,3 +74,147 @@ def test_calculate_distribution(module: CSOracle, csm: CSM):
...
+        pytest.param(
+            FrameTestParam(
+                epochs_per_frame=32,
+                initial_ref_slot=last_slot_of_epoch(100),
+                last_processing_ref_slot=0,
+                current_ref_slot=last_slot_of_epoch(164),
+                finalized_slot=last_slot_of_epoch(170),
+                expected_frame=(69, 164),
+            ),
+            id="frame_0",
```
vgorkavenko, line 163: it isn't frame 0, actually.
it's frame after 2 missing frames because of
```
initial_ref_slot=last_slot_of_epoch(100),
last_processing_ref_slot=0,
current_ref_slot=last_slot_of_epoch(164),
```

tests/modules/csm/test_csm_module.py
```diff
@@ -63,3 +74,147 @@ def test_calculate_distribution(module: CSOracle, csm: CSM):
...
+        ),
+        pytest.param(
+            FrameTestParam(
+                epochs_per_frame=0,
```
vgorkavenko, line 167: is it legal?

madlabman: It works for the test, so I used 0 here intentionally, so it will brake the test, if something will be changed, and will take our attention.

tests/modules/csm/test_csm_module.py
madlabman: Yeah, I have thought about a first report instead.

src/modules/csm/csm.py
madlabman: Changed to execution layer provider error instead. `last_processing_ref_slot` can't be greater than current slot, and it's pure requests to the Oracle contract, so it's rather something wrong with EL.

src/modules/csm/csm.py
madlabman: Done.

vgorkavenko (approved):

PR #501, opened 2024-08-08
fix: move metrics setting to `_check_duty`

vgorkavenko:


src/modules/csm/checkpoint.py
```diff
@@ -202,6 +204,9 @@ def _check_duty(
             self.state.add_processed_epoch(duty_epoch)
             self.state.commit()
             self.state.status()
+            unprocessed_epochs = self.state.unprocessed_epochs
+            CSM_UNPROCESSED_EPOCHS_COUNT.set(len(unprocessed_epochs))
+            CSM_MIN_UNPROCESSED_EPOCH.set(min(unprocessed_epochs or {EpochNumber(0)}))
```
madlabman, line 209: I'd wrap -1 instead.

src/modules/csm/checkpoint.py
```diff
@@ -202,6 +204,9 @@ def _check_duty(
             self.state.add_processed_epoch(duty_epoch)
             self.state.commit()
             self.state.status()
+            unprocessed_epochs = self.state.unprocessed_epochs
```
madlabman, line 207: Why so?

vgorkavenko: to do calcs once and be sure that the value will be the same for both of metrics

src/modules/csm/checkpoint.py
```diff
@@ -202,6 +204,9 @@ def _check_duty(
...
+            CSM_UNPROCESSED_EPOCHS_COUNT.set(len(unprocessed_epochs))
+            CSM_MIN_UNPROCESSED_EPOCH.set(min(unprocessed_epochs or {EpochNumber(-1)}))
```
F4ever, line 209: Maybe better
```suggestion
            if unprocessed_epochs:
                CSM_MIN_UNPROCESSED_EPOCH.set(min(unprocessed_epochs))
            else:
                CSM_MIN_UNPROCESSED_EPOCH.set(max(processed_epochs) + 1))
```

vgorkavenko: it will mislead

F4ever (approved):

src/modules/csm/checkpoint.py
F4ever: Let's do it outside lock?

F4ever: self.state.status() seems can be done outside

F4ever (approved):

PR #513, opened 2024-08-21
feat: exclude operators with slashed validators in the frame

vgorkavenko:


src/modules/csm/csm.py
```diff
@@ -219,6 +220,14 @@ def calculate_distribution(self, blockstamp: ReferenceBlockStamp) -> tuple[int,
                     # is not presented in the aggregates (e.g. exited, pending for activation etc).
                     continue
 
+                if v.validator.slashed:
+                    # It means that validator was active during the frame
+                    # and got slashed and didn't meet the exit epoch,
+                    # so we should not count such operator in the distribution
+                    if distribution.get(no_id):
```
madlabman, line 227: The check seems to be redundant.

src/modules/csm/csm.py
```diff
@@ -219,6 +220,14 @@ def calculate_distribution(self, blockstamp: ReferenceBlockStamp) -> tuple[int,
                     # is not presented in the aggregates (e.g. exited, pending for activation etc).
                     continue
 
+                if v.validator.slashed:
```
madlabman, line 223: Move it before try..catch block.

src/modules/csm/csm.py
vgorkavenko: removed

src/modules/csm/csm.py
vgorkavenko: we can't do that. otherwise we will remove good operators that had slashed validator in the previous frame, for example

madlabman (approved):

F4ever (approved):

PR #516, opened 2024-09-13
feat: tests for csm build report

vgorkavenko:


src/modules/csm/csm.py
```diff
@@ -110,18 +108,24 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
         else:
             logger.info({"msg": "No previous CID available"})
 
-        tree = self.make_tree(shares)
-        tree_cid: CID | None = None
+        if distributed > 0:
+            curr_tree = self.make_tree(shares)
+            if not curr_tree:
+                raise InconsistentData("No tree to publish but shares are distributed")
```
madlabman, line 114: I'm not sure if it makes sense. `make_tree` doesn't depend on `distributed`. I think it's better to check on line 98 invariant `distributed and shares`.

src/modules/csm/types.py
```diff
@@ -19,7 +19,7 @@ class ReportData:
     consensusVersion: int
     ref_slot: SlotNumber
     tree_root: HexBytes
-    tree_cid: CID | Literal[""]
+    tree_cid: CID
```
madlabman, line 22: `prev_cid` still can be `Literal[""]` and hence zeroed first report will result in `tree_cid = ""`.

src/web3py/extensions/csm.py
```diff
@@ -46,10 +46,8 @@ def get_csm_last_processing_ref_slot(self, blockstamp: BlockStamp) -> SlotNumber
     def get_csm_tree_root(self, blockstamp: BlockStamp) -> HexBytes:
         return self.fee_distributor.tree_root(blockstamp.block_hash)
 
-    def get_csm_tree_cid(self, blockstamp: BlockStamp) -> CID | None:
+    def get_csm_tree_cid(self, blockstamp: BlockStamp) -> CID:
         result = self.fee_distributor.tree_cid(blockstamp.block_hash)
-        if not result:
-            return None
```
madlabman, line 52: Why so? Empty string is not a valid CID and will result in CIDv1("") which is not correct.

tests/conftest.py
```diff
@@ -113,6 +114,9 @@ def keys_api_client(request, responses_path, web3):
 @pytest.fixture()
 def csm(web3):
     mock = Mock()
+    mock.module.MAX_OPERATORS_COUNT = UINT64_MAX
+    web3.ipfs = Mock()
+    web3.lido_contracts = Mock()
```
madlabman, line 119: Why not to keep it at the test level?

tests/modules/csm/test_csm_module.py
```diff
@@ -493,3 +498,249 @@ def test_collect_data_fulfilled_state(
...
+
+@dataclass(frozen=True)
+class BuildReportTestParam:
+    prev_root: str
+    prev_cid: str
+    prev_acc_shares: Iterable[tuple[NodeOperatorId, int]]
+    curr_distribution: Mock
+    curr_root: str
+    curr_cid: str
+    curr_log: str
```
madlabman, line 511: We have two CIDs, tree and log. It's not very clear from this params.

tests/modules/csm/test_csm_module.py
```diff
@@ -493,3 +498,249 @@ def test_collect_data_fulfilled_state(
...
+                        # distributed
+                        6,
+                        # shares
+                        defaultdict(int, {NodeOperatorId(0): 1, NodeOperatorId(1): 2, NodeOperatorId(2): 3}),
+                        # log
+                        Mock(),
+                    )
+                ),
+                curr_root="0x100e",
+                curr_cid="0x100c",
```
madlabman, line 558: It looks like bytes instead of CID.

src/modules/csm/csm.py
madlabman: It's guaranteed by `calculate_distribution` by the way.

tests/modules/csm/test_csm_module.py
vgorkavenko: fixed

tests/modules/csm/test_csm_module.py
vgorkavenko: fixed

tests/conftest.py
vgorkavenko: fixed

src/web3py/extensions/csm.py
vgorkavenko: changed

src/modules/csm/csm.py
vgorkavenko: changed

src/modules/csm/csm.py
```diff
@@ -98,7 +99,7 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
         prev_root = self.w3.csm.get_csm_tree_root(blockstamp)
         prev_cid = self.w3.csm.get_csm_tree_cid(blockstamp)
 
-        if bool(prev_root) != bool(prev_cid):
+        if bool(prev_root) != (prev_root is not ZERO_HASH):
```
madlabman, line 102:
```suggestion
        if bool(prev_cid) != (prev_root is not ZERO_HASH):
```

src/modules/csm/csm.py
```diff
@@ -110,13 +111,13 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
...
                 blockstamp.ref_slot,
-                tree_root=prev_root or HexBytes(32),
+                tree_root=prev_root,
                 tree_cid=prev_cid or "",
                 log_cid=log_cid,
                 distributed=0,
             ).as_tuple()
 
-        if prev_cid and prev_root:
+        if prev_cid and prev_root is not ZERO_HASH:
```
madlabman, line 120: Didn't quite get why the check for root at all.

src/modules/csm/csm.py
madlabman: If I'm right, there's a missing test case.

src/modules/csm/csm.py
vgorkavenko: practically we expect them both to be presented or not presented so i've tried to show it here

madlabman (approved):

PR #519, opened 2024-09-19
fix: negative rebase check with lido handleOracleReport simulation

F4ever:


src/providers/execution/contracts/lido.py
```diff
@@ -36,6 +37,67 @@ def handle_oracle_report(
...
+                cl_balance,
+                withdrawal_vault_balance,
+                el_rewards,
+                shares_to_burn,
+                accounting_oracle_address,
+                HexStr('0x' + ref_slot.to_bytes(32).hex()),
+                block_identifier,
+            )
+        except ValueError as error:
+            # {'code': -32602, 'message': 'invalid argument 2: hex number with leading zero digits'}
```
vgorkavenko, line 54: could we check exact message here? in current implementation we will always try to use Erigon format, but the error might be not about that

vgorkavenko (changes requested):

src/providers/execution/contracts/lido.py
F4ever: I'm afraid to rely on the node response, I would just guess if it doesn't work try another format

vgorkavenko (approved):

PR #521, opened 2024-09-23
Log env variables

hweawer:


src/variables.py
```diff
@@ -87,22 +87,66 @@
...
 
 def raise_from_errors(errors):
     if errors:
         raise ValueError("The following variables are required: " + ", ".join(errors))
+
+
+# All non-private env variables to the logs in main
+PUBLIC_ENV_VARS = {
+    'KEYS_API_URI': KEYS_API_URI,
+    'GW3_ACCESS_KEY': GW3_ACCESS_KEY,
```
F4ever, line 112: Remove

hweawer: Moved to private

F4ever (approved):

PR #542, opened 2024-10-16
Hotfix: FIx wrong active validators calculating 

F4ever:
Bug:
`_get_total_active_validators` calculates all validators instead of filtering only active validators


avsetsin (approved):
LGTM

PR #544, opened 2024-10-28
fix: expect `InitialEpochIsYetToArrive` error

vgorkavenko:


madlabman (approved):

src/modules/csm/csm.py
```diff
@@ -139,6 +144,15 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
             distributed=distributed,
         ).as_tuple()
 
+    def get_current_frame(self, blockstamp: BlockStamp) -> CurrentFrame:
+        try:
+            current_frame = super().get_current_frame(blockstamp)
```
F4ever, line 149:
```suggestion
            return super().get_current_frame(blockstamp)
```

PR #556, opened 2024-10-31
feat: EIP-7549 compatibility

madlabman:
- [EIP-7549](https://github.com/ethereum/EIPs/blob/a7fb2260ae2ea39bdd31886832c9e45452d0e76a/EIPS/eip-7549.md)
- [get_attesting_indices](https://github.com/ethereum/consensus-specs/blob/2c1f677187e6534aec77057a7d1cc746a40d3630/specs/electra/beacon-chain.md#modified-get_attesting_indices)
- [get_beacon_committee](https://github.com/ethereum/consensus-specs/blob/c060147812933b38327e2be619597b3a372d36ae/specs/phase0/beacon-chain.md#get_beacon_committee)

The "/eth/v1/beacon/blocks/{block_id}/attestations" is deprecated, but we can still use it for now; it appears to be easier to keep using it until Electra to make sure all nodes support the v2 version.

P.S. A minor bug related to not considering the length of a bitlist appears to have no impact on the oracle.


src/modules/csm/checkpoint.py
```diff
@@ -229,18 +234,47 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
+        for committee_idx in get_committee_indices(attestation):
+            committee = committees.get((attestation.data.slot, committee_idx), [])
+            att_bits = hex_bitlist_to_list(attestation.aggregation_bits)[committee_offset:][: len(committee)]
+            for index_in_committee in get_set_indices(att_bits):
+                committee[index_in_committee].included = True
+            committee_offset += len(committee)
+
+
+def get_committee_indices(attestation: BlockAttestation) -> list[CommitteeIndex]:
+    if not is_electra_attestation(attestation):
```
vgorkavenko, line 247: could we reframe it to `if is_before_electra_attestation` ?

src/modules/csm/checkpoint.py
```diff
@@ -229,18 +234,47 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
+        return [attestation.data.index]
+    return [str(i) for i in get_set_indices(hex_bitvector_to_list(attestation.committee_bits))]
+
+
+def is_electra_attestation(attestation: BlockAttestation) -> TypeGuard[BlockAttestationElectra]:
+    has_committee_bits = getattr(attestation, "committee_bits") is not None
+    has_zero_index = attestation.data.index == "0"
```
vgorkavenko, line 254: Does it work the same for v1 and v2 endpoints? Or `attestation.data.index` is removed for v2
Just to be safe in case of a version change in our client

src/modules/csm/checkpoint.py
madlabman: I think it has the same complexity.

src/modules/csm/checkpoint.py
madlabman: When I tested I had the same results, and we use v1 only for now.  `attestation.data.index` is not removed and kept equal to "0" as of spec.

src/modules/csm/checkpoint.py
vgorkavenko: semantically "not electra attestation" sounds like "any attestation from fork before and after electra" for me

vgorkavenko (approved):

src/providers/consensus/client.py
```diff
@@ -108,7 +110,9 @@ def get_block_details(self, state_id: SlotNumber | BlockRoot) -> BlockDetailsRes
         return BlockDetailsResponse.from_response(**data)
 
     @lru_cache(maxsize=256)
-    def get_block_attestations(self, state_id: SlotNumber | BlockRoot) -> list[BlockAttestation]:
+    def get_block_attestations(
+        self, state_id: SlotNumber | BlockRoot
```
F4ever, line 114:
```suggestion
        self, 
        state_id: SlotNumber | BlockRoot,
```

src/modules/csm/checkpoint.py
```diff
@@ -229,18 +234,48 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
+            committee_offset += len(committee)
+
+
+def get_committee_indices(attestation: BlockAttestation) -> list[CommitteeIndex]:
+    if is_eip7549_attestation(attestation):
+        return [str(i) for i in get_set_indices(hex_bitvector_to_list(attestation.committee_bits))]
+    return [attestation.data.index]
+
+
+def is_eip7549_attestation(attestation: BlockAttestation) -> TypeGuard[BlockAttestationElectra]:
```
F4ever, line 252:
```suggestion
def is_eip7549_attestation(attestation: BlockAttestation) -> bool:
```

F4ever: Maybe add this method to `BlockAttestation` class?

src/modules/csm/checkpoint.py
```diff
@@ -3,25 +3,28 @@
...
 logger = logging.getLogger(__name__)
 lock = Lock()
 
 
-class MinStepIsNotReached(Exception):
-    ...
+class MinStepIsNotReached(Exception): ...
+
+
+type BlockAttestation = BlockAttestationPhase0 | BlockAttestationElectra
```
F4ever, line 27: We have already declaration of BlockAttestation in another file.

src/providers/consensus/client.py
```diff
@@ -108,7 +110,9 @@ def get_block_details(self, state_id: SlotNumber | BlockRoot) -> BlockDetailsRes
...
+    ) -> list[BlockAttestationPhase0 | BlockAttestationElectra]:
```
F4ever, line 115: Actually it always returns BlockAttestation

src/modules/csm/checkpoint.py
```diff
@@ -229,18 +234,48 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
+    return _bytes_to_bool_list(bytes_)
 
 
-def _is_attested(bits: Sequence[bool], index: int) -> bool:
-    return bits[index]
+def hex_bitlist_to_list(bitlist: str) -> list[bool]:
+    bytes_ = hex_str_to_bytes(bitlist)
+    if not bytes_ or bytes_[-1] == 0:
+        raise ValueError(f"Got invalid {bitlist=}")
+    bitlist_len = int.from_bytes(bytes_, "little").bit_length() - 1
```
F4ever, line 274: why -1?

src/modules/csm/checkpoint.py
madlabman: `TypeGuard` really makes sense here, why to replace?

madlabman: > Maybe add this method to `BlockAttestation` class?

`BlockAttestation` is a pure type now.

src/modules/csm/checkpoint.py
```diff
@@ -229,18 +234,48 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
+def is_eip7549_attestation(attestation: BlockAttestation) -> TypeGuard[BlockAttestationElectra]:
+    # @see https://eips.ethereum.org/EIPS/eip-7549
+    has_committee_bits = getattr(attestation, "committee_bits") is not None
+    has_zero_index = attestation.data.index == "0"
+    if has_committee_bits and not has_zero_index:
+        raise ValueError(f"Got invalid {attestation=}")
+    return has_committee_bits and has_zero_index
+
+
+def get_set_indices(bits: Sequence[bool]) -> list[int]:
```
F4ever, line 258: https://github.com/ethereum/consensus-specs/blob/2c1f677187e6534aec77057a7d1cc746a40d3630/specs/electra/beacon-chain.md#get_committee_indices

src/modules/csm/checkpoint.py
madlabman: I've changed it a little bit, so it's gone now.

src/modules/csm/checkpoint.py
F4ever: Add this link pls

src/providers/consensus/client.py
madlabman: BlockAttestationResponse now

src/modules/csm/checkpoint.py
```diff
@@ -229,18 +234,48 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
+        committee_offset = 0
+        for committee_idx in get_committee_indices(attestation):
+            committee = committees.get((attestation.data.slot, committee_idx), [])
+            att_bits = hex_bitlist_to_list(attestation.aggregation_bits)[committee_offset:][: len(committee)]
+            for index_in_committee in get_set_indices(att_bits):
+                committee[index_in_committee].included = True
+            committee_offset += len(committee)
+
+
+def get_committee_indices(attestation: BlockAttestation) -> list[CommitteeIndex]:
```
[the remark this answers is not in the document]
madlabman: I think it can be added to the `BlockAttestationResponse` but maybe anytime we will really need this.

src/modules/csm/checkpoint.py
F4ever: This function returns bool. Or am I missing smth?

src/modules/csm/checkpoint.py
madlabman: I think it's not exactly what it does. This function is more generic.

F4ever (approved):

PR #568, opened 2024-11-28
Log get_validators_no_cache

hweawer:


src/providers/http_provider.py
```diff
@@ -93,7 +93,14 @@ def _get(
 
         for host in self.hosts:
             try:
-                return self._get_without_fallbacks(host, endpoint, path_params, query_params, stream)
+                data = self._get_without_fallbacks(host, endpoint, path_params, query_params, stream)
+                logger.debug({
+                    'msg': 'Get request',
+                    'url': endpoint,
+                    'data': f'{data}',
```
vgorkavenko, line 100: Data from `GET` endpoints can be quite large. Validators data and keys info can be as large as `300mb` (or even larger). Are we sure we want to print it? @F4ever.

hweawer: Logging only a single request, as discussed

src/providers/consensus/client.py
```diff
@@ -160,6 +160,7 @@ def get_validators(self, blockstamp: BlockStamp) -> list[Validator]:
     def get_validators_no_cache(self, blockstamp: BlockStamp, pub_keys: str | tuple | None = None) -> list[dict]:
         """Spec: https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidators"""
         try:
+            logger.info({'msg': 'Getting validators...'})
```
F4ever, line 163: blockstamp.slot_number
blockstamp.state_root
Add:
url without domen

F4ever: Move outside of try block

F4ever: Add final log with response details + with validators length

vgorkavenko (approved):

F4ever (approved):

PR #569, opened 2024-11-28
Catch encoding error

hweawer:


vgorkavenko (commented):
This is a `NotOkResponse` special case, right? Why do we want to log this separately?

PR #571, opened 2024-12-04
feat: account cl pending deposits validators

vgorkavenko:


src/constants.py
```diff
@@ -11,6 +11,7 @@
 # https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#gwei-values
 EFFECTIVE_BALANCE_INCREMENT = 2 ** 0 * 10 ** 9
 MAX_EFFECTIVE_BALANCE = 32 * 10 ** 9
+MIN_ACTIVATION_BALANCE = 32 * 10 ** 9
```
madlabman, line 14: Can we make both of these variables of type Gwei?

src/web3py/extensions/lido_validators.py
```diff
@@ -172,6 +173,14 @@ def merge_validators_with_keys(keys: list[LidoKey], validators: list[Validator])
 
         return lido_validators
 
+    @staticmethod
+    def calculate_pending_deposits_sum(lido_validators: list[LidoValidator]) -> int:
+        return sum(
+            MIN_ACTIVATION_BALANCE
```
madlabman, line 179: I think it makes sense to leave a note that it is currently true for Lido, but this may be subject to change once we implement compounding.

src/constants.py
```diff
@@ -1,3 +1,5 @@
+from src.types import Gwei
```
madlabman, line 1: Imported twice.

tests/web3_extentions/test_lido_validators.py
```diff
@@ -38,6 +39,24 @@ def test_get_lido_validators(web3, lido_validators, contracts):
...
+    validators.extend(ValidatorFactory.batch(30))
+    lido_keys = LidoKeyFactory.generate_for_validators(validators[:15])
+    lido_keys.extend(LidoKeyFactory.batch(10))
+
+    web3.lido_validators._kapi_sanity_check = Mock()
+
+    web3.cc.get_validators = Mock(return_value=validators)
+    web3.kac.get_used_lido_keys = Mock(return_value=lido_keys)
+
+    lido_validators = web3.lido_validators.get_lido_validators(blockstamp)
```
madlabman, line 54: Why not to build `lido_validators` directly?

madlabman (approved):

F4ever (approved):

Issue #407, opened 2023-10-06
ZeroDivisionError in new testnet


F4ever:
There are cases when Lido Oracle fails to do report
1. If there are no lido validators or zero balance on CL side
Then normal_report_cl_rebase will be 0 in bunker service.
`diff_current_with_normal = 1 - current_report_cl_rebase / normal_report_cl_rebase` - ZeroDivisionError here
2. If there are no validators at all.
Then total_predictable_validators_count will be zero in Ejector module.
`stake_volume = operator_state.predictable_validators_count / total_predictable_validators_count`

PR #572, opened 2024-12-09
feat: MIN_ACTIVATION_BALANCE, new correlation penalty

vgorkavenko:

PR #570, opened 2024-12-04
Support compound WC

hweawer:
https://github.com/ethereum/consensus-specs/blob/4dfd346acb4067212de033b8478f811279d543b4/specs/electra/beacon-chain.md

PR #561, opened 2024-11-06
feat: add test for modules on network fork

vgorkavenko:

PR #574, opened 2024-12-11
[EPIC] Oracle V5 (Pectra Duck)

F4ever:
**Pectra:**
- #570
- #572
- #571
- #556
- #577
- #579
- #575

**Improvement:**
- #573
- #568
- #522
- #521

**Bug fixes:**
- #557
- #537
- #523
- #582

**Tests:**
- #561 
- #530
 
P.S.
- Fixes https://github.com/lidofinance/lido-oracle/issues/407


src/modules/accounting/accounting.py
```diff
@@ -174,10 +173,8 @@ def _calculate_report(self, blockstamp: ReferenceBlockStamp):
         # or in the `execute_module` method
         if consensus_version == 1:
             report_data = self._calculate_report_v1(blockstamp)
-        elif consensus_version == 2:
-            report_data = self._calculate_report_v2(blockstamp)
         else:
-            raise IncompatibleException("Consensus version is not supported")
+            report_data = self._calculate_report_latest_version(consensus_version, blockstamp)
```
avsetsin, line 177: the previous solution was better
- branching happens at the top level
- exception is thrown explicitly if the contract has consensus version > method can handle

consider removing v1 support and leave branching here, smth like:

```python
if consensus_version == 2:
    report_data = self._calculate_report_v2(blockstamp)
elif consensus_version == 3:
    if not is_electra_activated:
        report_data = self._calculate_report_v3_before_electra(blockstamp)
    else:
        report_data = self._calculate_report_v3_after_electra(blockstamp)
else
    raise IncompatibleException("Consensus version is not supported")
```

src/constants.py
```diff
@@ -1,5 +1,8 @@
+from src.types import Gwei, SlotNumber
+
 # https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#misc
-FAR_FUTURE_EPOCH = 2 ** 64 - 1
+GENESIS_SLOT = SlotNumber(0)
```
avsetsin, line 4: Consider using `Slot` name for the type instead of `SlotNumber` to align with the CL spec

src/modules/accounting/accounting.py
```diff
@@ -214,21 +211,37 @@ def get_updated_modules_stats(
...
-        total_balance = Gwei(sum(int(validator.balance) for validator in lido_validators))
+        total_balance = sum(int(validator.balance) for validator in lido_validators)
+        logger.info({'msg': 'Calculate Lido active balance (in Gwei)', 'value': total_balance})
 
-        logger.info({'msg': 'Calculate lido state on CL. (Validators count, Total balance in gwei)', 'value': (count, total_balance)})
-        return ValidatorsCount(count), ValidatorsBalance(total_balance)
+        consensus_version = self.w3.lido_contracts.accounting_oracle.get_consensus_version(blockstamp.block_hash)
+        if consensus_version > 2:
+            spec = self.w3.cc.get_config_spec()
+            if blockstamp.ref_epoch >= int(spec.ELECTRA_FORK_EPOCH):
```
avsetsin, line 224: consider putting it in a separate method, like `is_electra_activated`, since it's widely used

src/modules/accounting/accounting.py
```diff
@@ -214,21 +211,37 @@ def get_updated_modules_stats(
...
 
-        count = len(lido_validators)
-        total_balance = Gwei(sum(int(validator.balance) for validator in lido_validators))
+        total_balance = sum(int(validator.balance) for validator in lido_validators)
+        logger.info({'msg': 'Calculate Lido active balance (in Gwei)', 'value': total_balance})
 
-        logger.info({'msg': 'Calculate lido state on CL. (Validators count, Total balance in gwei)', 'value': (count, total_balance)})
-        return ValidatorsCount(count), ValidatorsBalance(total_balance)
+        consensus_version = self.w3.lido_contracts.accounting_oracle.get_consensus_version(blockstamp.block_hash)
+        if consensus_version > 2:
```
avsetsin, line 222: 1. This code is harder to read and understand. Branching occurs deep in the method chain, which is hard to track. Consider having branching as high up in the chain as possible 
2. Consider using an exact version consensus check (`consensus_version == 3`)

src/modules/accounting/accounting.py
```diff
@@ -214,21 +211,37 @@ def get_updated_modules_stats(
...
+            spec = self.w3.cc.get_config_spec()
+            if blockstamp.ref_epoch >= int(spec.ELECTRA_FORK_EPOCH):
+                state = self.w3.cc.get_state_view(blockstamp)
+                pending_deposits = self.w3.lido_validators.calculate_pending_deposits_sum(
+                    lido_validators, state.pending_deposits
+                )
```
avsetsin, line 228: this code is misleading
- `pending_deposits` - contains the sum of only the part of pending_deposits belonging to lido validators 
- `calculate_pending_deposits_sum` - does not account for all pending_deposits

src/modules/accounting/accounting.py
```diff
@@ -214,21 +211,37 @@ def get_updated_modules_stats(
...
+                logger.info({'msg': 'Calculate Lido pending deposits (in Gwei)', 'value': pending_deposits})
```
avsetsin, line 229: Same here, the wording is incorrect. It's not Lido pending deposits, it's only a part of them. Please be careful with the terms you use in your code, as they can be misleading.

src/modules/accounting/accounting.py
```diff
@@ -366,13 +379,13 @@ def _calculate_report_v1(self, blockstamp: ReferenceBlockStamp) -> ReportData:
         extra_data_part_v1 = self._calculate_extra_data_report_v1(blockstamp)
         return self._combine_report_parts(1, blockstamp, rebase_part, modules_part, wq_part, extra_data_part_v1)
 
-    def _calculate_report_v2(self, blockstamp: ReferenceBlockStamp) -> ReportData:
+    def _calculate_report_latest_version(self, consensus_version: int, blockstamp: ReferenceBlockStamp) -> ReportData:
```
avsetsin, line 382: consider using exact version number instead of `latest_version` and do explicit checks for the consensus version to be sure that non updated oracles will not report after next consensus version bump

src/modules/ejector/ejector.py
```diff
@@ -203,23 +209,16 @@ def is_reporting_allowed(self, blockstamp: ReferenceBlockStamp) -> bool:
...
+                if is_fully_withdrawable_validator(v.validator, Gwei(int(v.balance)), on_epoch)
+            )
         )
 
-        return result
-
-    def _get_predicted_withdrawable_balance(self, validator: Validator) -> Wei:
-        return self.w3.to_wei(min(int(validator.balance), MAX_EFFECTIVE_BALANCE), 'gwei')
+    def _get_predicted_withdrawable_balance(self, validator: Validator) -> Gwei:
+        return Gwei(min(int(validator.balance), get_max_effective_balance(validator.validator)))
```
avsetsin, line 221: ⚠️  After these changes, `_get_predicted_withdrawable_balance` returns Gwei instead of Wei. 
The method used in `_get_predicted_withdrawable_epoch_post_electra` that expects Gwei and in `_get_withdrawable_lido_validators_balance` and in `_get_total_expected_balance` that expect Wei

avsetsin: Consider adding suffixes like `_gwei` and `_wei` to the names of such methods. This can help highlight such issues and catch them in advance

src/modules/ejector/ejector.py
```diff
@@ -247,10 +258,32 @@ def _get_predicted_withdrawable_epoch(
...
+            exit_balance_to_consume = per_epoch_churn
+        else:
+            earliest_exit_epoch = state_view.earliest_exit_epoch
+            exit_balance_to_consume = state_view.exit_balance_to_consume
+
+        exit_balance = sum(self._get_predicted_withdrawable_balance(v) for v in validators_to_eject)
+        balance_to_process = max(0, exit_balance - exit_balance_to_consume)
+        additional_epochs = math.ceil(balance_to_process / per_epoch_churn)
+
+        return EpochNumber(earliest_exit_epoch + additional_epochs + MIN_VALIDATOR_WITHDRAWABILITY_DELAY)
```
avsetsin, line 285: consider using the closest implementation of the method [`compute_exit_epoch_and_update_churn`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-compute_exit_epoch_and_update_churn) to the spec as well as a link to it. It helps a lot to keep track of changes on CL side.

Smth like (note, i didn't check the correctness off the code): 
```python
def _get_predicted_withdrawable_epoch_post_electra(
    self,
    blockstamp: ReferenceBlockStamp,
    validators_to_eject: list[Validator],
) -> EpochNumber:
    predicted_withdrawable_balance = sum(self._get_predicted_withdrawable_balance(v) for v in validators_to_eject)
    return compute_exit_epoch(state, predicted_withdrawable_balance)

def compute_exit_epoch(state: BeaconState, exit_balance: Gwei) -> Epoch:
    """"
    Modified version of `compute_exit_epoch_and_update_churn` from the spec without state mutation
    https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-compute_exit_epoch_and_update_churn
    """"
    earliest_exit_epoch = max(state.earliest_exit_epoch, compute_activation_exit_epoch(get_current_epoch(state)))
    per_epoch_churn = get_activation_exit_churn_limit(state)
    # New epoch for exits.
    if state.earliest_exit_epoch < earliest_exit_epoch:
        exit_balance_to_consume = per_epoch_churn
    else:
        exit_balance_to_consume = state.exit_balance_to_consume

    # Exit doesn't fit in the current earliest epoch.
    if exit_balance > exit_balance_to_consume:
        balance_to_process = exit_balance - exit_balance_to_consume
        additional_epochs = (balance_to_process - 1) // per_epoch_churn + 1
        earliest_exit_epoch += additional_epochs
        exit_balance_to_consume += additional_epochs * per_epoch_churn

    # The state mutation is removed from the original method

    # Consume the balance and update state variables.
    # state.exit_balance_to_consume = exit_balance_to_consume - exit_balance
    # state.earliest_exit_epoch = earliest_exit_epoch

    return earliest_exit_epoch
```

src/modules/ejector/ejector.py
```diff
@@ -285,36 +318,47 @@ def _get_latest_exit_epoch(self, blockstamp: ReferenceBlockStamp) -> tuple[Epoch
     @lru_cache(maxsize=1)
     def _get_sweep_delay_in_epochs(self, blockstamp: ReferenceBlockStamp) -> int:
         """Returns amount of epochs that will take to sweep all validators in chain."""
+        spec = self.w3.cc.get_config_spec()
         chain_config = self.get_chain_config(blockstamp)
-        total_withdrawable_validators = self._get_total_withdrawable_validators(blockstamp)
+        current_epoch = epoch_from_slot(blockstamp.slot_number, chain_config.slots_per_epoch)
```
avsetsin, line 323: ⚠️ we should use reference epoch here, not the current one

src/modules/ejector/ejector.py
```diff
@@ -285,36 +318,47 @@ def _get_latest_exit_epoch(self, blockstamp: ReferenceBlockStamp) -> tuple[Epoch
...
+        electra_epoch = int(spec.ELECTRA_FORK_EPOCH)
+        if self.get_consensus_version(blockstamp) in (1, 2) or current_epoch < electra_epoch:
+            return self._get_sweep_delay_in_epochs_pre_pectra(blockstamp)
```
avsetsin, line 326: 1. consider using branching on the top level of the methods chain
2. consider using a common method for the detection of the hardfork activation

README.md
```diff
@@ -3,7 +3,8 @@
 [![License: GPL v3](https://img.shields.io/badge/License-GPLv3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)
 [![Tests](https://github.com/lidofinance/lido-oracle/workflows/Tests/badge.svg?branch=daemon_v2)](https://github.com/lidofinance/lido-oracle/actions)
 
-Oracle daemon for Lido decentralized staking service: Monitoring the state of the protocol across both layers and submitting regular update reports to the Lido smart contracts.
+Oracle daemon for Lido decentralized staking service: Monitoring the state of the protocol across both layers and submitting regular update
+reports to the Lido smart contracts.
```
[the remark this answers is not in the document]
F4ever: Fixed

src/modules/accounting/accounting.py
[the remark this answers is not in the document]
F4ever: Postponed to V6. Created task

src/constants.py
F4ever: SlotNumber is more obvious name. Decided to use old version

src/modules/accounting/accounting.py
F4ever: Removed V1 support.
Better to not directly compare to 3. Because there will be a lot of changes, each new consensus version.
Branching at top level makes it hard to implement this thing and keep everything in mind.

src/modules/accounting/accounting.py
F4ever: https://github.com/lidofinance/lido-oracle/pull/574#discussion_r1928489293

vgorkavenko (approved):

PR #577, opened 2024-12-13
Updated sweep calculation

madlabman:
Thing to consider: the spec-like logic is heavy, but has almost the same results as the current approach with no pending withdrawals; pending withdrawals have meaningful effect when there are thousands of them which doesn't seem to me as a case for mainnet.

It seems the old algo has an error due to rounding down via `int()`.

Next steps: I would decide to go or not to go with the new algo. From my perspective it's better to abandon it and use the old approach with an updated multiplier to cover the case with pending withdrawals.


src/modules/ejector/ejector.py
```diff
@@ -286,19 +279,31 @@ def _get_latest_exit_epoch(self, blockstamp: ReferenceBlockStamp) -> tuple[Epoch
     def _get_sweep_delay_in_epochs(self, blockstamp: ReferenceBlockStamp) -> int:
```
F4ever, line 279: Maybe just split function into two
`_get_sweep_delay_in_epochs_v1`
`_get_sweep_delay_in_epochs_v3`

src/modules/ejector/ejector.py
```diff
@@ -286,19 +279,31 @@ def _get_latest_exit_epoch(self, blockstamp: ReferenceBlockStamp) -> tuple[Epoch
...
-        return total_withdrawable_validators
+        slots_to_sweep = math.ceil(total_withdrawable_validators / MAX_WITHDRAWALS_PER_PAYLOAD)
+        full_sweep_in_epochs = math.ceil(slots_to_sweep / chain_config.slots_per_epoch)
+        return math.ceil(full_sweep_in_epochs * self.AVG_EXPECTING_WITHDRAWALS_SWEEP_DURATION_MULTIPLIER)
+
+    def _get_withdrawable_validators(self, blockstamp: ReferenceBlockStamp) -> list[Validator]:
+        return [
+            v
+            for v in self.w3.cc.get_validators(blockstamp)
+            if is_partially_withdrawable_validator(v) or is_fully_withdrawable_validator(v, blockstamp.ref_epoch)
```
F4ever, line 305: Would we like to tweak the `is_partially_withdrawable_validator` function a bit to take into account validators with balances 32 ETH exactly?

hweawer (approved):

src/modules/submodules/consensus.py
```diff
@@ -87,6 +87,10 @@ def _get_consensus_contract_members(self, blockstamp: BlockStamp):
         consensus_contract = self._get_consensus_contract(blockstamp)
         return consensus_contract.get_members(blockstamp.block_hash)
 
+    @lru_cache(maxsize=1)
+    def consensus_version(self, blockstamp: BlockStamp):
```
F4ever, line 91: consensus_version -> get_consensus_version

F4ever (approved):

PR #582, opened 2025-01-07
fix: Prysm specific error text

vgorkavenko:
Due to the new error text
```
Response from eth/v1/beacon/states/0x448ed6cac5963b47566ee7fac30f649e09ac87da3d34e2613d7f224352e735ea/validators [404] with text: \"{\"message\":\"State not found\",\"code\":404}\" returned.\n"}
```

### DIFF
--- a/src/providers/consensus/client.py
+++ b/src/providers/consensus/client.py
@@ -181,7 +181,7 @@ class ConsensusClient(HTTPProvider):
 
             raise error
 
-    PRYSM_STATE_NOT_FOUND_ERROR = 'State not found: state not found in the last'
+    PRYSM_STATE_NOT_FOUND_ERROR = 'State not found'
 
     def __raise_on_prysm_error(self, errors: list[Exception]) -> Exception | None:
         """


src/providers/consensus/client.py
```diff
@@ -181,7 +181,7 @@ def get_validators_no_cache(self, blockstamp: BlockStamp, pub_keys: str | tuple
 
             raise error
 
-    PRYSM_STATE_NOT_FOUND_ERROR = 'State not found: state not found in the last'
+    PRYSM_STATE_NOT_FOUND_ERROR = 'State not found'
```
madlabman, line 184: It doesn't seem to be unique for Prysm though. [Lodestar](https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/api/impl/errors.ts#L5-L9) and [Nimbus](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/era_db.nim#L379) can respond with the same error.

Maybe it makes sense to raise a specific exception like `StateNotFound` and replace `*_with_prysm` methods with the corresponding `*_by_slot_number` ones?

vgorkavenko: Yep. Will test it locally

vgorkavenko: Lodestar responds `404` with empty body in case of unindexed state and with `incorrect string length` for indexed `validators` state, so I would disregard it at all because the client doesn't work for us.

Nimbus respond with `{"code":500,"message":"Not implemented yet"}` for requests with `state_root` and with `{"code":404,"message":"Could not get requested state"}` for requests on unindexed states. I've tested it on local instance, but you can try the same on [public ones](https://nimbus.guide/rest-api.html)

madlabman (approved):

F4ever (approved):

PR #583, opened 2025-01-09
feat: support Kubo provider

madlabman:


src/providers/ipfs/kubo.py
```diff
@@ -0,0 +1,83 @@
...
+from .cid import CID
+from .types import FetchError, IPFSProvider, PinError, UploadError
+
+logger = logging.getLogger(__name__)
+
+
+class Kubo(IPFSProvider):
+    """Client for [Kubo](https://github.com/ipfs/kubo) IPFS"""
+
+    PORT_GATEWAY = 8080
```
F4ever, line 17: I think this should be part of `host`

src/providers/ipfs/kubo.py
```diff
@@ -0,0 +1,83 @@
...
+            return resp.json()["Hash"]
+        except JSONDecodeError as ex:
+            raise UploadError from ex
+        except KeyError as ex:
+            raise UploadError from ex
+
+    def pin(self, cid: CID) -> None:
+        # @see https://docs.ipfs.tech/reference/kubo/rpc/#api-v0-pin-add
+
+        url = f"{self.host}:{self.PORT_RPC}/api/v0/pin/add"
```
F4ever, line 67: Can we make DRY here.
Seems exception catching and json parsing are same with `_upload` func

src/providers/ipfs/kubo.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            resp = requests.post(url, params={"arg": str(cid)}, timeout=self.timeout)
+            resp.raise_for_status()
+        except requests.RequestException as ex:
+            logger.error({"msg": "Request has been failed", "error": str(ex)})
+            raise PinError(cid) from ex
+
+        try:
+            pinned = resp.json()["Pins"][0]
```
F4ever, line 76: What are elements in the list, except first one

kubo/README.md
```diff
@@ -0,0 +1,21 @@
+## Kubo distribution for lido-oracle
+
+### Init scripts
+
+Kubo [executes
+scripts](https://github.com/ipfs/kubo/blob/fd50eb0fc385ec35cc2269646182920849b3c9b5/bin/container_daemon#L53)
```
F4ever, line 6:
```suggestion
Kubo [executes scripts](https://github.com/ipfs/kubo/blob/fd50eb0fc385ec35cc2269646182920849b3c9b5/bin/container_daemon#L53)
```

kubo/020-pin-cids.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/env sh
+
+set -eu
```
F4ever, line 3: How long it takes? What if Kubo already stores this cid?

As I understand correctly, failing of this script prevents Kubo to startup?

kubo/README.md
```diff
@@ -0,0 +1,21 @@
...
+found in the `/container-init.d/` in alphabetical order.
+
+#### `./010-set-conf.sh`
+
+- Updates configuration of Kubo node to get determinstic CIDs
+- Set ports to expected by the client
```
F4ever, line 12: let's move ports config outside, so guys can configure everything in their docker-compose files

src/main.py
```diff
@@ -166,6 +166,12 @@ def ipfs_providers() -> Iterator[IPFSProvider]:
             timeout=variables.HTTP_REQUEST_TIMEOUT_IPFS,
         )
 
+    if variables.KUBO_HOST:
```
F4ever, line 169: Maybe move it to the top?
And should require Kubo always?

kubo/README.md
madlabman: The ports are configured via env variables already.

kubo/020-pin-cids.sh
madlabman: > How long it takes?

The time to download CIDs depends on the count of files, network conditions, etc. In my tests it worked fast though.

> What if Kubo already stores this cid?

It will just move to the next one in the list.

> failing of this script prevents Kubo to startup

That's right. I think we can move the content of the script to a separate file and wrap a call to it to prevent kubo from failure on startup.

src/providers/ipfs/kubo.py
madlabman: There can be multiple CIDs, basically, in the case we provide multiple files to upload.

src/providers/ipfs/kubo.py
madlabman: There are subtle differences that makes a one-fits-all function not a good one.

src/providers/ipfs/kubo.py
madlabman: Both gateway and RPC are running on the same `host` but with different ports.

kubo/010-set-conf.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/env sh
+
+set -eu
+
+ipfs config --json Import.CidVersion 0
+ipfs config --json Import.UnixFSRawLeaves false
+ipfs config Import.UnixFSChunker size-262144
+ipfs config Import.HashFunction sha2-256
+
+ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/$GATEWAY_PORT
```
F4ever, line 10: lets provide this envs to Oracle

src/variables.py
```diff
@@ -14,6 +14,7 @@
 GW3_ACCESS_KEY: Final = from_file_or_env('GW3_ACCESS_KEY')
 GW3_SECRET_KEY: Final = from_file_or_env('GW3_SECRET_KEY')
 PINATA_JWT: Final = from_file_or_env('PINATA_JWT')
+KUBO_HOST: Final[str] = os.getenv('KUBO_HOST', '')
```
F4ever, line 17: add new vars to readme

src/variables.py
```diff
@@ -100,6 +101,8 @@ def check_uri_required_variables():
         errors.append('CONSENSUS_CLIENT_URI')
     if '' in KEYS_API_URI:
         errors.append('KEYS_API_URI')
+    if not KUBO_HOST:
```
F4ever, line 104: I just thought that this is not the best idea, at least for local development.
Also, Oracles in infra will be instantly broken

src/variables.py
madlabman: Done

kubo/010-set-conf.sh
madlabman: Provided via the corresponding `KUBO_` variables.

F4ever (approved):

kubo/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ipfs/kubo:v0.34.1
+
+ENV GATEWAY_PORT 8080
```
chasingrainbows, line 3: Better to use ARG here with some default value i guess

madlabman: ARG won't be persisted and I'm not sure it will be available as an env variable for scripts.

chasingrainbows: Maybe something like
```
ARG GATEWAY_PORT=8080
ARG API_PORT=5001
ARG CHAIN=mainnet

ENV GATEWAY_PORT=$GATEWAY_PORT
ENV API_PORT=$API_PORT
ENV CHAIN=$CHAIN
```
Why? Just because there is an option to configure this values with oracle envs + resolve ports conflict if needed

kubo/README.md
```diff
@@ -0,0 +1,21 @@
+## Kubo distribution for lido-oracle
```
chasingrainbows, line 1: Move it to docs folder please

kubo/Dockerfile
madlabman: Didn't get the point.

kubo/README.md
madlabman: Symlinked.

kubo/Dockerfile
chasingrainbows: > Didn't get the point.

You are right here, sry, forgot there is possbile to change env during container run without ARG, closing thread.

kubo/_pin-cids
```diff
@@ -0,0 +1,30 @@
...
+
+WGET_ARGS="--quiet --no-check-certificate"
+GATEWAY=${GATEWAY:-https://ipfs.io}
+
+echo "Fetching list of CIDs to pin"
+
+LIST=$(mktemp)
+trap "rm $LIST" EXIT
+
+wget https://raw.githubusercontent.com/lidofinance/csm-rewards/refs/heads/$CHAIN/cids -O$LIST $WGET_ARGS
```
chasingrainbows, line 13: With $CHAIN as "mainnet" it is return 404 currently, it is planned to add them later? With "artifact" it is returns some, i guess for local development?

src/providers/ipfs/kubo.py
```diff
@@ -0,0 +1,82 @@
...
+        super().__init__()
+        self.host = host
+        self.timeout = timeout
+        self.rpc_port = rpc_port
+        self.gateway_port = gateway_port
+
+    def fetch(self, cid: CID) -> bytes:
+        url = f"{self.host}:{self.gateway_port}/ipfs/{cid}"
+        try:
+            resp = requests.get(url, timeout=self.timeout)
```
chasingrainbows, line 36: Do we need some retries as well as timeout?

kubo/_pin-cids
madlabman: Yep, "artifacts" was to test only.

src/providers/ipfs/kubo.py
madlabman: See `MultiIPFSProvider` for retries.

chasingrainbows (approved):

PR #589, opened 2025-01-21
Refactor get state

madlabman:


src/providers/consensus/types.py
```diff
@@ -168,11 +151,43 @@ class SlotAttestationCommittee(FromResponse):
...
+@dataclass
+class PendingPartialWithdrawal(Nested):
+    validator_index: str
+    amount: Gwei
+    withdrawable_epoch: EpochNumber
+
+
 @dataclass
 class BeaconStateView(Nested, FromResponse):
     """A view to BeaconState with only the required keys presented"""
```
F4ever, line 172: add link to spec
https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#beaconstate

src/providers/consensus/types.py
```diff
@@ -168,11 +151,43 @@ class SlotAttestationCommittee(FromResponse):
...
 
     slot: SlotNumber
+    validators: list[ValidatorState]
+    balances: list[Gwei]
     # This fields are new in Electra, so here are default values for backward compatibility.
+    deposit_balance_to_consume: Gwei = Gwei(0)
```
F4ever, line 178: let's remove fields we don't use in Oracles

PR #599, opened 2025-01-24
feat: use is_electra_activated

madlabman:


vgorkavenko (approved):

src/modules/accounting/accounting.py
```diff
@@ -219,8 +219,7 @@ def _get_consensus_lido_state(self, blockstamp: ReferenceBlockStamp) -> tuple[Va
 
         consensus_version = self.w3.lido_contracts.accounting_oracle.get_consensus_version(blockstamp.block_hash)
```
F4ever, line 220: replace with self.get_consensus_version()

src/services/bunker_cases/midterm_slashing_penalty.py
```diff
@@ -236,7 +243,7 @@ def predict_midterm_penalty_in_frame_post_electra(
                 report_ref_epoch, all_slashed_validators, EpochNumber(midterm_penalty_epoch)
             )
 
-            if frame_ref_epoch < int(cl_spec.ELECTRA_FORK_EPOCH):
+            if not is_electra_activated(frame_ref_epoch):
```
F4ever, line 246: let's revers this one

PR #606, opened 2025-01-24
refactor: use numerical types for uints in beacon chain structures

madlabman:


src/services/safe_border.py
```diff
@@ -228,20 +228,18 @@ def _slashings_in_frame(self, frame: FrameNumber, slashed_pubkeys: set[str]) ->
         return len(slashed_validators) > 0
 
     def _filter_validators_with_earliest_exit_epoch(self, validators: list[Validator]) -> list[Validator]:
-        sorted_validators = sorted(validators, key=lambda validator: (int(validator.validator.exit_epoch)))
-        return filter_validators_by_exit_epoch(
-            sorted_validators, EpochNumber(int(sorted_validators[0].validator.exit_epoch))
-        )
+        sorted_validators = sorted(validators, key=lambda validator: (validator.validator.exit_epoch))
```
F4ever, line 231:
```suggestion
        sorted_validators = sorted(validators, key=lambda validator: validator.validator.exit_epoch)
```

src/utils/dataclass.py
```diff
@@ -71,7 +72,7 @@ def wrapper_decorator(*args, **kwargs):
                 raise DecodeToDataclassException(f'Type {type(list_of_elements)} is not supported.')
 
             if not list_of_elements:
-                return list_of_elements
+                return []
```
F4ever, line 75: list_of_elements could be a tuple
I propose to left it as it is

madlabman: In this case we have an incorrect typing. The function should return a list and we rely on this.

PR #607, opened 2025-01-28
[CSM] feat: new duties

vgorkavenko:


src/modules/csm/distribution.py
```diff
@@ -0,0 +1,222 @@
...
+
+            total_rewards_to_distribute = self.w3.csm.fee_distributor.shares_to_distribute(frame_blockstamp.block_hash)
+            rewards_to_distribute_in_frame = total_rewards_to_distribute - (total_distributed_rewards + total_rebate)
+
+            rewards_in_frame, log = self._calculate_distribution_in_frame(
+                frame, frame_blockstamp, rewards_to_distribute_in_frame, frame_module_validators
+            )
+
+            total_distributed_rewards += log.distributed_rewards
+            total_rebate += log.rebate_to_protocol
```
madlabman, line 51: `log` is a supplementary thing, I would avoid reading anything from it.

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,222 @@
...
+        network_perf = self._get_network_performance(frame)
+
+        for (_, no_id), validators in operators_to_validators.items():
+            logger.info({"msg": f"Calculating distribution for {no_id=}"})
+            log_operator = log.operators[no_id]
+
+            curve_id = self.w3.csm.accounting.get_bond_curve_id(no_id, blockstamp.block_hash)
+            perf_coeffs, perf_leeway, reward_share = self._get_curve_params(curve_id, blockstamp)
+
+            sorted_validators = sorted(validators, key=lambda v: v.index)
```
madlabman, line 102: The `validators` variable includes exited validators. Is it as intended?

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,222 @@
...
+        log_validator.performance = performance
+        log_validator.attestation_duty = attestation
+        if proposal:
+            log_validator.proposal_duty = proposal
+        if sync:
+            log_validator.sync_duty = sync
+
+        if performance > threshold:
+            # Count of assigned attestations used as a metrics of time the validator was active in the current frame.
+            # Reward share is a share of the operator's reward the validator should get. It can be less than 1.
```
madlabman, line 192: An example needed, e.g., given only 1 attestation made by a validator and `rewards_share < 1` we would end up sending everything to the treasury as a rebate. We cover this case in favor of operator.

If I get it right. If not, it's another reason to have an explainer here.

src/modules/csm/distribution.py
vgorkavenko: Result of refactoring. There should be active only

src/modules/csm/distribution.py
vgorkavenko: There should be `math.ceil`. I will fix it and add explainer

src/modules/csm/checkpoint.py
```diff
@@ -236,10 +299,119 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
+    @timeit(
+        lambda args, duration: logger.info(
+            {"msg": f"Sync Committee for epoch {args.epoch} prepared in {duration:.2f} seconds"}
+        )
+    )
+    def _prepare_sync_committee(
+        self, epoch: EpochNumber, duty_block_roots: list[SlotBlockRoot]
+    ) -> dict[SlotNumber, list[ValidatorDuty]]:
+
+        with lock:
```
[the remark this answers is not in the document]
vgorkavenko: This was made for avoid race condition in getting sync committee from network for close epochs

src/providers/consensus/client.py
```diff
@@ -115,10 +120,7 @@ def get_block_details(self, state_id: SlotNumber | BlockRoot) -> BlockDetailsRes
         return BlockDetailsResponse.from_response(**data)
 
     @lru_cache(maxsize=256)
-    def get_block_attestations(
-        self,
-        state_id: SlotNumber | BlockRoot,
-    ) -> list[BlockAttestation]:
+    def get_block_attestations_and_sync(self, state_id: SlotNumber | BlockRoot) -> tuple[list[BlockAttestation], SyncAggregate]:
```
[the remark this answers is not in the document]
vgorkavenko: A separate method is needed mainly because of `lru_cache` custom size

src/modules/csm/checkpoint.py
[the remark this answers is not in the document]
madlabman: I see. Ideally we need for a separate synchronization mechanism, but I'm good with this one, though we have to document it properly.

src/providers/consensus/client.py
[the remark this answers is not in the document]
madlabman: Attestations take the most of the block space, as I can see, so maybe to change the cache size for the `get_block_details`.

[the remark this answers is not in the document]
vgorkavenko: It will affect core part, so other oracles as well. I'd prefer to have separate method for our needs

src/modules/csm/distribution.py
vgorkavenko: Done

src/modules/csm/distribution.py
vgorkavenko: Fixed

src/modules/csm/csm.py
```diff
@@ -94,57 +91,48 @@ def build_report(self, blockstamp: ReferenceBlockStamp) -> tuple:
...
+            rewards_cid = self.publish_tree(rewards_tree)
+
+        if distribution.strikes:
+            strikes_tree = self.make_strikes_tree(distribution.strikes)
+            strikes_tree_root = strikes_tree.root
+            strikes_cid = self.publish_tree(strikes_tree)
+            if strikes_tree_root == last_report.strikes_tree_root:
+                logger.warning({"msg": "Strikes tree root is the same as the previous one"})
+            if strikes_cid == last_report.strikes_tree_cid:
+                logger.warning({"msg": "Strikes tree CID is the same as the previous one"})
```
madlabman, line 109: I don't think it worth a warning.

src/modules/csm/distribution.py
```diff
@@ -219,19 +240,25 @@ def process_validator_duties(
             #    87.55 ≈ 88 of 103 participation shares should be counted for the operator key's reward.
             #    The rest 15 participation shares should be counted for the protocol's rebate.
             #
-            participation_share = math.ceil(attestation.assigned * reward_share)
+            participation_share = math.ceil(duties.attestation.assigned * reward_share)
             participation_shares[validator.lido_id.operatorIndex] += participation_share
```
madlabman, line 244: What about to extract the `participation_share` part to the `ValidatorDutyOutcome` and do the modification of shares mapping at the outer level?

madlabman (approved):

PR #611, opened 2025-01-30
chore: check consensus version to enable EIP7549 support

madlabman:


src/modules/csm/checkpoint.py
```diff
@@ -227,8 +237,14 @@ def _prepare_committees(self, epoch: EpochNumber) -> Committees:
...
 
 
-def process_attestations(attestations: Iterable[BlockAttestation], committees: Committees) -> None:
+def process_attestations(
+    attestations: Iterable[BlockAttestation],
+    committees: Committees,
+    eip7549_supported: bool = True,
+) -> None:
     for attestation in attestations:
+        if is_eip7549_attestation(attestation) and not eip7549_supported:
```
F4ever, line 246: Add tests to check this condition

madlabman: Done.

PR #612, opened 2025-01-30
fix: check consensus version on state migration

madlabman:


src/modules/csm/state.py
```diff
@@ -102,7 +104,16 @@ def add_processed_epoch(self, epoch: EpochNumber) -> None:
...
         logger.info({"msg": f"Processed {len(self._processed_epochs)} of {len(self._epochs_to_process)} epochs"})
 
-    def migrate(self, l_epoch: EpochNumber, r_epoch: EpochNumber):
+    def migrate(self, l_epoch: EpochNumber, r_epoch: EpochNumber, consensus_version: int):
+        if consensus_version != self._consensus_version:
+            logger.warning(
+                {
+                    "msg": f"Cache was built for consensus version {self._consensus_version}. "
+                    f"Discarding data to migrate to consensus version {consensus_version}"
+                }
```
F4ever, line 113:
```suggestion
              logger.warning(
                  {
                      "msg": f"Cache was built for consensus version {self._consensus_version}. "
                      f"Discarding data to migrate to the consensus version {consensus_version}"
                  }
```

src/modules/csm/state.py
```diff
@@ -102,7 +104,16 @@ def add_processed_epoch(self, epoch: EpochNumber) -> None:
     def log_progress(self) -> None:
         logger.info({"msg": f"Processed {len(self._processed_epochs)} of {len(self._epochs_to_process)} epochs"})
 
-    def migrate(self, l_epoch: EpochNumber, r_epoch: EpochNumber):
+    def migrate(self, l_epoch: EpochNumber, r_epoch: EpochNumber, consensus_version: int):
+        if consensus_version != self._consensus_version:
+            logger.warning(
+                {
+                    "msg": f"Cache was built for consensus version {self._consensus_version}. "
+                    f"Discarding data to migrate to consensus version {consensus_version}"
```
F4ever, line 112:
```suggestion
                    f"Discarding data to migrate to the consensus version {consensus_version}"
```

src/modules/csm/state.py
```diff
@@ -102,7 +104,16 @@ def add_processed_epoch(self, epoch: EpochNumber) -> None:
...
+                }
+            )
+            self.clear()
```
F4ever, line 115: Maybe rename clear to `discard_collected_data`?

src/modules/csm/state.py
```diff
@@ -49,6 +49,8 @@ class State:
     _epochs_to_process: tuple[EpochNumber, ...]
     _processed_epochs: set[EpochNumber]
 
+    _consensus_version: int = 1
```
F4ever, line 52: Maybe setup it on initialization in `__init__`?

madlabman: Need to check backward compatibilty first.

src/modules/csm/state.py
madlabman: `clear` is fine because the function clears all the state.

src/modules/csm/state.py
madlabman: ChatGPT says no "the"

src/modules/csm/state.py
madlabman: `__init__` is not called upon restoring object via pickle. So we will get an `AttributeError` on access to the variable.

PR #616, opened 2025-02-03
Fetch constants latest block

hweawer:


F4ever:
All constant changes should considered to be after report and before ref_slot of next frame

PR #625, opened 2025-02-13
Web3py7

hweawer:


src/web3py/middleware.py
```diff
@@ -1,19 +1,103 @@
...
 from web3_multi_provider import NoActiveProviderError
 
 from src.metrics.prometheus.basic import EL_REQUESTS_DURATION
-from web3 import Web3
 
 logger = logging.getLogger(__name__)
 
 
+class Web3MetricsMiddleware(Web3Middleware):
+    contracts = []
```
chasingrainbows, line 19: Could `contracts ` could be declared in `__init__ `? Im not sure, but looks like `contracts` as class field could potentialy lead to memory leaks in case of multiple calls of `__init__ `.

src/web3py/middleware.py
```diff
@@ -1,19 +1,103 @@
...
+
+    def __init__(self, w3: Union["AsyncWeb3", "Web3"]):
+        super().__init__(w3)
+        abi_dir = './assets/'
```
chasingrainbows, line 23: [nit] Could be a constant across project

src/web3py/middleware.py
```diff
@@ -1,19 +1,103 @@
...
+                    )
+                    raise
+
+                # https://www.jsonrpc.org/specification#error_object
+                error = response.get("error")
+                code: int = 0
+                if isinstance(error, dict):
+                    code = error.get("code") or code
+
+                t.labels(
```
chasingrainbows, line 84: Looks like this call could be under `with EL_REQUESTS_DURATION.time() as t: `

tests/factory/bitarrays.py
```diff
@@ -1,41 +1,32 @@
...
 
-class BitListFactory(ModelFactory):
-    __model__ = BitList
-
+class BitListFactory(ModelFactory[BitList]):
     @classmethod
     def build(
         cls,
-        factory_use_construct: bool = False,
         set_indices: list[int] = [],
```
chasingrainbows, line 18: ``Looks like a classic python bug,` = []` will declare mutable list that persistent across method calls

i guess it would work like

a = BitListFactory.build(set_indices=[1, 2])
b = BitListFactory.build(set_indices=[3])
print(a)  # [True, True, False, False]  (ok)
print(b)  # [True, True, False, True]   (not ok, [False, False, False, True] expected)

need something kinda
```
def build(cls, set_indices: list[int] = None, bits_count: int = 0) -> BitList:
    set_indices = set_indices or []
```

chasingrainbows (approved):

pyproject.toml
```diff
@@ -23,27 +23,28 @@ timeout-decorator = "^0.5.0"
...
 pytest-cov = "^4.0.0"
-pydantic-factories = "^1.17.2"
 # {{{ stubs for mypy
 types-requests = "^2.28.11.15"
 types-setuptools = "^67.6.0.0"
 types-urllib3 = "^1.26.25.8"
 # }}}
 hypothesis = "^6.68.2"
-black = "^24.8"
+black = "^23.12.1"
```
F4ever, line 47: why black was downgraded?

src/modules/csm/checkpoint.py
```diff
@@ -121,6 +121,7 @@ def __init__(
         state: State,
         converter: Web3Converter,
         finalized_blockstamp: BlockStamp,
+        *,
```
F4ever, line 124: Why do we need to use * here

src/web3py/extensions/fallback.py
```diff
@@ -1,13 +1,19 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
     def get_all_providers(self) -> list[Any]:
         return self._providers  # type: ignore[attr-defined]
 
     def _get_chain_id_with_provider(self, provider_index: int) -> int:
-        return Web3.to_int(hexstr=self._providers[provider_index].make_request("eth_chainId", []).get('result'))  # type: ignore[attr-defined]
+        return Web3.to_int(
+            hexstr=self._providers[provider_index].make_request("eth_chainId", []).get('result'))  # type: ignore[attr-defined]
```
F4ever, line 19:
```suggestion
            hexstr=self._providers[provider_index].make_request("eth_chainId", []).get('result')  # type: ignore[attr-defined]
        )  
```

F4ever, line 19: Pls, split this line. Smth like
```python
func(
    somestaff
)
```

tests/factory/bitarrays.py
```diff
@@ -1,41 +1,33 @@
...
-        return model(
-            __root__=get_serialized_bytearray(
-                bit_list,
-                bits_count=bits_count or len(bit_list),
-                extra_byte=True,
-            )
+        set_indices = set_indices or []
+        bit_list: list[bool] = [False] * (max(set_indices, default=0) + 1)
+        for n in set_indices:
+            bit_list[n] = True
```
F4ever, line 24: 👍

F4ever (approved):

PR #633, opened 2025-02-19
[WIP] CSM v2

vgorkavenko:


src/modules/csm/state.py
```diff
@@ -135,41 +238,54 @@ def validate(self, l_epoch: EpochNumber, r_epoch: EpochNumber) -> None:
...
-        return not self.unprocessed_epochs
+        return ValidatorDuties(att_duty, prop_duty, sync_duty)
 
-    @property
-    def frame(self) -> tuple[EpochNumber, EpochNumber]:
-        if not self._epochs_to_process:
-            raise ValueError("Epochs to process are not set")
-        return min(self._epochs_to_process), max(self._epochs_to_process)
+    def get_att_network_aggr(self, frame: Frame) -> DutyAccumulator:
+        # TODO: exclude `active_slashed` validators from the calculation
```
chasingrainbows, line 255: Is it still actual?

vgorkavenko: Yes. We have PoC [here](https://github.com/lidofinance/lido-oracle/pull/669). Probably, will be accounted in v3

src/modules/csm/checkpoint.py
```diff
@@ -190,52 +229,183 @@ def _process(self, unprocessed_epochs: list[EpochNumber], duty_epochs_roots: dic
...
-                for validator_duty in committee:
-                    self.state.inc(
-                        validator_duty.index,
-                        included=validator_duty.included,
-                    )
             if duty_epoch not in self.state.unprocessed_epochs:
                 raise ValueError(f"Epoch {duty_epoch} is not in epochs that should be processed")
+            for att_committee in att_committees.values():
+                for att_duty in att_committee:
+                    self.state.increment_att_duty(
```
chasingrainbows, line 261: Doesn't look very optimal. We will call self.find_frame(epoch) multiple times inside for the same duty_epoch. By the way, I'm not sure if performance is critical here (I just noticed there is already threading + timeit for measuring used, so, maybe it is critical).

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,290 @@
...
+    total_rewards_map: defaultdict[NodeOperatorId, Shares]
+    total_rebate: Shares
+    strikes: dict[StrikesValidator, StrikesList]
+    logs: list[FramePerfLog]
+
+    def __init__(self, w3: Web3, converter: Web3Converter, state: State):
+        self.w3 = w3
+        self.converter = converter
+        self.state = state
+        # Distribution results
```
chasingrainbows, line 43: Do we really need to store distribution results within the same object? In my opinion, it would be better to go with stateless approach using a dedicated result DTO (e.g., defined with a dataclass). It is will be easer to mock/maintain/read/etc.

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,290 @@
...
+            #    it can be less than 1 due to the value from `CSParametersRegistry`.
+            #    In case of decimal value, the reward should be rounded up in favour of the operator.
+            #
+            #    Example:
+            #     - Validator was 103 epochs active in the frame (assigned 103 attestations)
+            #     - Reward share for this Operator's key is 0.85
+            #    87.55 ≈ 88 of 103 participation shares should be counted for the operator key's reward.
+            #    The rest 15 participation shares should be counted for the protocol's rebate.
+            #
+            participation_share = math.ceil(duties.attestation.assigned * reward_share)
```
chasingrainbows, line 235: math.ceil rounds a number UP to the nearest integer, favoring node operators (e.g., 100 * 0.851 = 85.1 → 86). Float precision can also mess things up.

Why not use decimal.Decimal for exact calculations, rounding down to favor the protocol? For example something like int((Decimal(duties.attestation.assigned) * Decimal(str(reward_share))).quantize(Decimal('0.01'), ROUND_FLOOR))

src/modules/csm/checkpoint.py
vgorkavenko: It should be `O(1)` due to `lru_cache` that wraps `find_frame`

src/modules/csm/distribution.py
vgorkavenko: Rounding UP in favor of the node operator was done in purpose. Just to avoid getting the result `(1 * 0.99) = 0` here. So at least `1` share will be for the node operator.

We don't expect much accuracy here. I think `math.ceil` is enough for our purpose.

chasingrainbows: > Rounding UP in favor of the node operator was done in purpose

Woah, that good, maybe we could leave here some comment, like "Rounding UP in favor of the node operator was done in purpose". Rounding can be source of leaking/earning money on large scale.

src/modules/csm/checkpoint.py
chasingrainbows: Didnt notice, thx

src/providers/execution/contracts/cs_parameters_registry.py
```diff
@@ -0,0 +1,150 @@
...
+from src.modules.csm.state import ValidatorDuties
+from src.providers.execution.base_interface import ContractInterface
+from src.utils.cache import global_lru_cache as lru_cache
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class PerformanceCoefficients:
+    attestations_weight: int = 54
```
chasingrainbows, line 16: Seems, more convenient to move it into src/constants.py.

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,290 @@
...
+            # is not presented in the aggregates (e.g. exited, pending for activation etc).
+            return ValidatorDutiesOutcome(participation_share=0, rebate_share=0, strikes=0)
+
+        log_validator = log_operator.validators[validator.index]
+
+        if validator.validator.slashed is True:
+            # It means that validator was active during the frame and got slashed and didn't meet the exit
+            # epoch, so we should not count such validator for operator's share.
+            log_validator.slashed = True
+            return ValidatorDutiesOutcome(participation_share=0, rebate_share=0, strikes=1)
```
chasingrainbows, line 208: Looks like slashed validators and low performers get the same punishment: `strikes=1`, `participation_share=0`, `rebate_share=0`. 

BUT IMO slashing is way worse than just underperforming. 

Should we give slashed validators more strikes (like 3) or send their share to the protocol? Feels weird to treat them the same as laggy validators. Thoughts?

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+            raise ValueError(f"Invalid rewards to distribute: {rewards_to_distribute=}")
+
+        rewards_distribution: dict[NodeOperatorId, int] = defaultdict(int)
+        total_shares = rebate_share + sum(participation_shares.values())
+
+        for no_id, no_participation_share in participation_shares.items():
+            if no_participation_share == 0:
+                # Skip operators with zero participation
+                continue
+            rewards_distribution[no_id] = rewards_to_distribute * no_participation_share // total_shares
```
chasingrainbows, line 259: Just in case, should we check that total_shares >=0 to dodge zero division? Looks like this is possible if all validators was slashed or not assigned (huh, near to impossible, but why not)

chasingrainbows, line 259: Also, the `//` rounds down, so we might lose rewards (e.g., `100 * 1 // 3 = 33`). Why not use `Decimal` for example and distribute the remainder? Or even round up? If it  was done in purpose, better to leave some comment.

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+
+        return rewards_distribution
+
+    @staticmethod
+    def validate_distribution(total_distributed_rewards, total_rebate, total_rewards_to_distribute):
+        if (total_distributed_rewards + total_rebate) > total_rewards_to_distribute:
```
chasingrainbows, line 265: Why not `if total_distributed_rewards + total_rebate != total_rewards_to_distribute:`? To make sure all rewards are distributed?

What happens to the undistributed reward? Is it supposed to remain in the protocol?

vgorkavenko: It is okay when `total_distributed_rewards + total_rebate` less than `total_rewards_to_distribute `. If no distribution at all in the current frame (for example, the whole module is under performance threshold), we should keep the shares on the contract to distribute them in the next report

src/modules/csm/distribution.py
vgorkavenko: Here we calculate the amount of `stETH` shares (they can't be decimal) and perform the calculations as if it were on-chain. If we round it up, the total math won't work because of the wrong `sum(rewards_distribution)` value

vgorkavenko: Not sure. Looks like this is impossible due to the math above. `no_participation_share` should be non-zero to reach  this calculation but `total_shares` depends on sum of `no_participation_share`

src/modules/csm/distribution.py
vgorkavenko: We don't have some business logic for this yet. Hence, there is no sense to change here smth

.github/workflows/mainnet_fork_tests.yml
```diff
@@ -21,10 +22,42 @@ permissions:
...
     steps:
       - uses: actions/checkout@v3
 
+      # TODO: Remove after upgrade to CSM v2 on Mainnet.
+      - name: Checkout CSM repo
+        uses: actions/checkout@v4
+        with:
+          repository: 'lidofinance/community-staking-module'
+          ref: 'develop'
+          path: 'testruns/community-staking-module'
```
madlabman, line 36:
```suggestion
        with:
          repository: 'lidofinance/community-staking-module'
          ref: 'develop'
          path: 'testruns/community-staking-module'
          persist-credentials: false
```

src/modules/csm/helpers/last_report.py
```diff
@@ -0,0 +1,94 @@
...
+            blockstamp,
+            rewards_tree_root,
+            strikes_tree_root,
+            rewards_tree_cid,
+            strikes_tree_cid,
+        )
+
+    @cached_property
+    def rewards(self) -> Iterable[RewardsTreeLeaf]:
+        if (self.rewards_tree_cid is None) != (self.rewards_tree_root == ZERO_HASH):
```
chasingrainbows, line 54: This checks already impemented in load, does it make sense to check it once again in both properties?

src/providers/execution/contracts/cs_parameters_registry.py
```diff
@@ -0,0 +1,150 @@
...
+    perf_leeway_data: IntervalMapping
+    reward_share_data: IntervalMapping
+    strikes_params: StrikesParams
+
+
+class CSParametersRegistryContract(ContractInterface):
+    abi_path = "./assets/CSParametersRegistry.json"
+
+    @lru_cache()
+    def get_performance_coefficients(
```
chasingrainbows, line 81: Dont see any tests for this one

tests/fork/test_csm_oracle_cycle.py
```diff
@@ -13,8 +18,88 @@ def hash_consensus_bin():
...
+
+    with LockedDir(testruns_folder_path):
+        if not os.path.exists(csm_repo_path / ".prepared"):
+            logger.info("TESTRUN Cloning community-staking-module repo")
+            subprocess.run(
+                ["git", "clone", "https://github.com/lidofinance/community-staking-module", csm_repo_path], check=True
+            )
+            os.chdir(csm_repo_path)
+            subprocess.run(["git", "checkout", "develop"], check=True)
+            subprocess.run(["just", "deps"], check=True)
```
chasingrainbows, line 45: need some docs, how to run these tests locally

tests/fork/test_csm_oracle_cycle.py
```diff
@@ -13,8 +18,88 @@ def hash_consensus_bin():
...
+
+
+@pytest.fixture(scope='session')
+def prepared_csm_repo(testruns_folder_path, csm_repo_path):
+
+    if os.environ.get("GITHUB_ACTIONS") == "true":
+        # CI should have the repo cloned and prepared
+        if os.path.exists(csm_repo_path):
+            return csm_repo_path
+        raise ValueError("No cloned community-staking-module repo found, but running in CI. Fix the workflow.")
```
chasingrainbows, line 33: Actually, i belive, this could be implemented though docker-compose and separate dockerfile, to unify testing process between CI and local machine.

chasingrainbows: Ideally:
1. build csm container
2. build oracle container
3. share data within common persistent volume in docker-compose
4. run tests

tests/fork/test_csm_oracle_cycle.py
```diff
@@ -13,8 +18,88 @@ def hash_consensus_bin():
...
+        logger.info("TESTRUN Deployed CSM v2")
+
+    logger.info("TESTRUN Updating to CSM v2")
+    with subprocess.Popen(
+        ['just', "vote-upgrade"],
+        env={
+            **os.environ,
+            'CHAIN': chain,
+            "ANVIL_PORT": str(anvil_port),
+            "RPC_URL": f"http://127.0.0.1:{anvil_port}",  # FIXME: actually unused by the script, remove when fixed
```
chasingrainbows, line 86: `# FIXME ` is it still actual?

tests/fork/test_csm_oracle_cycle.py
```diff
@@ -13,8 +18,88 @@ def hash_consensus_bin():
...
+        },
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.STDOUT,
+    ) as process:
+        process.wait()
+        assert process.returncode == 0, "Failed to update to CSM v2"
+        logger.info("TESTRUN Updated to CSM v2")
+
+    os.chdir(original_dir)
+    # TODO: update ABIs in `assets` folder?
```
chasingrainbows, line 97: is it still actual?

tests/fork/test_csm_oracle_cycle.py
```diff
@@ -47,7 +132,11 @@ def missed_initial_frame(frame_config: FrameConfig):
 )
 @pytest.mark.parametrize(
     'running_finalized_slots',
-    [start_before_initial_epoch, start_after_initial_epoch, missed_initial_frame],
+    [
+        start_before_initial_epoch,
+        # start_after_initial_epoch,
```
chasingrainbows, line 137: Why commented?

tests/integration/conftest.py
```diff
@@ -117,3 +124,62 @@ def burner_contract(web3_provider_integration, lido_locator_contract):
...
+# ║                                          CSM contracts                                           ║
+# ╚══════════════════════════════════════════════════════════════════════════════════════════════════╝
+
+
+@pytest.fixture
+def cs_module_contract(web3_provider_integration):
+    return get_contract(
+        web3_provider_integration,
+        CSModuleContract,
+        "0xdA7dE2ECdDfccC6c3AF10108Db212ACBBf9EA83F",  # mainnet deploy
```
chasingrainbows, line 139: If u want to bound any test to mainnet, u could use @pytest.mark.mainnet

tests/fork/test_csm_oracle_cycle.py
```diff
@@ -13,8 +18,88 @@ def hash_consensus_bin():
...
+            subprocess.run(["just", "deps"], check=True)
+            subprocess.run(["just", "build"], check=True)
+            subprocess.run(["touch", ".prepared"], check=True)
+            os.chdir(original_dir)
+
+    return csm_repo_path
+
+
+@pytest.fixture()
+def update_csm_to_v2(accounts_from_fork, forked_el_client: Web3, anvil_port: int, prepared_csm_repo: Path):
```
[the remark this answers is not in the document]
vgorkavenko: It deploys new contract implementations and updates existing proxies. The addresses remain the same

tests/fork/test_csm_oracle_cycle.py
vgorkavenko: reverted

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+from src.utils.web3converter import Web3Converter
+from src.web3py.extensions.lido_validators import LidoValidator, ValidatorsByNodeOperator
+from src.web3py.types import Web3
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ValidatorDutiesOutcome:
+    participation_share: int
```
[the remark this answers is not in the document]
vgorkavenko: It is not the same, but will think about adding another type

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+        log_operator: OperatorFrameSummary,
+    ) -> ValidatorDutiesOutcome:
+        if duties.attestation is None or duties.attestation.assigned == 0:
+            # It's possible that the validator is not assigned to any duty, hence it's performance
+            # is not presented in the aggregates (e.g. exited, pending for activation etc).
+            return ValidatorDutiesOutcome(participation_share=0, rebate_share=0, strikes=0)
+
+        log_validator = log_operator.validators[validator.index]
+
+        if validator.validator.slashed is True:
```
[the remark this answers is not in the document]
vgorkavenko: didn't get the point

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+                continue
+            rewards_distribution[no_id] = rewards_to_distribute * no_participation_share // total_shares
+
+        return rewards_distribution
+
+    @staticmethod
+    def validate_distribution(total_distributed_rewards, total_rebate, total_rewards_to_distribute):
+        if (total_distributed_rewards + total_rebate) > total_rewards_to_distribute:
+            raise ValueError(
+                f"Invalid distribution: {total_distributed_rewards + total_rebate} > {total_rewards_to_distribute}"
```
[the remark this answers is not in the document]
vgorkavenko: didn't get the point

src/modules/submodules/oracle_module.py
```diff
@@ -112,6 +113,12 @@ def _cycle(self):
         except ValueError as error:
             logger.error({'msg': 'Unexpected error.', 'error': str(error)})
 
+    @staticmethod
+    def _reset_cycle_timeout():
```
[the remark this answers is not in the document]
vgorkavenko: no, because it may hide some problems not related to stuck thread

src/modules/csm/csm.py
```diff
@@ -218,114 +207,15 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
         processor = FrameCheckpointProcessor(self.w3.cc, self.state, converter, blockstamp)
 
         for checkpoint in checkpoints:
-            if self.current_frame_range(self._receive_last_finalized_slot()) != (l_epoch, r_epoch):
-                logger.info({"msg": "Checkpoints were prepared for an outdated frame, stop processing"})
+            if self.get_epochs_range_to_process(self._receive_last_finalized_slot()) != (l_epoch, r_epoch):
```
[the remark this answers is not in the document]
vgorkavenko: This helps us avoid processing outdated checkpoints. For example, if checkpoint `N` takes too long to process, the frame is shifted and there is no point in processing checkpoint `N + 1`

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+                logger.info({"msg": f"No active validators for {no_id=} in the frame. Skipping"})
+                continue
+
+            logger.info({"msg": f"Calculating distribution for {no_id=}"})
+            log_operator = log.operators[no_id]
+
+            curve_params = self.w3.csm.get_curve_params(no_id, blockstamp)
+            log_operator.performance_coefficients = curve_params.perf_coeffs
+
+            sorted_active_validators = sorted(active_validators, key=lambda v: v.index)
```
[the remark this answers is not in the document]
vgorkavenko: done

chasingrainbows (approved):

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+    def _calculate_distribution_in_frame(
+        self,
+        frame: Frame,
+        blockstamp: ReferenceBlockStamp,
+        rewards_to_distribute: Shares,
+        operators_to_validators: ValidatorsByNodeOperator,
+        log: FramePerfLog,
+    ) -> tuple[dict[NodeOperatorId, Shares], Shares, Shares, dict[StrikesValidator, int]]:
+        total_rebate_share = 0
+        participation_shares: defaultdict[NodeOperatorId, Shares] = defaultdict(int)
```
vgorkavenko, line 126: It is not `Shares` it is something different. Mb `ActivityShares`

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+    def _get_module_validators(self, blockstamp: ReferenceBlockStamp) -> ValidatorsByNodeOperator:
+        return self.w3.lido_validators.get_module_validators_by_node_operators(
+            StakingModuleAddress(self.w3.csm.module.address), blockstamp
+        )
+
+    def _calculate_distribution_in_frame(
+        self,
+        frame: Frame,
+        blockstamp: ReferenceBlockStamp,
+        rewards_to_distribute: Shares,
```
vgorkavenko, line 121: Mb `stETHShares`

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+        participation_shares: defaultdict[NodeOperatorId, Shares] = defaultdict(int)
+        frame_strikes: dict[StrikesValidator, int] = {}
+
+        network_perf = self._get_network_performance(frame)
+
+        for (_, no_id), validators in operators_to_validators.items():
+            active_validators = [v for v in validators if self.state.data[frame].attestations[v.index].assigned > 0]
+            if not active_validators:
+                logger.info({"msg": f"No active validators for {no_id=} in the frame. Skipping"})
+                continue
```
vgorkavenko, line 135: need to test full skipped frame calculation due to no active validators

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+    @staticmethod
+    def get_validator_duties_outcome(
+        validator: LidoValidator,
+        duties: ValidatorDuties,
+        threshold: float,
+        reward_share: float,
+        perf_coeffs: PerformanceCoefficients,
+        log_operator: OperatorFrameSummary,
+    ) -> ValidatorDutiesOutcome:
+        if duties.attestation is None or duties.attestation.assigned == 0:
```
vgorkavenko, line 197: add some note about the case when you have no assigned attestations, but you're in sync committee

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+            #
+            #    Example:
+            #     - Validator was 103 epochs active in the frame (assigned 103 attestations)
+            #     - Reward share for this Operator's key is 0.85
+            #    87.55 ≈ 88 of 103 participation shares should be counted for the operator key's reward.
+            #    The rest 15 participation shares should be counted for the protocol's rebate.
+            #
+            participation_share = math.ceil(duties.attestation.assigned * reward_share)
+            rebate_share = duties.attestation.assigned - participation_share
+            assert rebate_share >= 0, f"Invalid rebate share: {rebate_share=}"
```
vgorkavenko, line 236: `raise ValueError`

src/modules/csm/distribution.py
```diff
@@ -0,0 +1,293 @@
...
+    @staticmethod
+    def calc_rewards_distribution_in_frame(
+        participation_shares: dict[NodeOperatorId, int],
+        rebate_share: int,
+        rewards_to_distribute: int,
+    ) -> dict[NodeOperatorId, Shares]:
+        if rewards_to_distribute < 0:
+            raise ValueError(f"Invalid rewards to distribute: {rewards_to_distribute=}")
+
+        rewards_distribution: dict[NodeOperatorId, int] = defaultdict(int)
```
vgorkavenko, line 252:
```suggestion
        rewards_distribution: dict[NodeOperatorId, Shares] = defaultdict(Shares)
```

tests/fork/test_csm_oracle_cycle.py
vgorkavenko: Yep. Can be implemented in the future (but probably, will be removed)

chasingrainbows (approved):

PR #634, opened 2025-02-19
CSM v2 strikes

madlabman:
The specification: https://hackmd.io/@lido/Sy0nRd36a

P.S. Bumped mypy for python 3.12 generics support, see [blog post](https://mypy-lang.blogspot.com/2024/10/mypy-112-released.html).

### DIFF
--- a/src/modules/csm/csm.py
+++ b/src/modules/csm/csm.py
@@ -549,6 +549,12 @@ class LastReport:
 
     @cached_property
     def rewards(self) -> Iterable[RewardsTreeLeaf]:
+        if (self.rewards_tree_cid is None) != (self.rewards_tree_root == ZERO_HASH):
+            raise InconsistentData(
+                "Got inconsistent previous rewards tree data: "
+                f"tree_root={self.rewards_tree_root.hex()} tree_cid={self.rewards_tree_cid=}"
+            )
+
         if self.rewards_tree_cid is None or self.rewards_tree_root == ZERO_HASH:
             logger.info({"msg": f"No rewards distribution as of {self.blockstamp=}."})
             return []
@@ -565,6 +571,12 @@ class LastReport:
 
     @cached_property
     def strikes(self) -> dict[StrikesValidator, StrikesList]:
+        if (self.strikes_tree_cid is None) != (self.strikes_tree_root == ZERO_HASH):
+            raise InconsistentData(
+                "Got inconsistent previous strikes tree data: "
+                f"tree_root={self.strikes_tree_root.hex()} tree_cid={self.strikes_tree_cid=}"
+            )
+
         if self.strikes_tree_cid is None or self.strikes_tree_root == ZERO_HASH:
             logger.info({"msg": f"No strikes reported as of {self.blockstamp=}."})
             return {}


src/modules/csm/types.py
```diff
@@ -10,8 +10,43 @@
 logger = logging.getLogger(__name__)
 
 
+class StrikesList(Sequence[int]):
```
vgorkavenko, line 13: Can it be `class StrikesList(list)` ? To get rid of `StrikesTreeJSONEncoder`

src/modules/csm/types.py
```diff
@@ -10,8 +10,43 @@
...
+    """Deque-like structure to store strikes"""
+
+    sentinel: int
```
vgorkavenko, line 16: Can we just use hardcoded `0` in `resize` ? To simplify the thing

src/modules/csm/csm.py
```diff
@@ -94,47 +97,73 @@ def execute_module(self, last_finalized_blockstamp: BlockStamp) -> ModuleExecute
...
-            raise InconsistentData(f"Invalid distribution: {sum(total_rewards.values())=} != {total_distributed=}")
+        if (prev_strikes_cid is None) != (prev_strikes_root == ZERO_HASH):
+            raise InconsistentData(f"Got inconsistent previous tree data: {prev_strikes_root=} {prev_strikes_cid=}")
 
+        total_distributed, total_rewards, raw_strikes, logs = self.calculate_distribution(blockstamp)
         log_cid = self.publish_log(logs)
 
-        if not total_distributed and not total_rewards:
-            logger.info({"msg": "No rewards distributed in the current frame"})
+        if not total_distributed and not raw_strikes:
```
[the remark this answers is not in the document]
vgorkavenko: From current perspective, fully empty report is less possible case. We will take a look at this place again due to upcoming `rebate` feature from another PR

src/modules/csm/csm.py
```diff
@@ -245,10 +281,18 @@ def calculate_distribution(
...
+
+            rewards_in_frame, strikes = self._calculate_distribution_in_frame(
+                frame,
+                frame_threshold,
+                rewards_to_distribute_in_frame,
+                operators_to_validators,
+                log,
             )
             distributed_in_frame = sum(rewards_in_frame.values())
+            logger.info({"msg": f"No rewards distributed in frame [{from_epoch};{to_epoch}]"})
```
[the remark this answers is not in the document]
vgorkavenko: It doesn't make sense if we have only one frame in report if I got it right

src/modules/csm/csm.py
```diff
@@ -275,24 +322,21 @@ def _get_ref_blockstamp_for_frame(
...
             log_operator = log.operators[no_id]
-            if no_id in stuck_operators:
-                log_operator.stuck = True
-                continue
             for validator in validators:
                 duty = self.state.data[frame].get(validator.index)
-                self.process_validator_duty(validator, duty, threshold, participation_shares, log_operator)
+                strike = self.process_validator_duty(validator, duty, threshold, participation_shares, log_operator)
+                pubkey = HexBytes(hex_str_to_bytes(validator.validator.pubkey))
+                strikes[(no_id, pubkey)].insert(0, strike)
```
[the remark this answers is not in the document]
vgorkavenko: If you are going to make it this way, you will have map of all active `no_id, pubkey` in memory where most of them will be with `0`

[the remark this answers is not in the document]
madlabman: Welp, I've implemented strikes per frame approach.

src/modules/csm/csm.py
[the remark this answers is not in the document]
madlabman: Fixed.

src/modules/csm/csm.py
```diff
@@ -368,32 +412,37 @@ def get_accumulated_rewards(self, cid: CID, root: HexBytes) -> Iterator[tuple[No
...
+        strikes: dict[StrikesValidator, list[int]],
+        blockstamp: BlockStamp,
+    ) -> dict[StrikesValidator, StrikesList]:
+        out: dict[StrikesValidator, StrikesList] = {}
+
+        for (no_id, pubkey), raw_strikes in strikes.items():
+            # NOTE: Cleanup sequences like [0,0,0] since they don't bring any information.
+            if not sum(raw_strikes):
+                continue
+            maxlen = self.w3.csm.get_strikes_params(no_id, blockstamp).lifetime
```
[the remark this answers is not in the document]
madlabman: Yeah, shrinking and a consequent restoring of a size may have interesting side effect. Implemented.

src/modules/csm/types.py
madlabman: I end up realizing I want to keep the value, but as a class-level constant to avoid using magic numbers.

src/modules/csm/csm.py
```diff
@@ -422,24 +421,37 @@ def get_accumulated_strikes(self, cid: CID, root: HexBytes) -> Iterator[StrikesT
...
-            maxlen = self.w3.csm.get_strikes_params(no_id, blockstamp).lifetime
-            if (no_id, pubkey) in out:
-                raise CSMError(f"Double accounting of strikes for {no_id=}")
-            out[(no_id, pubkey)] = StrikesList(raw_strikes, maxlen=maxlen)
+        for frame in self.state.frames:
+            strikes_in_frame = strikes_per_frame[frame]
+            for key in strikes_in_frame:
+                if key not in out:
+                    out[key] = StrikesList()
+                out[key].push(strikes_in_frame[key])
```
vgorkavenko, line 439: `out: defaultdict` ?

madlabman: It can be, but I decided to keep this branch explicit.

src/modules/csm/csm.py
```diff
@@ -93,47 +98,38 @@ def execute_module(self, last_finalized_blockstamp: BlockStamp) -> ModuleExecute
...
 
-        total_distributed, total_rewards, logs = self.calculate_distribution(blockstamp)
+        if total_distributed:
+            rewards_tree = self.make_rewards_tree(total_rewards)
+            rewards_tree_root = rewards_tree.root
+            rewards_cid = self.publish_tree(rewards_tree)
 
-        if total_distributed != sum(total_rewards.values()):
-            raise InconsistentData(f"Invalid distribution: {sum(total_rewards.values())=} != {total_distributed=}")
+        if strikes:
```
vgorkavenko, line 117: If no current strikes, but there were strikes before (due to shifting), we need to pass zero hash to wipe it in the contract to not have the tree consists of leafs with `[0,0,0,0,0]` because it doesn't make sense.

If we pass the same root as previous it can be interpreted as a result of non-shifted strikes. I'd like to avoid it.

src/modules/csm/csm.py
```diff
@@ -224,13 +220,19 @@ def collect_data(self, blockstamp: BlockStamp) -> bool:
...
+        dict[StrikesValidator, StrikesList],
+        list[FramePerfLog],
+    ]:
         """Computes distribution of fee shares at the given timestamp"""
         operators_to_validators = self.module_validators_by_node_operators(blockstamp)
 
-        total_distributed = Shares(0)
         total_rewards = defaultdict[NodeOperatorId, Shares](Shares)
+        total_distributed = Shares(0)
+        strikes = last_report.strikes
```
vgorkavenko, line 235: Does it mutate `last_report.strikes` as well? Can be a bit misleading during debug process. Let's copy it

src/modules/csm/csm.py
```diff
@@ -258,7 +273,16 @@ def calculate_distribution(
 
             logs.append(log)
 
-        return total_distributed, total_rewards, logs
+        if total_distributed != sum(total_rewards.values()):
+            raise InconsistentData(f"Invalid distribution: {sum(total_rewards.values())=} != {total_distributed=}")
+
+        for no_id, rewards in last_report.rewards:
+            total_rewards[no_id] += rewards
```
vgorkavenko, line 280:
```suggestion
        for no_id, last_report_rewards in last_report.rewards:
            total_rewards[no_id] += last_report_rewards
```
Confused me a bit because of the same cycle as above.

Can we consider to have some func to merge the values? With validation. And reuse it here and in `frames` cycle

src/modules/csm/csm.py
```diff
@@ -447,3 +508,69 @@ def _get_staking_module(self) -> StakingModule:
...
+            blockstamp,
+            rewards_tree_root,
+            strikes_tree_root,
+            rewards_tree_cid,
+            strikes_tree_cid,
+        )
+
+    @cached_property
+    def rewards(self) -> Iterable[RewardsTreeLeaf]:
+        if self.rewards_tree_cid is None or self.rewards_tree_root == ZERO_HASH:
```
vgorkavenko, line 548:
```suggestion
        if self.rewards_tree_cid is None and self.rewards_tree_root == ZERO_HASH:
```

src/modules/csm/csm.py
```diff
@@ -447,3 +508,69 @@ def _get_staking_module(self) -> StakingModule:
...
+        logger.info({"msg": "Restored rewards tree from IPFS dump", "root": repr(tree.root)})
+
+        if tree.root != self.rewards_tree_root:
+            raise ValueError("Unexpected rewards tree root got from IPFS dump")
+
+        return tree.values
+
+    @cached_property
+    def strikes(self) -> dict[StrikesValidator, StrikesList]:
+        if self.strikes_tree_cid is None or self.strikes_tree_root == ZERO_HASH:
```
vgorkavenko, line 564:
```suggestion
        if self.strikes_tree_cid is None and self.strikes_tree_root == ZERO_HASH:
```

src/providers/execution/contracts/cs_accounting.py
```diff
@@ -26,6 +28,7 @@ def fee_distributor(self, block_identifier: BlockIdentifier = "latest") -> Check
         )
         return Web3.to_checksum_address(resp)
 
+    @lru_cache
```
vgorkavenko, line 31: Not sure that it make sense here. We need to cache CSParams values by curve_id only

src/web3py/extensions/csm.py
```diff
@@ -40,15 +43,29 @@ def get_csm_last_processing_ref_slot(self, blockstamp: BlockStamp) -> SlotNumber
...
+    def get_strikes_tree_root(self, blockstamp: BlockStamp) -> HexBytes:
+        return self.strikes.tree_root(blockstamp.block_hash)
+
+    def get_strikes_tree_cid(self, blockstamp: BlockStamp) -> CID | None:
+        result = self.strikes.tree_cid(blockstamp.block_hash)
+        if result == "":
+            return None
+        return CIDv0(result) if is_cid_v0(result) else CIDv1(result)
+
+    @lru_cache
```
vgorkavenko, line 64: doesn't make much sense here for the same reason as above

src/providers/execution/contracts/cs_accounting.py
madlabman: We have only a node operator ID at hand, so we go: `no_id` -> `curve_id` -> `params`.
An operator has multiple validators, hence it will be queried multiple times, so we cache `no_id` -> `curve_id` part. Then we cache the `curve_id` -> `params` in `get_strikes_params` at the contract level. And, in csm extention `get_strikes_params` we have a redundant cache.

src/modules/csm/csm.py
madlabman: Doesn't make sense for me. Added a check that these two values are empty simultaneously instead.

src/modules/csm/csm.py
madlabman: The same as above.

vgorkavenko (approved):
LGTM!

PR #637, opened 2025-02-24
Removing filtration by earliest exit epoch in a prediction of earliest slashed epoch in bunker mode

hweawer:
# What

In the bunker mode we doesn't want to filter by earliest exit epoch.

### DIFF
--- a/src/services/safe_border.py
+++ b/src/services/safe_border.py
@@ -178,12 +178,12 @@ class SafeBorder(Web3Converter):
         last_finalized_request_id_epoch = self.get_epoch_by_slot(self._get_last_finalized_withdrawal_request_slot())
         # Since we are looking for the safe border epoch, we can start from the last finalized withdrawal request epoch
         # or the earliest activation epoch among the given validators for optimization
-        earliest_activation_epoch = self._get_validators_earliest_activation_epoch(validators)
+        earliest_activation_epoch = min((v.validator.activation_epoch for v in validators), default=0)
         start_epoch = max(last_finalized_request_id_epoch, earliest_activation_epoch)
 
         # We can stop searching for the slashed epoch when we reach the reference epoch
         # or the max possible earliest slashed epoch for the given validators
-        withdrawable_epoch = min(get_validators_withdrawable_epochs(validators))
+        withdrawable_epoch = min(v.validator.withdrawable_epoch for v in validators)
         max_possible_earliest_slashed_epoch = EpochNumber(withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR)
         end_epoch = min(self.blockstamp.ref_epoch, max_possible_earliest_slashed_epoch)
 
@@ -205,16 +205,6 @@ class SafeBorder(Web3Converter):
         slot_number = self.get_frame_first_slot(start_frame)
         epoch_number = self.get_epoch_by_slot(slot_number)
         return epoch_number
-
-    def _get_validators_earliest_activation_epoch(self, validators: list[Validator]) -> EpochNumber:
-        if len(validators) == 0:
-            return EpochNumber(0)
-
-        sorted_validators = sorted(
-            validators,
-            key=lambda validator: validator.validator.activation_epoch
-        )
-        return sorted_validators[0].validator.activation_epoch
 
     def _slashings_in_frame(self, frame: FrameNumber, slashed_pubkeys: set[str]) -> bool:
         """
@@ -276,7 +266,3 @@ def filter_validators_by_exit_epoch(validators: Iterable[Validator], exit_epoch:
 
 def get_validators_pubkeys(validators: Iterable[Validator]) -> list[HexStr]:
     return [HexStr(v.validator.pubkey) for v in validators]
-
-
-def get_validators_withdrawable_epochs(validators: Iterable[Validator]) -> list[EpochNumber]:
-    return [v.validator.withdrawable_epoch for v in validators]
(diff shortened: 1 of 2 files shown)


src/services/safe_border.py
```diff
@@ -95,7 +96,8 @@ def _get_negative_rebase_border_epoch(self) -> EpochNumber:
 
         latest_allowable_epoch = bunker_start_or_last_successful_report_epoch - self.finalization_default_shift
 
-        max_negative_rebase = self.w3.lido_contracts.oracle_daemon_config.finalization_max_negative_rebase_epoch_shift(self.blockstamp.block_hash)
+        max_negative_rebase = self.w3.lido_contracts.oracle_daemon_config.finalization_max_negative_rebase_epoch_shift(
+            self.blockstamp.block_hash)
```
F4ever, line 100:
```suggestion
            self.blockstamp.block_hash,
        )
```

src/services/safe_border.py
```diff
@@ -135,25 +138,19 @@ def _get_earliest_slashed_epoch_among_incomplete_slashings(self) -> EpochNumber
...
+        for validator in validators_slashed_non_withdrawable:
             predicted_epoch = self._predict_earliest_slashed_epoch(validator)
 
             if not predicted_epoch:
-                return self._find_earliest_slashed_epoch_rounded_to_frame(validators_with_earliest_exit_epoch)
+                return self._find_earliest_slashed_epoch_rounded_to_frame(validators_slashed_non_withdrawable)
 
-            if not earliest_predicted_epoch or earliest_predicted_epoch > predicted_epoch:
-                earliest_predicted_epoch = predicted_epoch
+            earliest_predicted_epoch = min(earliest_predicted_epoch or EpochNumber(sys.maxsize), predicted_epoch)
```
F4ever, line 153: I don't think we should use sys.maxsize here.
Pls, check what type is epoch number and setup it's value.
Or I'd better prefer old expression

src/services/safe_border.py
```diff
@@ -179,22 +176,20 @@ def _find_earliest_slashed_epoch_rounded_to_frame(self, validators: list[Validat
...
-        withdrawable_epoch = min(get_validators_withdrawable_epochs(validators))
         last_finalized_request_id_epoch = self.get_epoch_by_slot(self._get_last_finalized_withdrawal_request_slot())
-
-        earliest_activation_slot = self._get_validators_earliest_activation_epoch(validators)
-        max_possible_earliest_slashed_epoch = EpochNumber(withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR)
-
         # Since we are looking for the safe border epoch, we can start from the last finalized withdrawal request epoch
         # or the earliest activation epoch among the given validators for optimization
-        start_epoch = max(last_finalized_request_id_epoch, earliest_activation_slot)
+        earliest_activation_epoch = min((v.validator.activation_epoch for v in validators), default=0)
```
F4ever, line 182: move this two lines up

src/services/safe_border.py
```diff
@@ -179,22 +176,20 @@ def _find_earliest_slashed_epoch_rounded_to_frame(self, validators: list[Validat
...
+        start_epoch = max(last_finalized_request_id_epoch, earliest_activation_epoch)
 
         # We can stop searching for the slashed epoch when we reach the reference epoch
         # or the max possible earliest slashed epoch for the given validators
+        withdrawable_epoch = min(v.validator.withdrawable_epoch for v in validators)
```
F4ever, line 187: Let's check that validators is not empty list, or add default here

F4ever: I'd prefer to raise the exception at the beginning of the function in case of empty list

tests/modules/accounting/test_safe_border_unit.py
```diff
@@ -222,6 +213,58 @@ def test_get_earliest_slashed_epoch_among_incomplete_slashings_at_least_one_unpr
     assert safe_border._get_earliest_slashed_epoch_among_incomplete_slashings() == 1331
 
 
+###
+# validator 1:
```
F4ever, line 217: Add some text details

tests/modules/accounting/test_safe_border_unit.py
```diff
@@ -222,6 +213,58 @@ def test_get_earliest_slashed_epoch_among_incomplete_slashings_at_least_one_unpr
...
+# ------|------------------|---------------|-->
+#       slashed            exit            withdrawable
+#       epoch              epoch           epoch
+#
+#     ^
+#     expected
+#     safe
+#     border
+###
+def test_get_earliest_slashed_epoch_slashing_after_exit(safe_border, past_blockstamp, lido_validators):
```
F4ever, line 233: maybe
`test_get_earliest_slashed_epcoh_if_exiting_validator_slashed`

src/services/safe_border.py
```diff
@@ -179,22 +176,20 @@ def _find_earliest_slashed_epoch_rounded_to_frame(self, validators: list[Validat
         """
         Returns the earliest slashed epoch for the given validators rounded to the frame
         """
-        withdrawable_epoch = min(get_validators_withdrawable_epochs(validators))
         last_finalized_request_id_epoch = self.get_epoch_by_slot(self._get_last_finalized_withdrawal_request_slot())
-
-        earliest_activation_slot = self._get_validators_earliest_activation_epoch(validators)
-        max_possible_earliest_slashed_epoch = EpochNumber(withdrawable_epoch - EPOCHS_PER_SLASHINGS_VECTOR)
-
+        earliest_activation_epoch = min((v.validator.activation_epoch for v in validators), default=0)
```
F4ever, line 180: we can remove default here

F4ever (approved):

PR #639, opened 2025-02-26
Calculate midterm penalty based on slashings vector in state

hweawer:
# What

Change calculation of predicted midterm penalty.

# Why

At the moment we are calculating predicted midterm penalty based on an effective_balance of the validator. This calculations based on a `ref_slot` `effective_balance`. What we want to do instead is to calculate it based on `slashings` from the state as per [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-process_slashings). This will provide more accurate midterm slashing predictions.

# How

We are predicting the penalty in future, so we want to cut the epochs which will be replaced, by the time chain reach midterm slashing epoch. This is what `_cut_slashings` does


src/services/bunker_cases/midterm_slashing_penalty.py
```diff
@@ -232,20 +236,22 @@ def predict_midterm_penalty_in_frame_post_electra(
...
                 report_ref_epoch, all_slashed_validators, midterm_penalty_epoch
             )
 
             if is_electra_activated(frame_ref_epoch):
                 penalty_in_frame += MidtermSlashingPenalty.get_validator_midterm_penalty_electra(
-                    validator, bound_slashed_validators, total_balance
+                    validator, slashings, total_balance, midterm_penalty_epoch, report_ref_epoch
                 )
             else:
                 penalty_in_frame += MidtermSlashingPenalty.get_validator_midterm_penalty(
```
F4ever, line 257: I think we should update implementation here as well

src/services/bunker_cases/midterm_slashing_penalty.py
```diff
@@ -275,25 +281,42 @@ def get_validator_midterm_penalty(
...
+    @staticmethod
+    def _cut_slashings(
+        slashings: list[Gwei], midterm_penalty_epoch: EpochNumber, report_ref_epoch: EpochNumber
+    ) -> list[Gwei]:
+        """
+        Filters out slashing values based on epochs within a midterm penalty epoch. Slashings is a ring buffer on epochs.
+        @see https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-slash_validator
+        We want to filter out epochs in the past which will not be relevant at the time of midterm penalty epoch.
+        """
+        obsolete_indexes = {
```
F4ever, line 314: Maybe let's add sanity check that 
len(slashings) == EPOCHS_PER_SLASHINGS_VECTOR

src/services/bunker_cases/midterm_slashing_penalty.py
```diff
@@ -275,25 +281,42 @@ def get_validator_midterm_penalty(
...
-        penalty_per_effective_balance_increment = adjusted_total_slashing_balance // (
-            total_balance // EFFECTIVE_BALANCE_INCREMENT
+        slashings_sum = sum(MidtermSlashingPenalty._cut_slashings(slashings, midterm_penalty_epoch, report_ref_epoch))
+        adjusted_total_slashing_balance = min(
+            slashings_sum * PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX,
+            total_balance,
         )
-        effective_balance_increments = effective_balance // EFFECTIVE_BALANCE_INCREMENT
+        effective_balance = validator.validator.effective_balance
+        increment = EFFECTIVE_BALANCE_INCREMENT
```
F4ever, line 299: Optional:
maybe let's leave the previous version

F4ever (approved):

PR #662, opened 2025-03-19
feat(ORC-289): add kapi integration/unit tests

chasingrainbows:


src/main.py
```diff
@@ -75,7 +75,7 @@ def main(module_name: OracleModule):
     )
 
     logger.info({'msg': 'Check configured providers.'})
-    if Version(kac.get_status().appVersion) < Version('1.5.0'):
+    if Version(kac.get_status().appVersion) < variables.ALLOWED_KAPI_VERSION:
```
F4ever, line 78: I propose to move this to constant.

src/utils/keys.py
```diff
@@ -0,0 +1,35 @@
+"""
```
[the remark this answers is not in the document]
F4ever: If we store code in src it requires audits and quality review and testing.
I agree with @hweawer to move this to some kind of test utils.

tests/providers_clients/test_keys_api_client.py
```diff
@@ -1,52 +1,280 @@
...
+        assert keys1 == keys2
+        assert len(responses.calls) == 1
+
+    @responses.activate
+    def test_get_module_operators_keys__empty_response__empty_lists(
+        self,
+        keys_api_client,
+        empty_blockstamp,
+    ):
+        csm_module_address = cast(StakingModuleAddress, '0xdA7dE2ECdDfccC6c3AF10108Db212ACBBf9EA83F')
```
F4ever, line 193: Let's move to variable.
Mark this var as mainnet only.
Could be added sanity check for this test that chain_id == 1.
Mark all tests that depends on mainnet with new tag, eg. `pytest.mark.mainnet`

tests/providers_clients/test_keys_api_client.py
```diff
@@ -1,52 +1,280 @@
...
 
 @pytest.mark.unit
-def test_get_with_blockstamp_retries_exhausted(keys_api_client, monkeypatch):
-    variables.HTTP_REQUEST_SLEEP_BEFORE_RETRY_IN_SECONDS = 1
-    keys_api_client._get = Mock(
-        return_value=(
-            None,
-            {"meta": {"elBlockSnapshot": {"blockNumber": empty_blockstamp.block_number - 1}}},
+class TestUnitKeysAPIClient:
+    KEYS_API_MOCK_URL = 'http://localhost:8000/'
```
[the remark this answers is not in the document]
chasingrainbows: nope, all http requests during TestUnitKeysAPIClient is mocked, basically there is any url can be, just wrote localhost as something default

tests/providers_clients/test_keys_api_client.py
chasingrainbows: > Let's move to variable.
You mean variables.py?

> Mark this var as mainnet only.
What do you mean? By naming?

src/main.py
```diff
@@ -75,7 +75,7 @@ def main(module_name: OracleModule):
...
         raise IncompatibleException('Incompatible KAPI version. Required >= 1.5.0.')
```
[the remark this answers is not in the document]
chasingrainbows: done

src/variables.py
```diff
@@ -89,6 +90,8 @@
 
 CACHE_PATH: Final = Path(os.getenv("CACHE_PATH", "."))
 
+ALLOWED_KAPI_VERSION = Version('1.5.0')
```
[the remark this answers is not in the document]
chasingrainbows: done

src/utils/keys.py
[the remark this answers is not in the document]
chasingrainbows: done

tests/providers_clients/test_keys_api_client.py
chasingrainbows: > Let's move to variable.

The CSM_MODULE_ADDRESS variable already exists, but it is can be empty in the secrets, causing the test in CI to fail. I’ve decided to create a constant MAINNET_CSM_MODULE_ADDRESS. Please check it out. We could use CSM_MODULE_ADDRESS, but if it’s empty locally, one of the tests will fail. I believe all tests should work well locally by default in a newly cloned repository. If u disagree, we can go with CSM_MODULE_ADDRESS.

chasingrainbows: > Mark this var as mainnet only

done

> Could be added sanity check for this test that chain_id == 1.

This endpoint does not returns chain_id, but this is tested in test_get_status__response_version_is_allowed and test_check_providers_consistency__mainnet

chasingrainbows: > Let's move to variable.

Reused CSM_MODULE_ADDRESS from variables.py

tests/providers_clients/test_keys_api_client.py
```diff
@@ -1,52 +1,297 @@
...
+            assert lido_key.used is True
+            self._assert_lido_key(lido_key)
+
+    def test_get_module_operators_keys__csm_module__response_data_is_valid(
+        self,
+        keys_api_client,
+        empty_blockstamp,
+    ):
+        csm_module_operators_keys = keys_api_client.get_module_operators_keys(
+            module_address=variables.CSM_MODULE_ADDRESS, blockstamp=empty_blockstamp
```
F4ever, line 82: Let's move this variable to fixtures. 
eg was done for `LIDO_LOCATOR_ADDRESS`

chasingrainbows: Done

F4ever (approved):

PR #663, opened 2025-03-19
[PoC] Very ugly fixes to achieve reproducible builds of Docker image

n0guest:
## Changes

- clearing log files generated while an image is being build
- clearing LD auxilary cache
- disabled generation pyc for pip
- deleting generated pyc after poetry execution
- deleting RECORD files from python packages (content is randomized for some reason?)
- introduced CFLAGS to make repeatable build of shared library from lru-dict (sub-dep of web3)

## Docker tricks

Just an example of how to build:

```
SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct) docker buildx build --no-cache --output type=docker,name=lidofinance/lido-oracle,push=false,rewrite-timestamp=true .
```
Variable SOURCE_DATE_EPOCH uses a timestamp from the latest git commit, which is later used as reference point later to rewrite timestamps inside of image layers.

### BuildKit

Also you need BuildKit >= 0.13, because versions up to 0.12 are buggy. I've used this:

```
$ docker buildx inspect | grep -i buildkit
Name:          buildkit-container
Name:                  buildkit-container0
BuildKit daemon flags: --allow-insecure-entitlement=network.host
BuildKit version:      v0.20.1
 org.mobyproject.buildkit.worker.executor:         oci
 org.mobyproject.buildkit.worker.hostname:         e299ecd707a9
 org.mobyproject.buildkit.worker.network:          host
 org.mobyproject.buildkit.worker.oci.process-mode: sandbox
 org.mobyproject.buildkit.worker.selinux.enabled:  false
 org.mobyproject.buildkit.worker.snapshotter:      overlayfs
```

PS: Don't use this PR, it's very ugly and just PoC ¯\_(ツ)_/¯

## Related docs
* https://reproducible-builds.org/docs/source-date-epoch/
* https://docs.docker.com/build/ci/github-actions/reproducible-builds/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,16 @@
 FROM python:3.12.4-slim as base
 
+
 RUN apt-get update && apt-get install -y --no-install-recommends -qq \
     libffi-dev=3.4.4-1 \
     g++=4:12.2.0-3 \
     curl=7.88.1-10+deb12u12 \
+ && find /var/log -type f -name '*.log' -print -exec truncate -s0 '{}' \; \
+ && find /var/cache/ldconfig -type f -delete \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
+
+FROM base as builder
 
 ENV PYTHONUNBUFFERED=1 \
     PYTHONDONTWRITEBYTECODE=1 \
@@ -16,19 +21,28 @@ ENV PYTHONUNBUFFERED=1 \
     POETRY_NO_INTERACTION=1 \
     VENV_PATH="/.venv"
 
+# ugly hack for C extension from lru-dict
+ENV CFLAGS="-g0 -O2 -ffile-prefix-map=/src=."
+
 ENV PATH="$VENV_PATH/bin:$PATH"
 
-FROM base as builder
-
 ENV POETRY_VERSION=1.3.2
-RUN pip install --no-cache-dir poetry==$POETRY_VERSION
+
+RUN pip install --no-compile --no-cache-dir poetry==$POETRY_VERSION
 
 WORKDIR /
 COPY pyproject.toml poetry.lock ./
-RUN poetry install --only main --no-root
+RUN poetry install --only main --no-root --no-cache
 
+RUN rm -rf /root/.cache $VENV_PATH/src \
+  && find $VENV_PATH -type f -name 'RECORD' -delete \
+  && find $VENV_PATH -type f -name '*.pyc' -delete \
+  && find /root/.local -type f -name '*.pyc' -delete \
+  && find /usr/local -type f -name '*.pyc' -delete
 
 FROM base as production
+
+ENV VENV_PATH="/.venv"
 
 COPY --from=builder $VENV_PATH $VENV_PATH
 WORKDIR /app
@@ -36,8 +50,8 @@ COPY . .
 
 RUN apt-get clean && find /var/lib/apt/lists/ -type f -delete && chown -R www-data /app/
 
-ENV PROMETHEUS_PORT 9000
-ENV HEALTHCHECK_SERVER_PORT 9010
+ENV PROMETHEUS_PORT=9000
+ENV HEALTHCHECK_SERVER_PORT=9010
 
 EXPOSE $PROMETHEUS_PORT
 USER www-data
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,12 +1,17 @@
 FROM python:3.12.4-slim as base
 
+
 RUN apt-get update && apt-get install -y --no-install-recommends -qq \
     libffi-dev=3.4.4-1 \
     g++=4:12.2.0-3 \
     curl=7.88.1-10+deb12u12 \
+ && find /var/log -type f -name '*.log' -print -exec truncate -s0 '{}' \; \
+ && find /var/cache/ldconfig -type f -delete \
```
amarao, line 9: Oh, that's new.  I didn't knew about this one.

Dockerfile
```diff
@@ -1,12 +1,17 @@
 FROM python:3.12.4-slim as base
 
+
 RUN apt-get update && apt-get install -y --no-install-recommends -qq \
     libffi-dev=3.4.4-1 \
```
amarao, line 5: As we discussed at the sync, the main problem with apt is that build become reproducible, but not for long time, because Debian and Ubuntu are removing older package versions from the Archive.

As far as I understand, we should either use snapshots (a rather unreliable service, outside of the scope of Debian promises), or we can use 'base' image, which is build once and reused in 'FROM' form.

In this case build process consist of two:

(in some other location):

```
FROM python:3.12.4-slim
RUN apt-get update && apt-get install ... && rm ..
```
(build it as oracle-base:3.12.4)

And the main image for application:

```
FROM oracle-base:3.12.4 as base
...
```

The idea is that 'base' image build once (per debian dependencies) and then never updated (until need to update dependencies). It has fixed versions, so it's reproducible for some time, then it become unreproducible, but at that moment, it's a good idea to update it, because package version had changed, in 90% cases due to security fixes.

Dockerfile
```diff
@@ -16,28 +21,37 @@ ENV PYTHONUNBUFFERED=1 \
...
+# ugly hack for C extension from lru-dict
+ENV CFLAGS="-g0 -O2 -ffile-prefix-map=/src=."
 
-FROM base as builder
+ENV PATH="$VENV_PATH/bin:$PATH"
 
 ENV POETRY_VERSION=1.3.2
-RUN pip install --no-cache-dir poetry==$POETRY_VERSION
+
+RUN pip install --no-compile --no-cache-dir poetry==$POETRY_VERSION
```
amarao, line 31: I usually just remove poetry cache in the same stage (the same as with apt):

`RUN poetry install && rm ...`

Also, may be, you can get away with poetry from Debian:

`python3-poetry` package. There is some incompatibility between 1.8 and 2.0, I don't know what is used as base for python-3.12. Also, poetry (even old one from Debian packages) can be installed with `poetry self update`.

Dockerfile
```diff
@@ -16,28 +21,37 @@ ENV PYTHONUNBUFFERED=1 \
...
 
 WORKDIR /
 COPY pyproject.toml poetry.lock ./
-RUN poetry install --only main --no-root
+RUN poetry install --only main --no-root --no-cache
 
+RUN rm -rf /root/.cache $VENV_PATH/src \
```
amarao, line 37: It won't work this way, rm should be in the same layer as source of files.

```
RUN /bin/true &&
    poetry install &&
    rm -rf /root/.cache &&
    find ... etc..
```

Dockerfile
```diff
@@ -16,28 +21,37 @@ ENV PYTHONUNBUFFERED=1 \
...
 
 FROM base as production
 
+ENV VENV_PATH="/.venv"
+
 COPY --from=builder $VENV_PATH $VENV_PATH
 WORKDIR /app
 COPY . .
 
 RUN apt-get clean && find /var/lib/apt/lists/ -type f -delete && chown -R www-data /app/
```
amarao, line 51: And this should be removed in the 'RUN' with apt-install.

Here the sample of one of CI files I do. I usually do it not for reproducible reasons, but for smaller image (so I don't want additional files in any layer, which is, the same goal as with reproducible images).

https://github.com/lidofinance/ci-image/blob/master/Dockerfile

Note how big is RUN stanza for apt and poetry.

In this particular case we also want to purge away python3.11-dev, so both apt and poetry are in the same RUN (layer).

PR #679, opened 2025-04-02
feat(ORC-331): add docker dev image

chasingrainbows:
## Description

1. Security: Docker isolates the environment, reducing the risk of host compromise or data leaks from vulnerable dependencies.
2. Simplified onboarding: Developers can set up the environment quickly without manual dependency management. (+ automatic git hooks install, etc)
3. Consistency: Ensures a uniform setup across machines, minimizing "works on my machine" bugs. Also ensures all developers will use similar poetry version.

## Related Issue/Task
- Related task: ORC-331
- Epic: oracle-v6

## How Has This Been Tested?

## Checklist
- [x] Documentation updated (if required)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM python:3.12.4-slim AS base
 
+ARG POETRY_VERSION=1.3.2
 ARG SOURCE_DATE_EPOCH
 
 RUN apt-get update && apt-get install -y --no-install-recommends -qq \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -11,3 +11,55 @@ reproducible-build-oracle:
 		-t oracle-reproducible-container:local \
 		.
 	docker image inspect oracle-reproducible-container:local --format 'Image hash: {{.Id}}'
+
+DEV_CONTAINER_NAME = oracle-dev-container
+DEV_IMAGE = lidofinance/oracle:dev
+DEV_WORKDIR = /app
+EXEC_CMD = docker exec -w $(DEV_WORKDIR) -it $(DEV_CONTAINER_NAME)
```
madlabman, line 18: Can we have a switch to make it usable inside of a container as well as on a host machine without using docker?

chasingrainbows: I actually don’t want to maintain a Makefile for non-Docker things.

For example, take poetry-lock. Each developer has their own version of Poetry installed locally. I’ve noticed a few times that Roma, Alex, and I were using different Poetry versions, which didn’t match the one we use in the production Docker environment (1.3.2). For instance, version 1.3.2 has a flag --no-update, which should be used, but the latest version of Poetry removed it. We can’t predict in a Makefile which Poetry version is installed on a developer’s host machine. Creating a Makefile that’s robust enough to handle such variations would turn into a mess.

Also, the up command would become more difficult to manage. Right now, it’s reused everywhere, and that’s incredibly convenient.

The Docker stage and this Makefile for local development were created to address these issues as well. Development in a container could become the default approach to unify our dev and prod environments, preventing bugs and other problems. If someone doesn’t want to follow this approach, they can write their own Makefile based on their own software versions / container system / etc.

madlabman: > I actually don’t want to maintain a Makefile for non-Docker things.

The same commands work, what's to maintain here?

> For example, take poetry-lock.

Add a simple github action check to the `poetry.lock`. Everyone will eventually use the correct version of poetry.

> Also, the up command would become more difficult to manage. Right now, it’s reused everywhere, and that’s incredibly convenient.

Why not something like `make sh` that will call `up` and attach to a container's shell. And then execute all the make commands without a wrapper. And if someone wants it can create their own makefile/bash/just/etc to make it execute commands inside of docker without spawning a shell first? At the moment the makefile works only from host machine to docker, but make it other way around and it works in both environments out of the box.

madlabman: > Can you provide any example?

You already have "sh" target. The makefile is available inside of a container. You attach to a shell within a container and run "make something" and it works. Basically what I'm asking for is make it possible to set `EXEC_CMD` to `docker exec ...` only when asked explicitly and keep it empty otherwise.

chasingrainbows: > Basically what I'm asking for is make it possible to set EXEC_CMD to docker exec ... only when asked explicitly and keep it empty otherwise

But anyway, this will force us to remove `up` from all other commands or make it more complicated (it should be ignored when we run make in local mode and not when docker mode). Not all commands in Makefile even suitable for running inside of container (up/rebuild/reproducible-build-oracle), so it will bring additional confusion.

> The makefile is available inside of a container. You attach to a shell within a container and run "make something" and it works

Also this will disrupt host machine terminal workflow. Right now, terminal autocomplete and search in history works well because you run all make commands from ur host, if i will move it such way, Makefile will become for the local development first (autocomplete and search will work for local development only), which is not goal of this task.

Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM python:3.12.4-slim AS base
 
+ARG POETRY_VERSION=1.3.2
```
F4ever, line 3: let's use latest poetry version, this one too old

Dockerfile
```diff
@@ -16,29 +17,44 @@ ENV PYTHONUNBUFFERED=1 \
...
-ENV POETRY_VERSION=1.3.2
-RUN pip install --no-cache-dir poetry==$POETRY_VERSION
+ARG POETRY_VERSION
+RUN pip install --no-cache-dir poetry==${POETRY_VERSION}
 
 WORKDIR /
 COPY pyproject.toml poetry.lock ./
 
 # Building lru-dict from source for reproducible .so files by enforcing consistent CFLAGS across builds
 RUN poetry config --local installer.no-binary lru-dict && \
```
F4ever, line 36: Do we need lru-dict, since we updated web3py?

Dockerfile
chasingrainbows: Moved into separate task

Dockerfile
chasingrainbows: No need anymore, deleted.

F4ever (approved):

PR #690, opened 2025-04-07
Remove delayed validators from predicate

hweawer:


src/services/exit_order_iterator.py
```diff
@@ -253,8 +229,7 @@ def _eject_validator(self, gid: NodeOperatorGlobalIndex) -> LidoValidator:
 
     def _no_predicate(self, node_operator: NodeOperatorStats) -> tuple:
         return (
-            node_operator.delayed_validators,
-            - self._no_force_predicate(node_operator),
+            self._no_force_predicate(node_operator),
```
F4ever, line 232: Bug over here
```suggestion
            - self._no_force_predicate(node_operator),
```
Pls, write test that sorting predicates correctly sorted (desc in this case)

F4ever:
Also, `ACCOUNTING_DELAYED_VALIDATORS` metric should be removed

F4ever (approved):

PR #693, opened 2025-04-14
[ORC 366] Fix failing tests

F4ever:


tests/integration/contracts/test_lido.py
```diff
@@ -15,17 +15,17 @@ def test_lido_contract_call(lido_contract, accounting_oracle_contract, burner_co
...
-                    1397139100547000000000,
+                    389746,
+                    9303747261167584000000000,
+                    1337759360109000000000,
                     119464421677104745350,
                     0,
                     accounting_oracle_contract.address,
                     20390705,
                     # Call depends on contract state
-                    '0xfdc77ad0ea1ed99b1358beaca0d9c6fa831443f7f4c183302d9e2f76e4c9d0cb',
```
chasingrainbows, line 28: Why old address doesnt work, if chain remain the same?

F4ever: It's not the address, but recent block hash (few hours old). Currently, this is just hot fix until Vaults will be released, there will be some changes done in this method.

chasingrainbows (approved):

PR #710, opened 2025-05-14
feat: Upgrade web3-multi-http-provider

hweawer:
Adding metrics only to EL


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
```
chasingrainbows, line 1: Current poetry version - 1.3.2. Try `make poetry-lock`.

chasingrainbows (approved):

PR #735, opened 2025-07-26
[CSM v2] fix: missing pivot slot in block roots

vgorkavenko:
## Description
There is a case when slot on pivot_index is missing and have the same root as the last slot from the previous checkpoint. Now we are not marking it as missing and it causes double root processing. The fix is simple: get slot number of root on pivot_index and compare it with calculated pivot slot max(checkpoint_slot - SLOTS_PER_HISTORICAL_ROOT, 0) and if they are not equal it means that pivot slot is missing

## Checklist
- [x] New tests added

### DIFF
--- a/src/modules/csm/checkpoint.py
+++ b/src/modules/csm/checkpoint.py
@@ -170,15 +170,14 @@ class FrameCheckpointProcessor:
         # If `s` is `checkpoint_slot -> state.slot`, then it cannot yet be in `block_roots`.
         # So it is the index that will be overwritten in the next slot, i.e. the index of the oldest root.
         pivot_index = checkpoint_slot % SLOTS_PER_HISTORICAL_ROOT
-
-        # Replace duplicated roots with `None` to mark missing slots
-        br = [br[i] if i == pivot_index or br[i] != br[i - 1] else None for i in range(len(br))]
-
         # The oldest root can be missing, so we need to check it and mark it as well as other missing slots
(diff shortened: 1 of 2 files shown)


src/modules/csm/checkpoint.py
```diff
@@ -165,13 +170,27 @@ def exec(self, checkpoint: FrameCheckpoint) -> int:
...
-        return [br[i] if i == pivot_index or br[i] != br[i - 1] else None for i in range(len(br))]
+        # The oldest root can be missing, so we need to check it and mark it as well as other missing slots
+        pivot_block_root = br[pivot_index]
+        slot_by_pivot_block_root = self.cc.get_block_header(pivot_block_root).data.header.message.slot
+        calculated_pivot_slot = max(checkpoint_slot - SLOTS_PER_HISTORICAL_ROOT, 0)
+        is_pivot_missing = slot_by_pivot_block_root != calculated_pivot_slot
+
+        # Replace duplicated roots with `None` to mark missing slots
+        br = [
+            br[i] if br[i] != ZERO_BLOCK_ROOT and (i == pivot_index or br[i] != br[i - 1])
```
chasingrainbows, line 186: On the first loop iteration br[i - 1] will be br[-1], so first element will be compared with the last, is it supposed to work this way?

vgorkavenko: Yes

PR #746, opened 2025-08-07
fix: fetch used module operators keys and add sanity checks

### LINKED ISSUE #740: [CSM] Possibility to influence the report (and rewards distribution) by uploading validator keys already existing on CL.
A Node Operator in CSM performance oracle can specify a validator key that is already active on the CL. In this scenario, the Node Operator may receive staking rewards for validators they are not actually running.

Steps to Reproduce
- Add a validator key that already exists on the CL to the CSM with a corresponding bond.
- This key must be present in the CSM module at the time of report building.
- The CSM collects duties for all keys, regardless of whether the key was deposited through the module.
- The CSM rewards tree will include rewards allocated to the NO.

How to fix:
This issue arises due to the lack of filtering by used=true when fetching data from KAPI for reward distribution calculation in the get_module_validators_by_node_operators method, as implemented in [this code](https://github.com/lidofinance/lido-oracle/blob/ff2af8a9a591971ab9b3411b275bd9dce6860b14/src/web3py/extensions/lido_validators.py#L199).

madlabman:
## Description

Fixes #740

## How Has This Been Tested?

## Checklist

### DIFF
--- a/src/providers/keys/client.py
+++ b/src/providers/keys/client.py
@@ -16,9 +16,18 @@ class KAPIClientError(NotOkResponse):
     pass
 
 
+class KAPIInconsistentData(Exception):
+    pass
+
+
+class KAPIModule(TypedDict):
+    id: int
+    stakingModuleAddress: str
+
+
 class ModuleOperatorsKeys(TypedDict):
     keys: List[LidoKey]
-    module: dict
+    module: KAPIModule
     operators: list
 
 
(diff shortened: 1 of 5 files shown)


src/providers/keys/client.py
```diff
@@ -61,17 +70,24 @@ def _get_with_blockstamp(self, url: str, blockstamp: BlockStamp, params: dict |
     @lru_cache(maxsize=1)
     def get_used_lido_keys(self, blockstamp: BlockStamp) -> list[LidoKey]:
         """Docs: https://keys-api.lido.fi/api/static/index.html#/keys/KeysController_get"""
-        return [LidoKey.from_response(**x) for x in self._get_with_blockstamp(self.USED_KEYS, blockstamp)]
+        data = list(map(lambda x: LidoKey.from_response(**x), self._get_with_blockstamp(self.USED_KEYS, blockstamp)))
```
vgorkavenko, line 73: I think we should left list comprehension here.

vgorkavenko (approved):

PR #749, opened 2025-08-08
Feat: Revert 5.2 and 5.3

F4ever:
Reverts 5.2 and 5.3 releases


chasingrainbows (dismissed):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "lido-oracle"
-version = "5.3.0"
+version = "5.1.0"
```
madlabman, line 3: I guess we have to bump the version anyway?

src/variables.py
```diff
@@ -41,7 +41,7 @@
 
 # Transactions fee calculation variables
 MIN_PRIORITY_FEE: Final = int(os.getenv('MIN_PRIORITY_FEE', 50_000_000))
-MAX_PRIORITY_FEE: Final = int(os.getenv('MAX_PRIORITY_FEE', 100_000_000_000))
+MAX_PRIORITY_FEE: Final = int(os.getenv('MIN_PRIORITY_FEE', 100_000_000_000))
```
madlabman, line 44: It's wrong, it seems

madlabman (changes requested):

PR #789, opened 2025-09-27
Fix: Finalization batches build

F4ever:
In case of multiple finalization batches, the caching function receives an unhashable type, which raises an error when provided as an argument to a function decorated with lru_cache.
To fix this, we removed the cache from this function as a fast and simple solution.


TheDZhon (approved):
LGTM

madlabman (approved):

Psirex (approved):
LGTM👍🏻

avsetsin (approved):

damianrusinek (approved):

PR #795, opened 2025-10-01
wip: srv3 accounting

eddort:


eddort:
TODO:
- remove _kapi_sanity_check, we do not store the number of validators in MaxEB.

PR #798, opened 2025-10-10
fix: lazy oracle get_validator_stages

tamtamchik:
## Description
Fixes get_validator_stages

## How Has This Been Tested?

## Checklist


tests/modules/accounting/staking_vault/test_ifps_report.py
```diff
@@ -46,7 +46,9 @@ def test_ipfs_window(self, web3_integration):
         latest_onchain_ipfs_report_data_2 = sv.get_latest_onchain_ipfs_report_data(
             block_identifier=HexStr(tree.block_hash)
         )
-        bb_2 = web3_integration.ipfs.fetch(latest_onchain_ipfs_report_data_2.report_cid, FrameNumber(0))
-        tree_2 = StakingVaultIpfsReport.parse_merkle_tree_data(bb_2)
 
-        assert tree_2.block_number != tree.block_number
+        if latest_onchain_ipfs_report_data_2.report_cid:
```
chasingrainbows: If report_cid is None/empty, the test will simply skip this validation and pass. If something goes wrong, it's unclear whether the test passed because the assertion was correct or because it was skipped. Either assert that report_cid exists (if it should always be present) or make the skip explicit with proper reasoning.

chasingrainbows (dismissed):
If could be removed from test

tests/modules/accounting/staking_vault/test_ifps_report.py
tamtamchik: Fixed in f3343b6091582e0eec2d4382a024b3722d46fdb8

chasingrainbows (approved):

PR #802, opened 2025-10-17
feat: separate performance collection and distribution

vgorkavenko:
## Description

Separate Performance Collection module for CSMv3 and CMv2 Oracle daemons
https://hackmd.io/OkdagPjpTYKm7yN9sqHGhA?both

## Summary 

- Introduced the performance sidecar stack: `performance_collector` (collects/writes data) + `performance_web_server` (HTTP API), with Prometheus metrics and healthchecks.
- Added a reference `docker-compose.yml` for the Staking Module stack (Postgres + sidecars + `csm`/`cm`) and DB init script `scripts/init_performance_db.sh`.
- Added monitoring examples under `docs/monitoring/*` (Prometheus/Alertmanager configs + alert rules for performance sidecars and CSM/CM).
- Refactored module structure: extracted common runtime into `src/modules/oracles/common/*`, updated entrypoints/import paths.
- Generalized “CSM” into “staking module(s)”: added Curated Module oracle (`cm`) alongside Community Staking (`csm`).
- Updated checks and fork tests: switched to `STAKING_MODULE_ADDRESS`; added `CS_MODULE_ADDRESS` / `CURATED_MODULE_ADDRESS` for checks/fork runs; checks now skip on incompatible staking module contract versions.

## Related Issue/Task
- Related task: [CS-873](https://linear.app/lidofi/issue/CS-873/separate-collect-and-distribute-steps-in-oracle)

## How Has This Been Tested?
- [X] Local tests
- [X] Manual testing

## Checklist
- [X] Documentation updated (if required)


pyproject.toml
```diff
@@ -30,6 +30,9 @@ oz-merkle-tree = { git = "https://github.com/lidofinance/oz-merkle-tree", rev =
 multiformats = "^0.3.1"
 protobuf="^6.31.1"
 dag-cbor="^0.3.3"
+flask = "^3.0.0"
+waitress = "^3.0.2"
+pyroaring = "^1.0.3"
```
madlabman: Let's consider getting rid of the dependency.

src/modules/performance_collector/codec.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class AttDutiesMissCodec:
+
+    @staticmethod
+    def encode(misses: AttDutyMisses) -> bytes:
+        bm = BitMap(sorted(v for v in misses))
+        bm.shrink_to_fit()
+        bm.run_optimize()
+        return bm.serialize()
```
madlabman: Maybe simple run‑length encoding will work pretty much the same here.

src/modules/performance_collector/codec.py
```diff
@@ -0,0 +1,132 @@
...
+        header = struct.pack(cls.HEADER_FMT, cls.VERSION, len(att_bytes), len(proposals), len(syncs))
+        return header + prop_bytes + sync_bytes + att_bytes
+
+    @classmethod
+    def decode(cls, blob: bytes) -> EpochData:
+        if len(blob) < cls.HEADER_SIZE:
+            raise ValueError(f"Epoch blob too short to decode: header size is {cls.HEADER_SIZE} but full blob size is {len(blob)}")
+        ver, att_count, prop_count, sync_count = struct.unpack_from(cls.HEADER_FMT, blob, 0)
+        if ver != cls.VERSION:
+            raise ValueError(f"Unsupported epoch blob version: {ver}")
```
madlabman: Manual intervention is required then?

src/modules/csm/csm.py
```diff
@@ -162,66 +199,73 @@ def validate_state(self, blockstamp: ReferenceBlockStamp) -> None:
...
+                    continue
+
+                logger.info({
+                    "msg": "Requesting performance data from collector",
+                    "epoch": epoch
+                })
+                epoch_data = self.w3.performance.get_epoch(epoch)
+                if epoch_data is None:
+                    logger.warning({"msg": f"Epoch {epoch} is missing in Performance Collector"})
+                    continue
```
madlabman: It has to be an error, since we call the function when the range is available.

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+        finally:
+            self._conn.commit()
+
+    def store_demand(self, consumer: str, l_epoch: int, r_epoch: int) -> None:
+        with self.connection() as cur:
+            cur.execute(
+                "INSERT OR REPLACE INTO epochs_demand(consumer, l_epoch, r_epoch) VALUES(?, ?, ?)",
+                (consumer, l_epoch, r_epoch),
+            )
+            cur.execute("UPDATE epochs_demand_nonce SET value = value + 1")
```
madlabman: Make it a `updated_at` field in the `epoch_demands` table?

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+        self._conn.execute(
+            """
+            CREATE TABLE IF NOT EXISTS epochs_demand_nonce
+            (
+                value INTEGER NOT NULL
+            )
+            """
+        )
+        self._conn.execute("INSERT INTO epochs_demand_nonce (value) VALUES (0);")
```
madlabman: It feels like a separate function like `migrate`.

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+from src.types import EpochNumber
+from src.utils.range import sequence
+
+
+class DutiesDB:
+    def __init__(self, path: str):
+        self._path = path
+        self._conn = sqlite3.connect(
+            self._path, check_same_thread=False, timeout=variables.PERFORMANCE_COLLECTOR_DB_CONNECTION_TIMEOUT
+        )
```
madlabman: Doesn't it make sense to have it in the `connection` context manager instead?

madlabman: I guess `check_same_thread=False` doesn't make sense here because the web server creates a new connection for every request anyway.

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+        if self._conn:
+            self._conn.close()
+            self._conn = None
+
+    @contextmanager
+    def connection(self):
+        try:
+            yield self._conn.cursor()
+        finally:
+            self._conn.commit()
```
madlabman: Don't quite understand why commit if failed?

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+        self._conn.execute("INSERT INTO epochs_demand_nonce (value) VALUES (0);")
+        self._conn.commit()
+
+    def __del__(self):
+        if self._conn:
+            self._conn.close()
+            self._conn = None
+
+    @contextmanager
+    def connection(self):
```
madlabman:
```suggestion
    def cursor(self):
```

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+            """
+        )
+        self._conn.execute("INSERT INTO epochs_demand_nonce (value) VALUES (0);")
+        self._conn.commit()
+
+    def __del__(self):
+        if self._conn:
+            self._conn.close()
+            self._conn = None
```
madlabman: > It is not guaranteed that `__del__()` methods are called for objects that still exist when the interpreter exits.

I'm not sure if it's a problem at all, I guess the OS will remove any file handles and free memory anyway.

src/modules/performance_collector/db.py
madlabman: What if we have an optional `_conn` field that is opened via a separate method such as connect and reused in the`connection` context manager? The context manager checks whether a connection has been established and uses it, or opens a new one otherwise. Thus, clients that prefer to keep the connection open can use the `connect` method and close it eventually, while other clients can use the `connection` context manager. I would also consider using the latter only. I think it is cheap to open a connection, do whatever is needed and close it afterwards.

madlabman: To ensure thread safety I would also add an assert that we in the preferred mode of operation, see [threadsafety](https://docs.python.org/3/library/sqlite3.html#sqlite3.threadsafety).

src/modules/performance_collector/performance_collector.py
```diff
@@ -0,0 +1,172 @@
...
+
+        min_epoch_in_db = self.db.min_epoch()
+        max_epoch_in_db = self.db.max_epoch()
+        if not min_epoch_in_db and not max_epoch_in_db:
+            logger.info({
+                "msg": "Empty Performance Collector DB. Start with the default range calculation",
+                "start_epoch": start_epoch,
+                "end_epoch": end_epoch
+            })
+            return start_epoch, end_epoch
```
madlabman: A request for an epoch range might already exist though.

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+    def has_epoch(self, epoch: int) -> bool:
+        with self.connection() as cur:
+            cur.execute("SELECT 1 FROM duties WHERE epoch=? LIMIT 1", (int(epoch),))
+            ok = cur.fetchone() is not None
+        return ok
+
+    def min_epoch(self) -> int:
+        with self.connection() as cur:
+            cur.execute("SELECT MIN(epoch) FROM duties")
+            val = int(cur.fetchone()[0] or 0)
```
madlabman: `0` is a valid epoch I think.

src/modules/performance_collector/db.py
```diff
@@ -0,0 +1,168 @@
...
+        return val
+
+    def max_epoch(self) -> int:
+        with self.connection() as cur:
+            cur.execute("SELECT MAX(epoch) FROM duties")
+            val = int(cur.fetchone()[0] or 0)
```
madlabman: `0` is a valid epoch I think.

src/modules/performance_collector/codec.py
vgorkavenko: It should prevent reading outdated blobs from Performance Collector that we aren't able to decode due to different version of encoding algo

src/modules/performance_collector/performance_collector.py
vgorkavenko: It was implemented intentionally, but might be revised.

src/modules/oracles/staking_modules/base.py
```diff
@@ -162,66 +231,85 @@ def validate_state(self, blockstamp: ReferenceBlockStamp) -> None:
...
+        finalized_blockstamp = self._receive_last_finalized_slot()
+        validators = self.w3.cc.get_validators(finalized_blockstamp)
+
+        logger.info({
+            "msg": "Starting state fulfillment",
+            "total_frames": len(self.state.frames),
+            "total_validators": len(validators)
+        })
+
+        for l_epoch, r_epoch in self.state.frames:
```
chasingrainbows, line 267: How to understand that `self.state.frames` is not empty? 
I mean, looks like fulfill_state need some pre-requirments before run. If yes - better to check this requirments and drop exceptions if the requirements are not met.

src/modules/oracles/staking_modules/base.py
```diff
@@ -162,66 +231,85 @@ def validate_state(self, blockstamp: ReferenceBlockStamp) -> None:
...
+            logger.info({
+                "msg": "Processing frame",
+                "start_epoch": l_epoch,
+                "end_epoch": r_epoch,
+                "total_epochs": r_epoch - l_epoch + 1
+            })
+
+            for epoch in sequence(l_epoch, r_epoch):
+                if epoch not in self.state.unprocessed_epochs:
```
chasingrainbows, line 276: The same here

src/modules/csm/csm.py
```diff
@@ -162,66 +231,85 @@ def validate_state(self, blockstamp: ReferenceBlockStamp) -> None:
...
+
+                logger.info({
+                    "msg": "Requesting performance data from collector",
+                    "epoch": epoch
+                })
+                epoch_data = self.w3.performance.get_epoch_data(epoch)
+                if epoch_data is None:
+                    raise ValueError(f"Epoch {epoch} is missing in Performance Collector")
+
+                misses, props_vids, props_flags, syncs_vids, syncs_misses = (
```
chasingrainbows: Add types here

src/modules/performance/collector/checkpoint.py
```diff
@@ -183,13 +173,20 @@ def _get_block_roots(self, checkpoint_slot: SlotNumber):
...
         br = [
-            br[i] if br[i] != ZERO_BLOCK_ROOT and (i == pivot_index or br[i] != br[i - 1])
-            else None
+            br[i] if br[i] != ZERO_BLOCK_ROOT and (i == pivot_index or br[i] != br[i - 1]) else None
             for i in range(len(br))
         ]
         if is_pivot_missing:
             br[pivot_index] = None
 
+        logger.debug({
```
chasingrainbows:
```suggestion
        logger.info({
```

maybe

src/modules/performance/collector/checkpoint.py
```diff
@@ -394,47 +369,58 @@ def _get_dependent_root_for_proposer_duties(
...
-    committee = committees[slot]
+def process_sync(
+    sync_aggregate: SyncAggregate,
+    sync_duties: list[SyncDuty]
+) -> None:
     # Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#syncaggregate
     sync_bits = hex_bitvector_to_list(sync_aggregate.sync_committee_bits)
-    for index_in_committee in get_set_indices(sync_bits):
-        committee[index_in_committee].included = True
+    # Go through only UNSET indexes to get misses
```
chasingrainbows: Comments should answer "why", not "what". Code represents "what" by itself.

src/modules/sidecars/performance/collector/checkpoint.py
```diff
@@ -394,47 +369,58 @@ def _get_dependent_root_for_proposer_duties(
...
+    sync_aggregate: SyncAggregate,
+    sync_duties: list[SyncDuty]
+) -> None:
     # Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#syncaggregate
     sync_bits = hex_bitvector_to_list(sync_aggregate.sync_committee_bits)
-    for index_in_committee in get_set_indices(sync_bits):
-        committee[index_in_committee].included = True
+    # Go through only UNSET indexes to get misses
+    for index_in_committee in get_unset_indices(sync_bits):
+        sync_duties[index_in_committee].missed_count += 1
```
chasingrainbows, line 432: imo better to return modified dict explicitly

src/modules/sidecars/performance/collector/checkpoint.py
```diff
@@ -447,6 +433,11 @@ def get_set_indices(bits: Sequence[bool]) -> list[int]:
     return [i for i, bit in enumerate(bits) if bit]
 
 
+def get_unset_indices(bits: Sequence[bool]) -> list[int]:
```
chasingrainbows, line 472: Im not sure this code needs a separate function

src/modules/sidecars/performance/collector/collector.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Continuously collects performance data from Consensus Layer into db for the given epoch range.
+    """
+    # Timestamp of the last epochs demand update
+    last_epochs_demand_update: int = 0
+
+    def __init__(self, w3):
+        super().__init__(w3)
+        self.db = DutiesDB()
+        self.last_epochs_demand_update = self.get_epochs_demand_max_updated_at()
```
chasingrainbows, line 45: How much records get_epochs_demand_max_updated_at will return? Underhood method perform select of whole table - this is potentially dangerous. Maybe we should add some sort of chunking inside.

src/modules/sidecars/performance/collector/collector.py
```diff
@@ -0,0 +1,157 @@
...
+                # Remove from the DB just in case
+                self.db.delete_demand(demand.consumer)
+            start_epoch = min(start_epoch, demand.l_epoch)
+
+        missing_epochs = self.db.missing_epochs_in(start_epoch, end_epoch)
+        if missing_epochs:
+            start_epoch = min(missing_epochs)
+        else:
+            # Start from the next epoch after the last epoch in the DB.
+            start_epoch = EpochNumber(max_epoch_in_db + 1)
```
chasingrainbows, line 180: Looks like self.db.max_epoch able to return None which can leads to TypeError here

src/modules/performance/collector/collector.py
```diff
@@ -0,0 +1,157 @@
...
+        max_updated_at = self.get_epochs_demand_max_updated_at()
+        updated = self.last_epochs_demand_update != max_updated_at
+        if updated:
+            self.last_epochs_demand_update = max_updated_at
+            return True
+        return False
+
+    def get_epochs_demand_max_updated_at(self) -> int:
+        max_updated_at = 0
+        for demand in self.db.get_epochs_demands():
```
chasingrainbows: You can replace it with something just like `SELECT MAX(updated_at) FROM epochs_demands`

src/modules/sidecars/performance/common/db.py
```diff
@@ -0,0 +1,189 @@
...
+            pool_recycle=3600,  # Recycle connections every hour
+            pool_size=10,  # Connection pool size
+            max_overflow=20,  # Maximum overflow connections
+        )
+        self._setup_database()
+
+    @staticmethod
+    def _get_database_url() -> str:
+        """Get PostgreSQL database URL from environment variables"""
+        host = variables.PERFORMANCE_DB_HOST
```
chasingrainbows, line 65: I believe we need to add validation to all of these parameters. We constantly have issues with holders with incorrect values.

src/modules/performance/common/db.py
```diff
@@ -0,0 +1,189 @@
...
+        if variables.PERFORMANCE_COLLECTOR_DB_RETENTION_EPOCHS <= 0:
+            return
+        threshold = int(current_epoch) - variables.PERFORMANCE_COLLECTOR_DB_RETENTION_EPOCHS
+        if threshold <= 0:
+            return
+
+        with self.get_session() as session:
+            duties_to_delete = session.exec(select(Duty).where(Duty.epoch < threshold)).all()
+            for duty in duties_to_delete:
+                session.delete(duty)
```
chasingrainbows: Why not something like `session.exec(delete(Duty).where(Duty.epoch < threshold))` with a single query?

src/modules/performance/common/db.py
```diff
@@ -0,0 +1,189 @@
...
+            session.commit()
+
+    def is_range_available(self, l_epoch: EpochNumber, r_epoch: EpochNumber) -> bool:
+        if int(l_epoch) > int(r_epoch):
+            raise ValueError("Invalid epoch range")
+
+        with self.get_session() as session:
+            count = session.exec(select(Duty).where(Duty.epoch >= l_epoch, Duty.epoch <= r_epoch)).all()
```
chasingrainbows: Just use sql built-in count, `select(func.count(...))`

src/variables.py
```diff
@@ -96,13 +104,28 @@
...
+PERFORMANCE_WEB_SERVER_METRICS_PREFIX: Final = os.getenv("PERFORMANCE_WEB_SERVER_METRICS_PREFIX", "lido_performance_web")
+
+PERFORMANCE_COLLECTOR_DB_RETENTION_EPOCHS: Final = int(os.getenv('PERFORMANCE_COLLECTOR_DB_RETENTION_EPOCHS', 28 * 225 * 6))
+PERFORMANCE_COLLECTOR_DB_CONNECTION_TIMEOUT: Final = int(os.getenv('PERFORMANCE_COLLECTOR_DB_CONNECTION_TIMEOUT', 30))
+
+PERFORMANCE_DB_HOST: Final = os.getenv('PERFORMANCE_DB_HOST', 'localhost')
+PERFORMANCE_DB_PORT: Final = int(os.getenv('PERFORMANCE_DB_PORT', 5432))
+PERFORMANCE_DB_NAME: Final = os.getenv('PERFORMANCE_DB_NAME', 'performance')
+PERFORMANCE_DB_USER: Final = os.getenv('PERFORMANCE_DB_USER', 'performance')
+PERFORMANCE_DB_PASSWORD: Final = os.getenv('PERFORMANCE_DB_PASSWORD', 'performance')
```
chasingrainbows, line 133: Add it into PRIVATE_ENV_VARS

src/modules/performance/common/db.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def missing_epochs_in(self, l_epoch: EpochNumber, r_epoch: EpochNumber) -> list[EpochNumber]:
+        if l_epoch > r_epoch:
+            raise ValueError("Invalid epoch range")
+
+        with self.get_session() as session:
+            present_duties = session.exec(
+                select(Duty.epoch).where(Duty.epoch >= l_epoch, Duty.epoch <= r_epoch).order_by(Duty.epoch)
+            ).all()
+            present = [int(epoch) for epoch in present_duties]
```
chasingrainbows: `present` could be set. `if epoch not in present` is not optimal so much if present is list.

src/modules/performance/web/server.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+@app.get("/check-epochs", response_model=bool)
+async def epochs_check(
+    from_epoch: EpochNumber = Query(..., alias="from"),
+    to_epoch: EpochNumber = Query(..., alias="to"),
+    db: DutiesDB = Depends(get_db),
+):
+    await validate_epoch_bounds(from_epoch, to_epoch)
+    return bool(db.is_range_available(from_epoch, to_epoch))
```
chasingrainbows: If you want to use async view, your queries to DB must be async as well, or you will block event loop.

Remove async from views, or add async/await to DutiesDB methods.

src/modules/performance/web/server.py
```diff
@@ -0,0 +1,122 @@
...
+    status: str = "ok"
+
+
+app = FastAPI(title="Performance Collector API")
+attach_metrics(app)
+
+_db_instance: Optional[DutiesDB] = None
+
+
+async def get_db() -> DutiesDB:
```
chasingrainbows: There is probability of race condition here. As i understand, the main approach in fastapi for this is using [lifespan](https://fastapi.tiangolo.com/advanced/events/#lifespan-function) function.

src/modules/performance/web/server.py
```diff
@@ -0,0 +1,122 @@
...
+    to_epoch: EpochNumber = Query(..., alias="to"),
+    db: DutiesDB = Depends(get_db),
+):
+    await validate_epoch_bounds(from_epoch, to_epoch)
+    return db.missing_epochs_in(from_epoch, to_epoch)
+
+
+@app.get("/epochs", response_model=list[Duty])
+async def epochs_data(
+    from_epoch: EpochNumber = Query(..., alias="from"),
```
chasingrainbows: Need limits on from/to + timeouts

src/modules/performance/web/server.py
```diff
@@ -0,0 +1,122 @@
...
+    # Prepare logging config with the app-wise formatter
+    logging_config = LOGGING_CONFIG.copy()
+    for formatter_name in logging_config["formatters"]:
+        logging_config["formatters"][formatter_name] = {
+            "()": JsonFormatter,
+        }
+    
+    uvicorn.run(
+        app,
+        host="0.0.0.0",
```
chasingrainbows: Move to env?

chasingrainbows (changes requested):

PR #814, opened 2025-11-05
feat: correctly handle doppelgänger pubkeys

tamtamchik:
## Description
Fix doppelgänger pubkeys acceptance for TV calculations.

### Initial State

`Vault1`: 1 ETH balance
`Vault2`: 32 ETH balance

#### Step 1: External Predeposit to validator1

Predeposit validator1 using external ETH (not from PDG)
Validator1 configured with vault1's withdrawal credentials (WC1)
Status: Pending validator
vault1 totalValue: 1 ETH (unchanged, since PDG state = None)

#### Step 2: PDG Predeposit to Same Validator

Use PDG to predeposit to validator1 again
This time with vault2's withdrawal credentials (WC2)
Deposit: 1 ETH from vault2, stage remaining 31 ETH
vault2 totalValue: 31 ETH (32 - 1 = 31, since validator doesn't point to vault2 yet)

#### Step 3: The Exploit
**vault1 totalValue becomes 2 ETH!** 
Why?

vault1 still has 1 ETH in balance
validator1 now has PDG state = PREDEPOSITED (with WC2)
BUT: System incorrectly attributes the predeposited validator to vault1 because it was originally associated with WC1
vault1 gains +1 ETH credit for the predeposited validator

#### Step 4: Borrow and Compensate Race

vault1 borrows against the inflated 2 ETH totalValue
System then compensates vault2 for the staged ETH
vault2 totalValue: Returns to 32 ETH
vault1 totalValue: Drops back to 1 ETH (but debt already issued)
Result: Bad debt created

#### Step 5: Parallel Exploitation
Repeat steps 1-4 multiple times in parallel to amplify bad debt creation.

### Root Cause
The system fails to properly handle validator credential switches, allowing double-counting of validator value across vaults during the window between predeposit and compensation.

## Related Issue/Task
- Related branches: https://github.com/lidofinance/core/pull/1546

## How Has This Been Tested?

## Checklist


src/providers/execution/contracts/lazy_oracle.py
```diff
@@ -105,28 +106,35 @@ def get_all_vaults(self, block_identifier: BlockIdentifier = 'latest') -> list[V
...
                 'block_identifier': repr(block_identifier),
                 'to': self.address,
             })
 
-            out.update({ pk.to_0x_hex(): ValidatorStage(stage) for pk, stage in zip(batch, response) })
+            out.update({
+                pk.to_0x_hex(): ValidatorStatus(
+                    stage=ValidatorStage(status.stage) if isinstance(status.stage, ValidatorStage) else ValidatorStage.NONE,
+                    staking_vault=status.stakingVault,
+                    node_operator=status.nodeOperator,
```
TheDZhon: didn't get why for stage we handle missing field, but for stakingVault and nodeOperator we don't

TheDZhon (dismissed):
LGTM, though have a question

src/modules/accounting/types.py
```diff
@@ -424,3 +424,10 @@ class ValidatorStage(Enum):
     PROVEN = 2
     ACTIVATED = 3
     COMPENSATED = 4
+
+
+@dataclass(frozen=True)
+class ValidatorStatus:
+    stage: ValidatorStage
```
F4ever, line 431: If this can be none - it should be marked as optional

src/services/staking_vaults.py
```diff
@@ -116,10 +117,14 @@ def get_vaults_total_values(
                 # - ACTIVATED: add full balance + pending deposits
                 # All other stages are skipped as not related to the non-eligible for activation validators
                 else:
-                    stage = inactive_validator_stages.get(validator_pubkey, ValidatorStage.NONE)
-                    if stage == ValidatorStage.PREDEPOSITED:
+                    status = inactive_validator_statuses.get(validator_pubkey)
```
F4ever, line 120: Is there a chance non vault's key can be stored in `validators_by_vault` structure?

src/services/staking_vaults.py
```diff
@@ -116,10 +117,14 @@ def get_vaults_total_values(
...
+                    # Skip if validator pubkey in PDG is not associated with the current vault
+                    if status is None or status.staking_vault != vault_address:
+                        continue
+
+                    if status.stage == ValidatorStage.PREDEPOSITED:
                         vault_total += int(gwei_to_wei(MIN_DEPOSIT_AMOUNT))
```
F4ever, line 126: Why we can't use actual validators balance?

src/services/staking_vaults.py
tamtamchik: No, we built this structure based on the `vaults` list that was returned from the on-chain.

src/services/staking_vaults.py
tamtamchik: This is because, on-chain, until proven, we have only 1 ETH locked in a pre-deposit guarantee for vault. But the idea of taking all the balance needs to be checked.

tests/modules/accounting/staking_vault/test_staking_vaults.py
```diff
@@ -453,6 +454,79 @@ def test_existing_validator_with_wrong_withdrawal_credentials(self):
...
+            # predeposited but for vault 1 wc
+            pubkey: ValidatorStatus(
+                stage=ValidatorStage.PREDEPOSITED,
+                staking_vault=self.vault_adr_1,
+                node_operator="0x0000000000000000000000000000000000000000",
+            )
+        }
+
+        # --- Web3 Mock ---
+        w3_mock = MagicMock()
```
chasingrainbows, line 498: There is a better mock, that can be used instead, [in conftest.py](https://github.com/lidofinance/lido-oracle/blob/2184a8590ec651a2c51718d21347e2d865d7c534/tests/conftest.py#L113)

chasingrainbows (approved):

src/modules/accounting/types.py
tamtamchik: it's not optional, it can be NONE in enum

F4ever (approved):

F4ever:
One integration test fails - expected. Will be fixed after protocol upgrade

PR #816, opened 2025-11-11
fix: pending validators in total value

### LINKED ISSUE #817: [stVaults] Total Value (TV) miscalculation for validators in pending deposit state
The `get_vaults_total_values` method does not correctly account for validators that exist only as pending deposits and have not yet been converted to actual validators on the consensus layer.

#### Current Behavior
The TV calculation logic assumes all validators exist in the validators list and checks their activation eligibility status. However, when a validator is only in the pending deposit state (e.g., deposit transaction submitted but not yet processed by the beacon chain), it won't appear in the validators list at all, causing its value to be completely excluded from the vault's TV calculation.

#### Expected Behavior
Validators in pending deposit state should contribute to the vault's TV based on their PDG status (in PREDEPOSITED), even before they appear as actual validators on the consensus layer.

#### Impact
Vault TV appears artificially lower than the actual deposited value
Creates a temporary "gap" in TV reporting between deposit submission and validator activation
Misleading metrics for vaults with recently submitted deposits

#### Root Cause
The method iterates only over `validators_by_vault`, which contains validators that already exist on the consensus layer. Pending deposits that haven't been converted to validators yet are present in `pending_deposits` but have no corresponding entry in the validators list to trigger their inclusion in TV.

#### Solution
Need to add logic to handle pending deposits that don't have corresponding validators yet, possibly by:

- Iterating through `pending_deposits` to identify orphaned deposits (no matching validator)
- Mapping these deposits to their respective vaults
- Adding their 1 ETH to the vault's TV in case the PDG status (in PREDEPOSITED)

tamtamchik:
## Description
Fix for the issue found by @Psirex
We are missing a pending deposits statuses check in the total value calculation.

## Related Issue/Task
- Issue: #817 
- Epic: stValuts, Oracle v7

## How Has This Been Tested?

## Checklist


src/services/staking_vaults.py
```diff
@@ -135,54 +119,140 @@ def get_vaults_total_values(
...
+            # All other stages are skipped as not related to the non-eligible for activation validators
+            else:
+                status = validator_statuses.get(validator_pubkey)
+                # Skip if validator pubkey in PDG is not associated with the current vault
+                if status is None or status.staking_vault != vault_address:
+                    continue
+
+                if status.stage == ValidatorStage.PREDEPOSITED:
+                    vault_total += int(gwei_to_wei(MIN_DEPOSIT_AMOUNT))
+                elif status.stage == ValidatorStage.ACTIVATED:
```
F4ever, line 315: can validator be activated and have far future activation epoch?

src/services/staking_vaults.py
```diff
@@ -135,54 +119,140 @@ def get_vaults_total_values(
...
+            # - ACTIVATED: add full balance + pending deposits
+            # All other stages are skipped as not related to the non-eligible for activation validators
+            else:
+                status = validator_statuses.get(validator_pubkey)
+                # Skip if validator pubkey in PDG is not associated with the current vault
+                if status is None or status.staking_vault != vault_address:
+                    continue
+
+                if status.stage == ValidatorStage.PREDEPOSITED:
+                    vault_total += int(gwei_to_wei(MIN_DEPOSIT_AMOUNT))
```
F4ever, line 314: What if validator already has pending deposit in 31 ETH. Should we add it already here to check if balance > 32 ETH?

src/services/staking_vaults.py
```diff
@@ -135,54 +119,140 @@ def get_vaults_total_values(
...
+        for deposit in vault_pending_deposits:
+            deposit_pubkey = deposit.pubkey
+            status = validator_statuses.get(deposit_pubkey)
 
-        return vault_to_validators
+            if status is None or status.staking_vault != vault_address:
+                continue
+
+            # Only PREDEPOSITED stage is possible for pending deposits
+            if status.stage == ValidatorStage.PREDEPOSITED:
```
F4ever: how validator could be predeposited if it's not yet on CL?

src/services/staking_vaults.py
```diff
@@ -135,54 +119,140 @@ def get_vaults_total_values(
...
-    ) -> dict[str, Gwei]:
-        deposits: defaultdict[str, int] = defaultdict(int)
-        for deposit in pending_deposits:
-            deposits[deposit.pubkey] += int(deposit.amount)
-
-        return {pubkey: Gwei(deposits[pubkey]) for pubkey in deposits}
-
-    @staticmethod
-    def _get_validators_by_vault(validators: list[Validator], vaults: VaultsMap) -> VaultToValidators:
+    def _calculate_vault_total_value(
```
F4ever, line 237: Can unguaranteed deposits be activated on connected to Lido vault?

src/services/staking_vaults.py
```diff
@@ -95,37 +96,20 @@ def get_vaults_total_values(
             validator becomes eligible for activation.
         """
         validators_by_vault = self._get_validators_by_vault(validators, vaults)
+        unmatched_pending_deposits_by_vault = self._get_pending_deposits_by_vault(pending_deposits, vaults)
```
folkyatina: It states that this pending deposits is "unmatched" but it looks like we just creating mapping between vault address and pending deposits inside. Without matching or unmatching it.

folkyatina (changes requested):
Looks like we skipped filtering out pending deposits with the known validator pubkey.

src/services/staking_vaults.py
tamtamchik: Yes, looks like I forgot to rename back, as all check for matching moved to lookup function

src/services/staking_vaults.py
```diff
@@ -135,54 +119,140 @@ def get_vaults_total_values(
...
             if has_far_future_activation_eligibility_epoch(v.validator)
             and v.validator.withdrawal_credentials in vault_wcs
         ]
 
+        all_validator_pubkeys = [v.validator.pubkey for v in validators]
+        deposit_pubkeys = [
+            deposit.pubkey
+            for deposit in pending_deposits
+            if deposit.withdrawal_credentials in vault_wcs
+            and deposit.pubkey not in all_validator_pubkeys
```
folkyatina: Looks like we'll need all_validator_pubkeys to be a set to avoid iteration over the list here.

src/services/staking_vaults.py
tamtamchik: Yes, but this is a valid TV drop as for unguaranteed deposit we need to withdraw from the vault

src/services/staking_vaults.py
tamtamchik: Stage check against on-chain status, as soon as pre-deposit done on-chain status changes and deposit goes to the CL, and there may stuck in pending.

src/services/staking_vaults.py
tamtamchik: No, because only 1 ETH acts as a guatantee on the EL in PDG

src/services/staking_vaults.py
tamtamchik: Yes, this is the exact case after proving the validator, it's activated on-chain, but 31 eth is coming either as pending or as already applied balance thus we can use `total_validator_balance`

src/services/staking_vaults.py
TheDZhon: yeah, should be just `pending_deposits_by_vault`

src/services/staking_vaults.py
```diff
@@ -135,54 +119,140 @@ def get_vaults_total_values(
...
+                elif status.stage == ValidatorStage.ACTIVATED:
+                    vault_total += int(total_validator_balance)
 
-        vault_to_validators: VaultToValidators = defaultdict(list)
-        for validator in validators:
-            wc = validator.validator.withdrawal_credentials
 
-            if vault_info := wc_to_vault.get(wc):
-                vault_to_validators[vault_info.vault].append(validator)
+        for deposit in vault_pending_deposits:
```
TheDZhon: ⚠️ there should probably be set to protect against double accounting (pre-seed pending deposit with PDG, and send multiple deposits atop)

TheDZhon (commented):
🎷

src/services/staking_vaults.py
```diff
@@ -135,54 +119,140 @@ def get_vaults_total_values(
...
+        """
+        Calculates the total pending amount for each pubkey.
+        """
+        deposits: defaultdict[str, int] = defaultdict(int)
+        for deposit in pending_deposits:
+            deposits[deposit.pubkey] += int(deposit.amount)
+
+        return {pubkey: Gwei(deposits[pubkey]) for pubkey in deposits}
+
+    def _lookup_pubkey_statuses(
```
TheDZhon: it feels that this fn loads 'unconfirmed keys' from two completely different cases, while it saves computational resource, the long-term support becomes more tricky;

would suggest to separate both cases and maintain them separately

TheDZhon:
Let's try to visualize the flow and invariants here

src/services/staking_vaults.py
tamtamchik: Fixed in 5b2e699

src/services/staking_vaults.py
tamtamchik: Fixed in 38e9b40

src/services/staking_vaults.py
```diff
@@ -85,47 +86,38 @@ def get_vaults_total_values(
             - PREDEPOSITED: add 1 ETH to TV, as only the predeposit is counted and not the validator balance
             - ACTIVATED: count as `already passed activation`, thus add full balance + pending deposits to TV
             - all other stages are skipped as not related to the non-eligible for activation validators
+        3. A pending deposit is included in the TV if it is NOT associated with a validator yet, and the PDG stage is
+           PREDEPOSITED. It can't be ACTIVATED, as activation happenst only for created validators.
```
folkyatina:
```suggestion
           PREDEPOSITED. It can't be ACTIVATED, as activation happens only for created validators.
```

src/services/staking_vaults.py
```diff
@@ -85,47 +86,38 @@ def get_vaults_total_values(
             - PREDEPOSITED: add 1 ETH to TV, as only the predeposit is counted and not the validator balance
             - ACTIVATED: count as `already passed activation`, thus add full balance + pending deposits to TV
             - all other stages are skipped as not related to the non-eligible for activation validators
+        3. A pending deposit is included in the TV if it is NOT associated with a validator yet, and the PDG stage is
```
TheDZhon: not every pending deposit though, bc we basically say: "at least a single pending deposit, with a pubkey not associated with any of the validators, adds always exactly 1 ETH of the guarantee to the vault's TV"

src/services/staking_vaults.py
```diff
@@ -135,54 +127,190 @@ def get_vaults_total_values(
...
+                                     Validators and Pending Deposits Contributions
+                                                          │
+                     ┌────────────────────────────────────┼────────────────────────────────────┐
+                     │                                    │                                    │
+                     ▼                                    ▼                                    ▼
+            ┌─────────────────────────┐        ┌─────────────────────────┐        ┌─────────────────────────┐
+            │   ELIGIBLE VALIDATORS   │        │ NON-ELIGIBLE VALIDATORS │        │  UNMATCHED PENDING      │
+            │                         │        │                         │        │  DEPOSITS               │
+            │  activation_elig ≠      │        │  activation_elig =      │        │                         │
+            │  FAR_FUTURE_EPOCH       │        │  FAR_FUTURE_EPOCH       │        │  (no validator yet)     │
```
TheDZhon:
```suggestion
            │  FAR_FUTURE_EPOCH       │        │  FAR_FUTURE_EPOCH       │        │   pubkey not in any     │
            │                         │        │                         │        │   vault validator)      │
```

src/services/staking_vaults.py
```diff
@@ -135,54 +127,190 @@ def get_vaults_total_values(
...
+                if status is None or status.staking_vault != vault_address:
+                    continue
+
+                if status.stage == ValidatorStage.PREDEPOSITED:
+                    vault_total += int(gwei_to_wei(MIN_DEPOSIT_AMOUNT))
+                elif status.stage == ValidatorStage.ACTIVATED:
+                    vault_total += int(total_validator_balance)
+
+        # Must be set to avoid double counting of pending deposits already associated with validators
+        for deposit_pubkey in {p.pubkey for p in vault_pending_deposits}:
```
TheDZhon:
```suggestion
        # Process only unmatched pending deposits (those without a corresponding validator).
        # Deposits with matching validators are already accounted for in the validator loop above.
        # Using set difference to exclude pubkeys that have validators prevents double-counting
        # when a PREDEPOSITED validator also has pending deposits in flight.
        matched_validator_pubkeys = {v.validator.pubkey for v in vault_validators}
        unmatched_deposit_pubkeys = {p.pubkey for p in vault_pending_deposits} - matched_validator_pubkeys
        for deposit_pubkey in unmatched_deposit_pubkeys:
```

src/services/staking_vaults.py
```diff
@@ -135,54 +127,190 @@ def get_vaults_total_values(
...
+            status = validator_statuses.get(deposit_pubkey)
+            # Skip if deposit pubkey in PDG is not associated with the current vault
+            if status is None or status.staking_vault != vault_address:
+                continue
+
+            # Only PREDEPOSITED stage is possible for pending deposits
```
TheDZhon:
```suggestion
            # Only PREDEPOSITED stage is possible for unmatched pending deposits
```

src/services/staking_vaults.py
TheDZhon, line 1: ⚠️ 

Test case for `test_staking_vaults.py`:

```python
    @pytest.mark.unit
    def test_predeposited_validator_with_pending_deposit_no_double_count(self):
        """
        This test ensures that when a PREDEPOSITED validator also has pending deposits
        in flight (e.g., someone top-ups the validator), we only count 1 ETH (the predeposit)
        and NOT double-count by also processing the pending deposit.

        Scenario:
        - Validator V exists with pubkey P, balance 1 ETH, NOT eligible for activation
        - Pending deposit D exists for the same pubkey P (additional 31 ETH)
        - PDG stage for P is PREDEPOSITED

        Expected: Only 1 ETH is counted (the guaranteed predeposit), not 2 ETH.
        """
        pubkey = '0x862d53d9e4313374d202f2b28e6ffe64efb0312f9c2663f2eef67b72345faa8932b27f9b9bb7b476d9b5e418fea99001'

        validators: list[Validator] = [
            Validator(
                index=ValidatorIndex(1),
                balance=Gwei(1000000000),  # 1 ETH
                validator=ValidatorState(
                    pubkey=pubkey,
                    withdrawal_credentials=self.vault_wc_0,
                    effective_balance=Gwei(1000000000),
                    slashed=False,
                    activation_eligibility_epoch=FAR_FUTURE_EPOCH,
                    activation_epoch=FAR_FUTURE_EPOCH,
                    exit_epoch=FAR_FUTURE_EPOCH,
                    withdrawable_epoch=FAR_FUTURE_EPOCH,
                ),
            ),
        ]

        pending_deposits: list[PendingDeposit] = [
            # Additional pending deposit to the same validator while PREDEPOSITED
            # This could be a top-up or an attacker's deposit
            PendingDeposit(
                pubkey=pubkey,
                withdrawal_credentials=self.vault_wc_0,
                amount=Gwei(31000000000),  # 31 ETH
                signature='0xb5b222b452892bd62a7d2b4925e15bf9823c4443313d86d3e1fe549c86aa8919d0cdd1d5b60d9d3184f3966ced21699f124a14a0d8c1f1ae3e9f25715f40c3e7b81a909424c60ca7a8cbd79f101d6bd86ce1bdd39701cf93b2eecce10699f40b',
                slot=SlotNumber(259388),
            ),
        ]

        validator_statuses = {
            pubkey: ValidatorStatus(
                stage=ValidatorStage.PREDEPOSITED,
                staking_vault=self.vault_adr_0,
                node_operator="0x0000000000000000000000000000000000000000",
            )
        }

        # --- Web3 Mock ---
        w3_mock = MagicMock()
        lazy_oracle_mock = MagicMock()
        lazy_oracle_mock.get_validator_statuses = MagicMock(return_value=validator_statuses)

        w3_mock.lido_contracts.lazy_oracle = lazy_oracle_mock
        self.staking_vaults = StakingVaultsService(w3_mock)

        vaults_total_values = self.staking_vaults.get_vaults_total_values(self.vaults, validators, pending_deposits)

        # Expected: Only 1 ETH (the predeposit), NOT 2 ETH (would be double-counted without the fix)
        expected_vault_adr_0 = sum(
            [
                self.vaults[self.vault_adr_0].aggregated_balance,
                gwei_to_wei(1000000000),  # 1 ETH predeposit only, pending deposit is NOT counted separately
            ]
        )

        expected = {
            self.vault_adr_0: expected_vault_adr_0,
            self.vault_adr_1: 0,
            self.vault_adr_2: 2000900000000000000,
            self.vault_adr_3: 1000000000000000000,
        }

        assert vaults_total_values == expected

```

TheDZhon:
```
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                       CONSENSUS LAYER                                                           │
│                                                                                                                 │
│   PDG STAGE          │  NONE  │  PREDEPOSITED  │     PROVEN     │    ACTIVATED    │   COMPENSATED               │
│   ───────────────────┼────────┼────────────────┼────────────────┼─────────────────┼──────────────               │
│   TV CONTRIBUTION    │   0    │     1 ETH      │       0        │  balance + pend │      0                      │
│                                                                                                                 │
│   ┌─────────────────────────────────────────────────────────────────────────────────────────────┐               │
│   │ IF activation_eligibility_epoch != FAR_FUTURE_EPOCH:                                        │               │
│   │   TV += validator.balance + pending_deposits  (regardless of PDG stage)                     │               │
│   └─────────────────────────────────────────────────────────────────────────────────────────────┘               │
│                                                                                                                 │
│   ┌─────────────────────────────────────────────────────────────────────────────────────────────┐               │
│   │ ATTACK VECTORS MITIGATED:                                                                   │               │
│   │ ✓ Front-running: PDG vault address must match                                               │               │
│   │ ✓ Side deposits: Only counted after activation eligibility                                  │               │
│   │ ✓ Doppelganger: Cross-vault pubkey reuse rejected                                           │               │
│   │ ✓ Double-count: PREDEPOSITED validator + pending deposit counted as 1 ETH                   │               │
│   └─────────────────────────────────────────────────────────────────────────────────────────────┘               │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

TheDZhon (changes requested):
⚠️  it has a bug it seems

src/services/staking_vaults.py
tamtamchik: Fixed in d542612

src/services/staking_vaults.py
tamtamchik: Fixed in d542612

src/services/staking_vaults.py
tamtamchik: Fixed in d542612

src/services/staking_vaults.py
tamtamchik: Fixed in d542612

src/services/staking_vaults.py
tamtamchik: Added in d542612

TheDZhon (dismissed):
LGTM

src/services/staking_vaults.py
```diff
@@ -135,54 +133,196 @@ def get_vaults_total_values(
...
+        return statuses_by_vault
+
+    @staticmethod
+    def _calculate_vault_total_value(
+        vault_address: str,
+        vault_aggregated_balance: Wei,
+        vault_validators: list[Validator],
+        vault_pending_deposits: list[PendingDeposit],
+        total_pending_amount_by_pubkey: dict[str, Gwei],
+        validator_statuses: dict[str, ValidatorStatus],
```
folkyatina:
```suggestion
        vault_validator_statuses: dict[str, ValidatorStatus],
```

src/services/staking_vaults.py
```diff
@@ -135,54 +133,196 @@ def get_vaults_total_values(
...
+                vault_total += int(total_validator_balance)
+
+            # For not-yet-eligible validators, use PDG stages:
+            # - PREDEPOSITED: add 1 ETH (guaranteed)
+            # - ACTIVATED: add full balance + pending deposits
+            # All other stages are skipped as not related to the non-eligible for activation validators
+            else:
+                status = validator_statuses.get(validator_pubkey)
+                # Skip if validator pubkey in PDG is not associated with the current vault
+                if status is None or status.staking_vault != vault_address:
```
folkyatina: looks like the second check is obsolete now but can be left as is

src/services/staking_vaults.py
```diff
@@ -135,54 +133,196 @@ def get_vaults_total_values(
...
+                    vault_total += int(gwei_to_wei(MIN_DEPOSIT_AMOUNT))
+                elif status.stage == ValidatorStage.ACTIVATED:
+                    vault_total += int(total_validator_balance)
+
+        # Process only unmatched pending deposits (those without a corresponding validator).
+        # Deposits with matching validators are already accounted for in the validator loop above.
+        # Using set difference to exclude pubkeys that have validators prevents double-counting
+        # when a PREDEPOSITED validator also has pending deposits in flight.
+        matched_validator_pubkeys = {v.validator.pubkey for v in vault_validators}
+        unmatched_deposit_pubkeys = {p.pubkey for p in vault_pending_deposits} - matched_validator_pubkeys
```
folkyatina: So, here we have another way to calculate unmatched deposits on the vault level?

src/services/staking_vaults.py
```diff
@@ -135,54 +133,196 @@ def get_vaults_total_values(
...
+        for deposit_pubkey in unmatched_deposit_pubkeys:
+            status = validator_statuses.get(deposit_pubkey)
+            # Skip if deposit pubkey in PDG is not associated with the current vault
+            if status is None or status.staking_vault != vault_address:
```
folkyatina: Something ambiguous in this combination of ifs, but looks correct.

folkyatina (commented):
Some notions

folkyatina (dismissed):
🚛

src/services/staking_vaults.py
TheDZhon, line 67: I would avoid using such a dangerous default value here and there tbh for the `block_identifier` param bc it's enormously hard to debug if the error is slipped into the call flows

TheDZhon (dismissed):
LGTM, just a hygienic concerns about the default block identifier values

src/services/staking_vaults.py
tamtamchik: Yes, the concern is valid, and was raised by Certora on review, but this is used project-wide. The fix must be in the backlog sometime later (@F4ever correct me if I'm wrong)

F4ever: Yeah, we have this task for srv3 scope. But I think it could be removed already

src/services/staking_vaults.py
```diff
@@ -135,54 +138,192 @@ def get_vaults_total_values(
...
+
+            # For not-yet-eligible validators, use PDG stages:
+            # - PREDEPOSITED: add 1 ETH (guaranteed)
+            # - ACTIVATED: add full balance + pending deposits
+            # All other stages are skipped as not related to the non-eligible for activation validators
+            else:
+                status = vault_validator_statuses.get(validator_pubkey)
+                # Skip if validator pubkey not found in PDG
+                if status is None:
+                    continue
```
skhomuti, line 311: this check is not quite obvious
currently missed keys from PDG laid in the zero key (as from the empty vault value of batchValidatorStatuses call), so there'll be value with `status.stage == ValidatorStage.NONE` and zero instead of vault address

previously this check was safer
```
# Skip if validator pubkey in PDG is not associated with the current vault
if status is None or status.staking_vault != vault_address:
   continue
```
but I think it'd be better filter out zeroed vault from the results of `_get_pubkey_statuses_by_vault`

F4ever (approved):

src/services/staking_vaults.py
tamtamchik: Thanks! Definitely something to consider, but all those pubkeys will be added to the set under `zero vault address` and will not be applied to any vault.

PR #818, opened 2025-12-03
CM v2 - CSM v3 Support | Performance Collector

vgorkavenko:
## Description

CM v2 - CSM v3 Support

AND

Separate Performance Collection module for CSMv3 and CMv2 Oracle daemons
https://hackmd.io/OkdagPjpTYKm7yN9sqHGhA?both

## Summary 

- Introduced the performance sidecar stack: `performance_collector` (collects/writes data) + `performance_web_server` (HTTP API), with Prometheus metrics and healthchecks.
- Added a reference `docker-compose.yml` for the Staking Module stack (Postgres + sidecars + `csm`/`cm`) and DB init script `scripts/init_performance_db.sh`.
- Added monitoring examples under `docs/monitoring/*` (Prometheus/Alertmanager configs + alert rules for performance sidecars and CSM/CM).
- Refactored module structure: extracted common runtime into `src/modules/oracles/common/*`, updated entrypoints/import paths.
- Generalized “CSM” into “staking module(s)”: added Curated Module oracle (`cm`) alongside Community Staking (`csm`).
- Updated checks and fork tests: switched to `STAKING_MODULE_ADDRESS`; added `CS_MODULE_ADDRESS` / `CURATED_MODULE_ADDRESS` for checks/fork runs; checks now skip on incompatible staking module contract versions.

## Related Issue/Task
- Related task: [CS-873](https://linear.app/lidofi/issue/CS-873/separate-collect-and-distribute-steps-in-oracle)

## How Has This Been Tested?
- [X] Local tests
- [X] Manual testing

## Checklist
- [X] Documentation updated (if required)


.github/workflows/mainnet_fork_tests.yml
```diff
@@ -82,7 +55,8 @@ jobs:
           CONSENSUS_CLIENT_URI: ${{ secrets.CONSENSUS_CLIENT_URI }}
           KEYS_API_URI: ${{ secrets.KEYS_API_URI }}
           LIDO_LOCATOR_ADDRESS: "0xC1d0b3DE6792Bf6b4b37EccdcC24e45978Cfd2Eb"
-          CSM_MODULE_ADDRESS: "0xdA7dE2ECdDfccC6c3AF10108Db212ACBBf9EA83F"
+          CS_MODULE_ADDRESS: "0xdA7dE2ECdDfccC6c3AF10108Db212ACBBf9EA83F"
+          # CURATED_MODULE_ADDRESS: "0xdA7dE2ECdDfccC6c3AF10108Db212ACBBf9EA83F" # TODO: replace with actual address
```
chasingrainbows, line 82: Actual address is avaiable already?

docs/monitoring/README.md
```diff
@@ -0,0 +1,19 @@
...
+- CSM oracle (`csm-oracle:9000/metrics`)
+- CM oracle (`cm-oracle:9000/metrics`)
+
+## Files
+
+- `docs/monitoring/prometheus.yml` - scrape jobs + alerting + rule files wiring
+- `docs/monitoring/alerts.yml` - alert rules for the services above (includes epoch/DB health for Performance Collector and perf-oracles)
+- `docs/monitoring/alertmanager.yml` - routing/receivers skeleton (plug in Slack/OpsGenie/etc.)
+
+## Notes
```
chasingrainbows, line 24: Missed instructions how to use this configs (copy to project root, run commands, etc)

scripts/init_performance_db.sh
```diff
@@ -0,0 +1,100 @@
...
+# Admin connection settings
+DB_HOST="${PERFORMANCE_DB_HOST:-localhost}"
+DB_PORT="${PERFORMANCE_DB_PORT:-5432}"
+ADMIN_USER="${PERFORMANCE_DB_ADMIN_USER:-postgres}"
+ADMIN_PASSWORD="${PERFORMANCE_DB_ADMIN_PASSWORD}"
+
+# New user/database settings
+DB_NAME="${PERFORMANCE_DB_NAME:-performance}"
+DB_USER="${PERFORMANCE_DB_USER:-performance}"
+DB_PASSWORD="${PERFORMANCE_DB_PASSWORD:-performance}"
```
chasingrainbows: Better to drop error if PERFORMANCE_DB_USER/PERFORMANCE_DB_PASSWORD is empty. Default password "performance" could bring security risks.

src/modules/common/daemon_module.py
```diff
@@ -0,0 +1,104 @@
...
+    Base class for daemon-like modules.
+
+    Provides common functionality for:
+    - Running in daemon mode
+    - Cycle handling
+    - Getting last finalized slot
+    - Timeout management
+    """
+
+    _slot_threshold: SlotNumber = SlotNumber(0)
```
chasingrainbows: Better to move into init, if there is no intention to share this value between all sub-class objects.

src/modules/oracles/common/consensus.py
```diff
@@ -483,7 +483,7 @@ def _get_web3_converter(self, blockstamp: BlockStamp) -> Web3Converter:
     @lru_cache(maxsize=1)
     def get_frame_number_by_slot(self, blockstamp: ReferenceBlockStamp) -> FrameNumber:
         converter = self._get_web3_converter(blockstamp)
-        frame_number = converter.get_frame_by_slot(blockstamp.ref_slot)
+        frame_number = converter.get_frame_by_slot(SlotNumber(blockstamp.ref_slot + 1))
```
chasingrainbows, line 490: Why `+ 1`?

src/modules/oracles/common/oracle_module.py
```diff
@@ -114,35 +86,8 @@ def _cycle(self):
...
-        ORACLE_BLOCK_NUMBER.labels('finalized').set(bs.block_number)
-        return bs
-
-    def run_cycle(self, blockstamp: BlockStamp):
-        logger.info({'msg': 'Execute module.', 'value': blockstamp})
-        result = self.execute_module(blockstamp)
-        pulse()
-        if result is ModuleExecuteDelay.NEXT_FINALIZED_EPOCH:
-            self._slot_threshold = blockstamp.slot_number
+        except Exception as error:
```
chasingrainbows: except Exception to raise it looks senseless

src/modules/oracles/common/runtime.py
```diff
@@ -0,0 +1,164 @@
...
+        modules['lido_validators'] = LidoValidatorsProvider
+
+    if config.use_staking_module_contracts:
+        modules['staking_module'] = StakingModuleContracts
+
+    if config.use_ipfs:
+        ipfs = IPFS(web3, ipfs_providers(), retries=variables.HTTP_REQUEST_RETRY_COUNT_IPFS)
+        modules['ipfs'] = lambda: ipfs
+
+    if config.use_performance_client:
```
chasingrainbows:
```suggestion
    if config.use_performance_client and variables.PERFORMANCE_COLLECTOR_URI:
```
Also, maybe we could add some validation of PERFORMANCE_COLLECTOR_URI

src/modules/oracles/staking_modules/base.py
```diff
@@ -0,0 +1,430 @@
...
+
+        # XXX: We put a stone here to make sure, that even with only 1 node operator in the tree, it's still possible to
+        # claim rewards. The CSModule contract skips pulling rewards if the proof's length is zero, which is the case
+        # when the tree has only one leaf.
+        stone = NodeOperatorId(self.w3.staking_module.module.MAX_OPERATORS_COUNT)
+        shares[stone] = 0
+
+        # XXX: Remove the stone as soon as we have enough leafs to build a suitable tree.
+        if stone in shares and len(shares) > 2:
+            shares.pop(stone)
```
chasingrainbows: Better to return shares, and reassign them. Or remove shares modification from this method, to not break the logic flow. Implicit modification of objects worsens readability imo.

src/modules/oracles/staking_modules/base.py
```diff
@@ -0,0 +1,430 @@
...
+            raise ValueError("No shares to build a tree")
+
+        # XXX: We put a stone here to make sure, that even with only 1 node operator in the tree, it's still possible to
+        # claim rewards. The CSModule contract skips pulling rewards if the proof's length is zero, which is the case
+        # when the tree has only one leaf.
+        stone = NodeOperatorId(self.w3.staking_module.module.MAX_OPERATORS_COUNT)
+        shares[stone] = 0
+
+        # XXX: Remove the stone as soon as we have enough leafs to build a suitable tree.
+        if stone in shares and len(shares) > 2:
```
chasingrainbows: Looks likea `stone in shares` always will be True becasue of `shares[stone] = 0`

src/modules/oracles/staking_modules/base.py
```diff
@@ -0,0 +1,430 @@
...
+        last_ref_slot = self.w3.staking_module.get_last_processing_ref_slot(blockstamp)
+        ref_slot = self.get_initial_or_current_frame(blockstamp).ref_slot
+        return last_ref_slot == ref_slot
+
+    def is_contract_reportable(self, blockstamp: BlockStamp) -> bool:
+        return not self.is_main_data_submitted(blockstamp)
+
+    def is_reporting_allowed(self, blockstamp: ReferenceBlockStamp) -> bool:
+        on_pause = self.report_contract.is_paused('latest')
+        CONTRACT_ON_PAUSE.labels("csm").set(on_pause)
```
chasingrainbows: maybe

```suggestion
        CONTRACT_ON_PAUSE.labels(self.consumer).set(on_pause)
```

src/modules/oracles/staking_modules/base.py
```diff
@@ -0,0 +1,430 @@
...
+            })
+        except (ConnectionError, TimeoutError, OSError) as error:
+            logger.warning({
+                "msg": "Unexpected error during Performance Collector demand cleanup",
+                "consumer": self.consumer,
+                "error": str(error),
+            })
+
+    def execute_module(self, last_finalized_blockstamp: BlockStamp) -> ModuleExecuteDelay:
+        if not self._check_compatability(last_finalized_blockstamp):
```
chasingrainbows: typo in class method name

```suggestion
        if not self._check_compatibility(last_finalized_blockstamp):
```

src/modules/oracles/staking_modules/base.py
```diff
@@ -0,0 +1,430 @@
...
+
+        collected = self.collect_data()
+        if not collected:
+            return ModuleExecuteDelay.NEXT_FINALIZED_EPOCH
+
+        self.process_report(report_blockstamp)
+        return ModuleExecuteDelay.NEXT_SLOT
+
+    @duration_meter()
+    def set_epochs_range_to_collect(self, blockstamp: BlockStamp):
```
chasingrainbows: Looks like a private method 

```suggestion
    def _set_epochs_range_to_collect(self, blockstamp: BlockStamp):
```

src/modules/oracles/staking_modules/base.py
```diff
@@ -0,0 +1,430 @@
...
+        if current_epochs_range != (l_epoch, r_epoch):
+            logger.info({
+                "msg": f"Updating {self.consumer} epochs demand for Performance Collector",
+                "old": current_epochs_range,
+                "new": (l_epoch, r_epoch)
+            })
+            self.w3.performance.post_epochs_demand(self.consumer, l_epoch, r_epoch)
+
+    @duration_meter()
+    def collect_data(self) -> bool:
```
chasingrainbows: Looks like a private method 

```suggestion
    def _collect_data(self) -> bool:
```

src/modules/oracles/staking_modules/base.py
```diff
@@ -0,0 +1,430 @@
...
+        return tree
+
+    def make_strikes_tree(self, strikes: dict[StrikesValidator, StrikesList]) -> StrikesTree:
+        if not strikes:
+            raise ValueError("No strikes to build a tree")
+        tree = StrikesTree.new(tuple((no_id, pubkey, strikes) for ((no_id, pubkey), strikes) in strikes.items()))
+        logger.info({"msg": "New strikes tree built for the report", "root": repr(tree.root)})
+        return tree
+
+    def publish_tree(self, tree: Tree) -> CID:
```
chasingrainbows: There is no reason to add methods such as publish_tree/publish_log around self.w3.ipfs. It is just adds some logs that already exists inside self.w3.ipfs.

src/modules/oracles/staking_modules/base.py
chasingrainbows: Well there is a lot such methods in file

src/modules/sidecars/performance/collector/checkpoint.py
```diff
@@ -259,100 +267,106 @@ def _check_duties(
...
+        )
+        self._maybe_refresh_db_metrics()
+
+    def _maybe_refresh_db_metrics(self, interval_seconds: float = 30.0) -> None:
+        with self._metrics_lock:
+            now = time.time()
+            if now - self._last_metrics_refresh < interval_seconds:
+                return
+            self._last_metrics_refresh = now
+        self._refresh_db_metrics()
```
chasingrainbows: Why `_refresh_db_metrics` call out of lock?

src/modules/sidecars/performance/collector/collector.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    Continuously collects performance data from Consensus Layer into db for the given epoch range.
+    """
+    _slot_threshold = SlotNumber(0)
+
+    # Timestamp of the last epochs demand update
+    last_epochs_demand_update: int | None = None
+
+    def __init__(self, cc: ConsensusClient):
+        self.cc = cc.cc if hasattr(cc, "cc") else cc
```
chasingrainbows: Well, how is it possible ConsensusClient obj has not cc attr?

src/modules/sidecars/performance/collector/collector.py
```diff
@@ -0,0 +1,212 @@
...
+        self.db = DutiesDB(
+            connect_timeout=variables.PERFORMANCE_COLLECTOR_DB_CONNECTION_TIMEOUT,
+            statement_timeout_ms=variables.PERFORMANCE_COLLECTOR_DB_STATEMENT_TIMEOUT_MS,
+        )
+        self.last_epochs_demand_update = self.get_epochs_demand_max_updated_at()
+
+    def _get_consensus_client(self):
```
chasingrainbows: Im not sure we need separate method with "Any" type as return, for this, in base class. There is only one ConsensusClient in project, and we can use it as type.

src/modules/sidecars/performance/collector/collector.py
```diff
@@ -0,0 +1,212 @@
...
+from src.utils.web3converter import ChainConverter
+
+logger = logging.getLogger(__name__)
+
+
+class PerformanceCollector(DaemonModule):
+    """
+    Continuously collects performance data from Consensus Layer into db for the given epoch range.
+    """
+    _slot_threshold = SlotNumber(0)
```
chasingrainbows: Already exists at base class

src/modules/sidecars/performance/common/db.py
```diff
@@ -0,0 +1,215 @@
...
+                demand.r_epoch = r_epoch
+                demand.updated_at = int(time())
+            else:
+                demand = EpochsDemand(consumer=consumer, l_epoch=l_epoch, r_epoch=r_epoch, updated_at=int(time()))
+                session.add(demand)
+            session.commit()
+
+    def delete_demand(self, consumer: str) -> None:
+        with self.get_session() as session:
+            demand = session.get(EpochsDemand, consumer)
```
chasingrainbows: You dont need to do sql select before delete, you can just delete then supress exception if record does not exits

src/modules/sidecars/performance/common/db.py
```diff
@@ -0,0 +1,215 @@
...
+        if connect_timeout:
+            connect_args["connect_timeout"] = connect_timeout
+        if self._statement_timeout_ms is not None:
+            connect_args["options"] = f"-c statement_timeout={self._statement_timeout_ms}"
+
+        return create_engine(
+            self._get_database_url(),
+            echo=False,
+            pool_pre_ping=True,  # Enable connection health checks
+            pool_recycle=3600,  # Recycle connections every hour
```
chasingrainbows: Better to add such settings to envs as well

src/modules/sidecars/performance/common/db.py
```diff
@@ -0,0 +1,215 @@
...
+        self,
+        *,
+        connect_timeout: int | None = None,
+        statement_timeout_ms: int | None = None,
+    ):
+        self._statement_timeout_ms = statement_timeout_ms
+        self.engine = self._build_engine(connect_timeout)
+        self._setup_database()
+
+    def _build_engine(self, connect_timeout: int | None):
```
chasingrainbows: Return type?

src/modules/sidecars/performance/common/db.py
```diff
@@ -0,0 +1,215 @@
...
+    def _get_database_url() -> str:
+        """Get PostgreSQL database URL from environment variables"""
+        host = variables.PERFORMANCE_DB_HOST
+        port = variables.PERFORMANCE_DB_PORT
+        name = variables.PERFORMANCE_DB_NAME
+        user = variables.PERFORMANCE_DB_USER
+        password = variables.PERFORMANCE_DB_PASSWORD
+        return f"postgresql://{user}:{password}@{host}:{port}/{name}"
+
+    def _setup_database(self):
```
chasingrainbows: Return type?

src/modules/sidecars/performance/common/db.py
```diff
@@ -0,0 +1,215 @@
...
+                session.commit()
+
+    def store_epoch(
+        self,
+        epoch: EpochNumber,
+        att_misses: AttDutyMisses,
+        proposals: list[ProposalDuty],
+        syncs: list[SyncDuty],
+    ) -> None:
+        # TODO: test that store and get are consistent
```
chasingrainbows, line 110: Todo?

.github/workflows/mainnet_fork_tests.yml
vgorkavenko: Not yet

src/modules/sidecars/performance/collector/collector.py
vgorkavenko: It isn't possible. Weird change

src/modules/sidecars/performance/web/server.py
```diff
@@ -0,0 +1,185 @@
...
+def query_epoch_range(
+    from_epoch: EpochNumber = Query(..., alias="from"),
+    to_epoch: EpochNumber = Query(..., alias="to"),
+) -> tuple[EpochNumber, EpochNumber]:
+    validate_epoch_bounds(from_epoch, to_epoch)
+    return from_epoch, to_epoch
+
+
+@app.get("/health", response_model=HealthCheckResp)
+def health():
```
chasingrainbows: Should we check postgress connection here at least?

src/modules/sidecars/performance/web/server.py
```diff
@@ -0,0 +1,185 @@
...
+    logger.info("Response: %s | Status: %s", url_info, response.status_code)
+
+    return response
+
+
+def get_db() -> DutiesDB:
+    return cast(DutiesDB, app.state.db)
+
+
+def validate_epoch_bounds(l_epoch: EpochNumber, r_epoch: EpochNumber) -> None:
```
chasingrainbows: fastapi way to validate things is pydantic

For example
```
  from pydantic import BaseModel, Field, field_validator
  class EpochsDemandRequest(BaseModel):
      consumer: str = Field(..., min_length=1, max_length=255)
      l_epoch: EpochNumber
      r_epoch: EpochNumber

      @field_validator('consumer')
      @classmethod
      def consumer_not_blank(cls, v: str) -> str:
          if not v.strip():
              raise ValueError('consumer cannot be blank')
          return v
```

src/modules/sidecars/performance/web/server.py
```diff
@@ -0,0 +1,185 @@
...
+
+@app.get("/demands/{consumer}", response_model=EpochsDemand | None)
+def one_epochs_demand(consumer: str, db: DutiesDB = Depends(get_db)):
+    return db.get_epochs_demand(consumer)
+
+
+@app.post("/demands", response_model=EpochsDemand)
+def set_epochs_demand(demand_to_add: EpochsDemandRequest, db: DutiesDB = Depends(get_db)):
+    validate_epoch_bounds(demand_to_add.l_epoch, demand_to_add.r_epoch)
+    db.store_demand(demand_to_add.consumer, demand_to_add.l_epoch, demand_to_add.r_epoch)
```
chasingrainbows: Is it possible that between db.store_demand call and return db.get_epochs_demand, delete_epochs_demand view will be called?

chasingrainbows: i mean looks like store_demand could return demand, to not call sql select get_epochs_demand here

src/modules/sidecars/performance/web/server.py
```diff
@@ -0,0 +1,185 @@
...
+    l_epoch, r_epoch = epoch_range
+    return db.is_range_available(l_epoch, r_epoch)
+
+
+@app.get("/missing-epochs", response_model=list[EpochNumber])
+def epochs_missing(
+        epoch_range: tuple[EpochNumber, EpochNumber] = Depends(query_epoch_range),
+        db: DutiesDB = Depends(get_db),
+):
+    l_epoch, r_epoch = epoch_range
```
chasingrainbows: Validation missed

src/modules/sidecars/performance/web/server.py
```diff
@@ -0,0 +1,185 @@
...
+        )
+
+
+def query_epoch_range(
+    from_epoch: EpochNumber = Query(..., alias="from"),
+    to_epoch: EpochNumber = Query(..., alias="to"),
+) -> tuple[EpochNumber, EpochNumber]:
+    validate_epoch_bounds(from_epoch, to_epoch)
+    return from_epoch, to_epoch
+
```
chasingrainbows: Not found tests for this http endpoints (with direct http requests, responses asserting etc)

src/providers/performance/client.py
```diff
@@ -0,0 +1,51 @@
...
+    NotOkResponse,
+)
+from src.types import EpochNumber
+
+
+class PerformanceClientError(NotOkResponse):
+    pass
+
+
+class PerformanceClient(HTTPProvider):
```
chasingrainbows, line 15: Any tests for this one?

src/modules/sidecars/performance/web/server.py
vgorkavenko: No tests at all for that yet

src/providers/performance/client.py
vgorkavenko: No tests at all for that yet. Our plan is to add them during testnet

src/main.py
```diff
@@ -1,191 +1,57 @@
...
-logger = logging.getLogger(__name__)
 
+def main(module: OracleModule):
+    # pylint: disable=import-outside-toplevel,too-many-return-statements
+
+    if module is OracleModule.CHECK:
+        errors = variables.check_uri_required_variables()
+        variables.raise_from_errors(errors)
+        from src.modules.checks import entrypoint as check_entrypoint
+        return sys.exit(check_entrypoint.run())
```
chasingrainbows: Return cant be reached if u call sys.exit, if i remember it right

src/main.py
```diff
@@ -1,191 +1,57 @@
...
 
-logger = logging.getLogger(__name__)
 
+def main(module: OracleModule):
+    # pylint: disable=import-outside-toplevel,too-many-return-statements
+
+    if module is OracleModule.CHECK:
+        errors = variables.check_uri_required_variables()
+        variables.raise_from_errors(errors)
+        from src.modules.checks import entrypoint as check_entrypoint
```
chasingrainbows, line 21: What the reason to have separate `entrypoint` module for each oracle module? Looks like In most cases it just duplicates code without need. Code will be much cleaner if we just left such code in main.

(Also funny - It is break search inside my PyCharm - all function has the same name for "run" function and module, IDE cant recognize it correctly)

docs/development.md
```diff
@@ -9,8 +9,15 @@ Using Docker allows you to develop in an environment that closely mirrors produc
...
-2. Run the command `make up` to build the image and start the container. (`make` is likely already installed on your system as a standard build tool on Unix-like environments. If not, you can [download it here](https://www.gnu.org/software/make/).)  
-3. Run `make install-pre-commit` to install pre-commit hooks. The hook will be installed on your host machine, but tests and linting will run inside the container.
+2. Prepare environment file:
+
+```bash
+cp .env.example .env
+# edit .env with your endpoints / addresses
+```
+
+3. Run the command `make up` to build the image and start the container. (`make` is likely already installed on your system as a standard build tool on Unix-like environments. If not, you can [download it here](https://www.gnu.org/software/make/).)
```
chasingrainbows, line 19: Need to add command to Makefile to up this via docker-compose with postgres, etc

tests/conftest.py
```diff
@@ -9,11 +9,10 @@
 from web3 import EthereumTesterProvider
```
chasingrainbows, line 10: Need to add ability to write tests with usage of test DB locally (via docker + make test).

src/constants.py
```diff
@@ -33,6 +32,7 @@
 # https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#time-parameters
 SLOTS_PER_HISTORICAL_ROOT = 2**13  # 8192
 # https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#sync-committee
```
hweawer, line 36:
```suggestion
# https://github.com/ethereum/consensus-specs/blob/master/specs/altair/beacon-chain.md#sync-committee
```

src/modules/sidecars/performance/collector/checkpoint.py
```diff
@@ -133,38 +113,57 @@ def __setitem__(self, sync_committee_period: int, value: SyncCommittee):
 
 class FrameCheckpointProcessor:
```
hweawer, line 114: Tests like `test_checkpoints_processor_get_block_roots` now should have a different params names

src/modules/sidecars/performance/common/db.py
```diff
@@ -0,0 +1,222 @@
...
+from src.modules.sidecars.performance.common.types import ProposalDuty, SyncDuty, AttDutyMisses
+from src.types import EpochNumber
+from src.utils.range import sequence
+
+
+class Duty(SQLModel, table=True):
+    __tablename__: ClassVar[str] = "duties"
+
+    epoch: int = Field(primary_key=True)
+    attestations: list[int] = Field(default=None, sa_column=Column(ARRAY(Integer())))
```
hweawer: consider empty arrays as default value

src/main.py
vgorkavenko: They are actually different 🤔 
The main reason was to hide internal things and make `main.py` as clean as possible.

src/constants.py
vgorkavenko: I would suggest to do that (if even need) in the separate PR. Definitely out of scope.

tests/conftest.py
vgorkavenko: Not sure that it is worth to up a real Postgres. In-memory sqlite would work for testing purposes.

vgorkavenko: We also need to determine what kind of tests we expect to write using a real database (even in-memory)?

chasingrainbows: > Not sure that it is worth to up a real Postgres. In-memory sqlite would work for testing purposes.

There is no reason to use sqlite locally, because with docker there is a literally lower overhead for this. Dev containers is primary way to develop oracle. Sqlite helps only for local setup, which is less secure and can also work with postgress container anyway.

Looks for me like less work (no need to add sqlite crutches to codebase) + more consistent solution with production environment (postgres container can be reused both for tests and local dev setup).

chasingrainbows: > We also need to determine what kind of tests we expect to write using a real database (even in-memory)?

integration tests for api endpoints + db layer methods

src/main.py
chasingrainbows: > clean as possible

IMO, now it is harder to read, too many jumps with almost the same objects and calls inside. I don't insist here, you can ask guys opinion @hweawer @F4ever as well.

Issue #836, opened 2025-12-30
Oracle fee underpayment (stVaults) due to block-number–based time accounting when CL slots are missed


TheDZhon:
### Summary

The Lido oracle currently derives elapsed time for fee and APR calculations using **execution-layer block numbers**, implicitly assuming a linear mapping between block numbers and real elapsed time.

This assumption breaks when **Consensus Layer (CL) slots are missed** (i.e., no execution payload is produced). In such cases, real time passes but `block.number` does not advance, causing **systematic undercounting of elapsed time** and **underpayment of vault fees**.

Based on observed missed-slot rates, the underpayment is estimated at approximately **~0.5%** for stVaults fees.

---

### Root Cause

When a CL slot is missed:

* No execution-layer block is produced
* The chain head does not advance
* `block.number` remains unchanged
* Real wall-clock time continues to elapse

However, multiple oracle code paths compute elapsed time as:

```
blocks_elapsed / BLOCKS_PER_YEAR
```

instead of using actual elapsed time in seconds.

As a result:

* Missed slots are not accounted for
* Elapsed time is underestimated
* Cumulative staking vault fees are undercharged

---

### Impact

* Vault fees are **systematically underpaid**
* The discrepancy accumulates over time
* The magnitude of underpayment scales with validator missed-slot rate
* Using public timing dashboards, the average missed-slot ratio for top validators is ~0.5%, leading to a comparable fee shortfall

This is an **economic correctness issue**, not a safety or liveness failure, but it affects long-term protocol accounting accuracy.

---

### Suggested Mitigation (Non-Prescriptive)

* Make fee and APR calculations **seconds-based**, not block-based
* Replace `block_elapsed / BLOCKS_PER_YEAR` with:

  * `seconds_elapsed / SECONDS_IN_YEAR`
* Derive elapsed time from:

  * Execution-layer block timestamps
  * CL timestamps where appropriate
* Ensure missed CL slots are explicitly accounted for in oracle logic

---

### Honorable Mention

🙏 **Honorable mention to the whitehat researcher** who identified, documented, and provided a clear proof-of-concept for this issue, including concrete test cases and quantitative impact analysis.

TheDZhon:
https://github.com/lidofinance/lido-oracle/tree/fix/stvaults-fee-timestamps

PR #838, opened 2026-01-07
feat: use slots and timestamps for fees

tamtamchik:
Fix for issue #836


src/services/staking_vaults.py
```diff
@@ -495,39 +496,94 @@ def is_tree_root_valid(self, expected_tree_root: str, merkle_tree: StakingVaultI
...
+    @staticmethod
+    def _calculate_liquidity_fee_by_events(
         vault_address: str,
         liability_shares: Shares,
         liquidity_fee_bp: int,
-        events: defaultdict[str, list[VaultEventType]],
-        prev_block_number: BlockNumber,
-        current_block: BlockNumber,
+        vault_events: list[VaultEventType],
+        prev_report_timestamp: int,
```
folkyatina:
```suggestion
        prev_refSlot_timestamp: int,
```

src/services/staking_vaults.py
```diff
@@ -495,39 +496,94 @@ def is_tree_root_valid(self, expected_tree_root: str, merkle_tree: StakingVaultI
...
+        current_report_timestamp: int,
```
folkyatina:
```suggestion
        current_refSlot_timestamp: int,
```

src/services/staking_vaults.py
```diff
@@ -551,60 +607,77 @@ def calc_liquidity_fee(
...
+        vault_events.sort(
+            key=lambda event: (
+                StakingVaultsService._get_event_effective_timestamp(
+                    event,
+                    vault_address,
+                    block_timestamps,
+                    seconds_per_slot,
+                ),
+                event.log_index,
+            ),
```
folkyatina: _get_event_effective_timestamp will push some events from one bucket to the next one where they will mix with other events from this block depending on log_index, but breaking the order, so we need to add block_number to the tuple also. 

```suggestion
                event.block_number,
                StakingVaultsService._get_event_effective_timestamp(
                    event,
                    vault_address,
                    block_timestamps,
                    seconds_per_slot,
                ),
                event.log_index,
            ),
```

folkyatina: Looks like it won't work anyway, because continuous mint/burn/mint/burn in one block will be broken anyway. As it will be translated to mint/mint/burn/burn. 

We need to use effective timestamp only to calculate elapsed time to calculate fees but it should not change the order of events.

src/services/staking_vaults.py
```diff
@@ -551,60 +607,77 @@ def calc_liquidity_fee(
...
 
-                current_block = event.block_number
+            current_report_timestamp = event_timestamp
 
-            blocks_elapsed_between_events = current_block - prev_block_number
-            minted_steth_on_event = get_steth_by_shares(liability_shares, pre_total_pooled_ether, pre_total_shares)
-            vault_liquidity_fee += StakingVaultsService.calc_fee_value(
-                minted_steth_on_event, blocks_elapsed_between_events, core_apr_ratio, liquidity_fee
-            )
+        time_elapsed_seconds = current_report_timestamp - prev_report_timestamp
```
folkyatina: Do we charge for the whole period here? 0_0 But before we charged for the time between events only.

src/services/staking_vaults.py
```diff
@@ -551,60 +607,77 @@ def calc_liquidity_fee(
...
-                elif isinstance(event, BadDebtSocializedEvent):
-                    if vault_address == event.vault_donor:
-                        liability_shares += event.bad_debt_shares
-                    else:
-                        liability_shares -= event.bad_debt_shares
+                else:
+                    liability_shares -= event.bad_debt_shares
 
-                current_block = event.block_number
+            current_report_timestamp = event_timestamp
```
folkyatina: let's have a separate variable for the counter with a distinct name

folkyatina (commented):
Some minor naming nitpicks and possible issue with event ordering.

src/services/staking_vaults.py
tamtamchik: This is `report_interval_seconds` (I'll rename the variable) and is used for fee calculations for the whole report time (like infrastructure and reservation and liquidity in case we do not have events)

docs/testing.md
```diff
@@ -19,3 +19,52 @@ To run tests with a coverage report, run `pytest --cov=src tests` in the root di
...
+
+Mutation testing is used to evaluate the quality of the test suite by introducing small changes (mutations) to the source code and verifying that tests catch these changes.
+
+### Running Mutation Tests
+
+Run mutation tests on specific files or directories:
+
+```bash
+# Run on a specific mutation target and specific test files
+poetry run mutmut run \
```
chasingrainbows, line 33: You also could add it to Makefile, to run in inside docker easly. Something like
```
test-mutations: up
	$(EXEC_CMD) poetry run mutmut run --paths-to-mutate$${MUTATION_TEST_PATH} --runner="pytest -x -m unit" && poetry run mutmut results
```

src/utils/block.py
```diff
@@ -0,0 +1,232 @@
...
+    num_anchors: int,
+) -> None:
+    """Fill timestamps using anchor point strategy for large ranges."""
+    anchor_indices = _select_anchor_indices(len(blocks), num_anchors)
+    anchor_blocks = [blocks[i] for i in anchor_indices]
+
+    blocks_to_fetch = [b for b in anchor_blocks if b not in timestamps]
+    if blocks_to_fetch:
+        fetched = _batch_fetch_timestamps(w3, blocks_to_fetch)
+        for block, ts in zip(blocks_to_fetch, fetched):
```
chasingrainbows: Consider to add strict=True to zip (https://docs.python.org/3/library/functions.html#zip, search "strict"). Without the strict=True, any bug that results in iterables of different lengths will be silenced. With strict - exception will be raised.

src/utils/block.py
```diff
@@ -0,0 +1,232 @@
...
+        block_data = w3.eth.get_block(block)
+        timestamps[block] = int(block_data["timestamp"])
+
+
+def _fetch_endpoints(
+    w3: Web3, blocks: list[BlockNumber], timestamps: dict[BlockNumber, int]
+) -> None:
+    """Fetch first and last block timestamps if not cached."""
+    first_block, last_block = blocks[0], blocks[-1]
+    if first_block not in timestamps:
```
chasingrainbows: Hmm, looks like there is possible to fetch both first and last block by single request instead of two, by reuse of _batch_fetch_timestamps.

src/utils/block.py
```diff
@@ -0,0 +1,232 @@
...
+
+    return indices
+
+
+def _is_batch_available(w3: Web3) -> bool:
+    """Check if Web3 batch requests are available (not a mock)."""
+    batch_requests = getattr(w3, "batch_requests", None)
+    if batch_requests is None:
+        return False
+    return not hasattr(batch_requests, "_mock_name")
```
chasingrainbows: IMO production code should not contain crutches specific for tests purposes only. How it could be resolved from my point of view:
1. Consider to use this one mock of w3 - tests/conftest.py -> web3 in your unit tests. It is based on eth_tester, and mocks w3 object better, follows w3 object interface fully.
2. w3 mock from p.1 able to rise `Web3TypeError: Batch requests are not supported by this provider.`, because it follows real w3 object interface. So, you basically need to replace this line with exception checking, or even remove it completely.

tests/modules/accounting/staking_vault/test_block_timestamps_integration.py
```diff
@@ -0,0 +1,409 @@
...
+from web3 import Web3
+
+from src.utils.block import get_block_timestamps
+
+# Ethereum mainnet seconds per slot (post-merge)
+SECONDS_PER_SLOT = 12
+
+
+@pytest.mark.integration
+@pytest.mark.skip("This is a long running development test, not for regular CI runs.")
```
chasingrainbows: Tests that cant be run at CI should be separate scripts outside of repo IMO. There is no reason for them, if we cant run them in CI. They will die quickly because nobody will support them.

chasingrainbows: Maybe this test could be checker in src/modules/checks/suites? How long it runs?

chasingrainbows: Rewrite it as unit test also an option imo

tests/modules/accounting/staking_vault/test_fee_calculation.py
```diff
@@ -0,0 +1,100 @@
...
+import pytest
+
+from src.services.staking_vaults import StakingVaultsService
+from tests.modules.accounting.staking_vault.conftest import FeeTestConstants
+
+
+class TestCalcFeeValue:
+    """Tests for calc_fee_value static method."""
+
+    @pytest.mark.unit
```
chasingrainbows: U can mark whole class with  @pytest.mark.unit, to not mark each test separately

tests/modules/accounting/staking_vault/test_fee_get_vaults.py
```diff
@@ -0,0 +1,649 @@
...
+
+    @pytest.mark.unit
+    def test_zero_time_elapsed_allowed(self):
+        """Verifies that zero time elapsed between reports is handled correctly without
+        errors. Ensures same-slot reports don't inflate fees or cause calculation failures.
+        """
+        vault_adr = VaultAddresses.VAULT_0
+        vault = VaultInfoFactory.build(vault=vault_adr, liability_shares=0, max_liability_shares=0)
+
+        w3_mock = MagicMock()
```
chasingrainbows: You can use tests/conftest.py -> web3 mock, to skip such code

tests/modules/accounting/staking_vault/test_fee_get_vaults.py
```diff
@@ -0,0 +1,649 @@
...
+            core_apr_ratio=Decimal("0"),
+            pre_total_pooled_ether=Wei(0),
+            pre_total_shares=0,
+            frame_config=FrameConfig(initial_epoch=0, epochs_per_frame=1, fast_lane_length_slots=0),
+            chain_config=ChainConfig(slots_per_epoch=32, seconds_per_slot=12, genesis_time=0),
+            current_frame=FrameNumber(0),
+        )
+
+        call_kwargs = service._calculate_vault_fee_components.call_args.kwargs
+        assert call_kwargs["vault_total_value"] == 0
```
chasingrainbows: I belive we should add some `assert_called_with` for vaults contracts mocks here and below

pyproject.toml
```diff
@@ -55,6 +55,7 @@ pre-commit = "3.8"
 ipython = "^9.0"
 isort = "^6.0.1"
 conventional-pre-commit = "^4.0.0"
+mutmut = "^2.5.1"
```
chasingrainbows, line 58: Mutatuion testing looks pretty cool! Good idea, thx a lot!

But what about CI? I believe we should add it at least as optional pipeline.

chasingrainbows:
Well done sir! Looks pretty well, i like it! Most important:
1. Consider to add mutation testing to CI
2. `return not hasattr(batch_requests, "_mock_name")` - This looks a bit dirty and could probably be removed. I listed some options above.

Other threads not critical

pyproject.toml
tamtamchik: Will do! Thx!

tests/modules/accounting/staking_vault/test_fee_calculation.py
tamtamchik: Great! Didn't know that! Thanks!

src/services/staking_vaults.py
```diff
@@ -495,35 +497,70 @@ def is_tree_root_valid(self, expected_tree_root: str, merkle_tree: StakingVaultI
...
+    def calc_fee_value(value: Decimal, time_elapsed_seconds: int, core_apr_ratio: Decimal, fee_bp: int) -> Decimal:
+        """Compute fee value over elapsed seconds using core APR and basis points."""
         return (
             value
-            * Decimal(block_elapsed)
+            * Decimal(time_elapsed_seconds)
             * core_apr_ratio
             * Decimal(fee_bp)
-            / Decimal(SLOTS_PER_YEAR * TOTAL_BASIS_POINTS)
+            / Decimal(SECONDS_IN_YEAR * TOTAL_BASIS_POINTS)
```
F4ever, line 507: What if year is leap?

src/services/staking_vaults.py
```diff
@@ -495,35 +497,70 @@ def is_tree_root_valid(self, expected_tree_root: str, merkle_tree: StakingVaultI
...
         )
 
     @staticmethod
-    # pylint: disable=too-many-branches
-    def calc_liquidity_fee(
+    def _apply_event_backward(
```
F4ever: There is no iteration inside - maybe rename to just `apply_event`

src/services/staking_vaults.py
```diff
@@ -542,70 +579,62 @@ def calc_liquidity_fee(
...
-                minted_steth_on_event = get_steth_by_shares(liability_shares, pre_total_pooled_ether, pre_total_shares)
-                vault_liquidity_fee += StakingVaultsService.calc_fee_value(
-                    minted_steth_on_event, blocks_elapsed_between_events, core_apr_ratio, liquidity_fee
+        # Track the boundary from the last processed event/report as we walk backward.
+        prev_event_timestamp = current_ref_slot_timestamp
+
+        # We iterate through events backwards, calculating liquidity fee for each interval based
+        # on the `liability_shares` and the elapsed time between events.
+        # Sort by (block_number, log_index) in reverse order to process events backwards in time.
+        vault_events.sort(key=lambda e: (e.block_number, e.log_index), reverse=True)
```
F4ever, line 596: can somehow event order in one block affect resulted fees?

src/services/staking_vaults.py
```diff
@@ -542,70 +579,62 @@ def calc_liquidity_fee(
...
+
+        interval_seconds = prev_event_timestamp - prev_ref_slot_timestamp
+        minted_steth_on_event = get_steth_by_shares(liability_shares, pre_total_pooled_ether, pre_total_shares)
+        vault_liquidity_fee += StakingVaultsService.calc_fee_value(
+            minted_steth_on_event, interval_seconds, core_apr_ratio, liquidity_fee
+        )
+
         return vault_liquidity_fee, liability_shares
 
     def _get_start_point_for_fee_calculations(
```
F4ever: Seems better split it into two functions?
get_prev_vault_ipfs_report
get_last_processing_slot

src/services/staking_vaults.py
```diff
@@ -638,7 +675,7 @@ def _get_start_point_for_fee_calculations(
             # If any vault-related event occurred in the same block as the previous IPFS report,
             # it has already been included in that report. To avoid overlapping calculations,
             # we shift the starting point by one block forward.
-            return prev_ipfs_report, ref_block.block_number + 1
+            return prev_ipfs_report, SlotNumber(int(last_processing_ref_slot)), ref_block.block_number + 1
```
F4ever: Maybe return here blockstamp as it is.
And next do
```
        events, connected_vaults_set = self._get_vault_events_for_fees(
            vault_hub=vault_hub,
            # Do not incude events from last block of last frame
            from_block=prev_blockstamp.block_number + 1,
            to_block=blockstamp.block_number,
        )
```

F4ever: 3 return parameters and each for different proposes looks confusing

src/utils/block.py
```diff
@@ -0,0 +1,232 @@
...
+    Since missed slots are rare (~1-2% of slots), this typically requires
+    only 1-2 HTTP round-trips for any number of blocks.
+    """
+    if not block_numbers:
+        return {}
+
+    sorted_blocks = sorted(block_numbers)
+    timestamps: dict[BlockNumber, int] = {}
+
+    _fill_timestamps_batched(w3, sorted_blocks, seconds_per_slot, timestamps)
```
F4ever: mutating var in functions is bad practice - maybe just return timestamps?
```
def calculate_timestamps_batched(w3, sorted_blocks, seconds_per_slot) -> list[int]:
```

src/utils/block.py
```diff
@@ -0,0 +1,232 @@
...
+
+def _fill_timestamps_batched(
+    w3: Web3,
+    blocks: list[BlockNumber],
+    seconds_per_slot: int,
+    timestamps: dict[BlockNumber, int],
+    num_anchors: int = 10,
+) -> None:
+    """Fill timestamps using batch RPC with adaptive anchor strategy."""
+    if not blocks:
```
F4ever: we already checked not blocks above.
This is internal/private function, so we are sure it shouldn't be called outside

src/utils/block.py
```diff
@@ -0,0 +1,232 @@
+"""Execution layer block utilities."""
+
+from typing import Any
+
+from eth_typing import BlockNumber
+from web3 import Web3
+
+
+def get_block_timestamps(
```
F4ever, line 20: I propose simpler implementation:

```python
sorted_list_of_blocks = [1,8,25,40, ..., 100]

timestamps = calculate_range(
    sorted_list_of_blocks,
    get_block(blocks[0]).timestamp,
    get_block(blocks[-1]).timestamp,
)

def calculate_range(blocks: list[int], first_block_ts, last_block_ts):
    result = {}

    result[blocks[0]] = first_block_ts
    result[blocks[-1]] = last_block_ts

    if len(blocks) == 2:
        return result
    elif (blocks[-1] - blocks[0]) * 12 == last_block_ts - first_block_ts:
        # no empty blocks
        for block in blocks[1:-1]:
            result[block] = first_block_ts + (block - blocks[0]) * 12
    else:
        median = len(blocks) // 2
        median_block_ts = get_block(blocks[median]).timestamp

        result.update(calculate_range(blocks[:median + 1], first_block_ts, median_block_ts))
        result.update(calculate_range(blocks[median:], median_block_ts, last_block_ts))
    
    return result
```

Could be extended with next block:
```python
elif len(blocks) < FETCH_TRESHOLD:
    result.update(fetch_blocks_timestamps(blocks[1:-1]))

def fetch_blocks_timestamps(blocks):
    with w3.batch_requests() as batch:
        ...
```

src/utils/block.py
```diff
@@ -0,0 +1,232 @@
...
+
+def _sequential_fetch(w3: Web3, blocks: list[BlockNumber]) -> list[int]:
+    """Fetch block timestamps sequentially."""
+    return [int(w3.eth.get_block(block)["timestamp"]) for block in blocks]
+
+
+def _try_batch_fetch(w3: Web3, blocks: list[BlockNumber]) -> list[int]:
+    """Try batch fetch, fall back to sequential on failure."""
+    try:
+        with w3.batch_requests() as batch:
```
hweawer: Providers can limit number of requests per batch example [Alchemy](https://www.alchemy.com/docs/reference/batch-requests#what-is-the-batch-requests-limit-over-http), [Chainnodes](https://www.chainnodes.org/docs/FAQs/rate_limits#batch-request-limits).  Maybe it is worth adding a limit and split single batch into multiple?

src/services/staking_vaults.py
```diff
@@ -542,70 +579,62 @@ def calc_liquidity_fee(
...
+
+            # Because we are iterating backward in time, events must be applied in reverse.
+            # E.g., a burn reduces shares in the future, so going backward we add them back.
+            liability_shares, liquidity_fee = StakingVaultsService._apply_event_backward(
+                event, vault_address, liability_shares, liquidity_fee
             )
 
+            prev_event_timestamp = event_timestamp
+
+        interval_seconds = prev_event_timestamp - prev_ref_slot_timestamp
```
hweawer, line 635: consider adding same safety check after the loop

```
 if interval_seconds < 0:
                raise ValueError(
                    f"Negative event interval for vault {vault_address}. "
                    f"{prev_event_timestamp=} {event_timestamp=}"
                )
```

src/services/staking_vaults.py
tamtamchik: Yes, if fee rate changed in the same block but before/after minting/burning we'll have slightly different fees.

docs/testing.md
tamtamchik: Fixed in 139500585332c0ee87ebddd75289001ceb5ea7fc

src/utils/block.py
tamtamchik: Fixed in b4c61e5

src/services/staking_vaults.py
F4ever: Can we somehow order them? Or it's not possible?

src/utils/block.py
tamtamchik: Fixed in 7923d12

src/utils/block.py
tamtamchik: Fixed in 7923d12

src/utils/block.py
tamtamchik: Fixed in 7923d12

src/utils/block.py
tamtamchik: Fixed in 7923d12

src/utils/block.py
tamtamchik: Fixed in 7923d12 by introducing BATCH_FETCH_MAX

src/services/staking_vaults.py
tamtamchik: This is a good question. I used a regular year, and it looks like a problem. Better use 365.25 days.
Looks like a constant need to be updated from `31536000` ⇒ `31557600`
Fixed in 5ca636f

tests/modules/accounting/staking_vault/test_block_timestamps_integration.py
tamtamchik: Will rewrite to unit test, thank you!

src/services/staking_vaults.py
tamtamchik: Tried to fix in 59ae74d

src/services/staking_vaults.py
tamtamchik: Tried to solve in 59ae74d

tests/modules/accounting/staking_vault/test_fee_get_vaults.py
tamtamchik: Thx! Used this approach!

tests/modules/accounting/staking_vault/test_fee_get_vaults.py
tamtamchik: Replaced with .assert* functions

pyproject.toml
tamtamchik: Not that easy :( Would probably skip for now, as we need to have a separate integration that should cover all the tests

src/utils/block.py
```diff
@@ -0,0 +1,112 @@
...
+    # Check for missed slots: if block N is at time T, block N+k should be at T + k*slot_time
+    expected_last_ts = first_ts + (blocks[-1] - blocks[0]) * seconds_per_slot
+    if expected_last_ts == last_ts:
+        # No missed slots: calculate all timestamps arithmetically
+        base = blocks[0]
+        return [first_ts + (b - base) * seconds_per_slot for b in blocks]
+
+    # Missed slot(s) detected
+    intermediates = blocks[1:-1]
+    if len(intermediates) <= BATCH_FETCH_MAX:
```
tamtamchik: If BATCH_FETCH_MAX is 1 should not use batching and must use regular requests

chasingrainbows (dismissed):

hweawer (dismissed):

src/services/staking_vaults.py
tamtamchik: https://www.quicknode.com/docs/ethereum/eth_getLogs – The integer of the log index position in the block. So we don't need to check transaction here, thus this sorting is correct

F4ever (dismissed):

src/services/staking_vaults.py
```diff
@@ -542,236 +579,311 @@ def calc_liquidity_fee(
...
                 │                 │              │           │
-                └─────────────────┴──────────────┴───────────┴────────▶ block_number (X)
+                └─────────────────┴──────────────┴───────────┴────────▶ time (seconds)
 
                                  mintEvent      burnEvent  current_block
 
                         ◄─────────────────────── processing backwards in time
         """
         vault_liquidity_fee = Decimal(0)
         liquidity_fee = liquidity_fee_bp
```
folkyatina, line 589: Looks like it's the same but it's very different. Maybe `liquidity_fee_rate` or `liquidity_fee_amount` for the first one. `vault` prefix looks ambiguous anyway.

folkyatina (dismissed):

src/utils/block.py
```diff
@@ -0,0 +1,120 @@
...
+    seconds_per_slot: int,
+) -> dict[BlockNumber, int]:
+    """
+    Fetch execution block timestamps for a set of block numbers.
+
+    Uses a hybrid strategy to minimize RPC round-trips:
+      1) Batch-fetch first and last block timestamps.
+      2) If timestamps match expected difference (no missed slots), calculate
+         all intermediate timestamps arithmetically.
+      3) If segment is small (<= BATCH_FETCH_MAX intermediates), batch-fetch all.
```
Psirex: `BATCH_FETCH_MAX` - seems like outdated naming of the variable

src/services/staking_vaults.py
```diff
@@ -542,236 +579,311 @@ def calc_liquidity_fee(
...
+        prev_report_blockstamp = get_blockstamp(
+            cc=self.w3.cc,
+            slot=prev_ref_slot,
+            last_finalized_slot_number=SlotNumber(int(prev_ref_slot) + slots_per_frame),
+        )
+
+        # Events are fetched forward over (event_start_block, current_block] and applied backward by timestamp.
+        events, connected_vaults_set = self._get_vault_events_for_fees(
+            vault_hub=vault_hub,
+            # Do not incude events from last block of last frame
```
Psirex: typo in the "include"

Psirex (approved):
LGTM 👍🏻 Just a couple of minor comments about typos in the comments

src/utils/block.py
F4ever:
```suggestion
      3) If segment is small (<= BLOCK_BATCH_SIZE_LIMIT intermediates), batch-fetch all.
```

src/services/staking_vaults.py
F4ever:
```suggestion
            # Do not include events from last block of last frame
```

F4ever (approved):

Ziars (approved):

PR #849, opened 2026-01-27
Feature/orc 575 oracle metrics and alerts

chasingrainbows:
## Description
  1. init_basic_metrics() — pre-initializes all label combinations for TRANSACTIONS_COUNT, CYCLE_COUNT, and LAST_CYCLE_TIMESTAMP at startup, so metrics exist with value 0 immediately and Prometheus queries work correctly across restarts.
  2. New metrics for oracle health monitoring:
  - lido_oracle_cycle_count{result="success|error"} (Counter) — total completed cycles by result, allows alerting on error rate growth
  - lido_oracle_last_cycle_timestamp{result="success|error"} (Gauge) — unix timestamp of the last cycle, allows alerting on time() - gauge > threshold to detect a stalled oracle
  3. Cycle metrics are recorded in BaseModule._cycle() via a finally block, covering all three modules (accounting, ejector, csm). Skip cycles (waiting for new finalized slot) count as success since the oracle is operating normally.

## Related Issue/Task
- Related task: ORC-575
- Epic: v8

## How Has This Been Tested?

## Checklist

### DIFF
--- a/src/metrics/prometheus/basic.py
+++ b/src/metrics/prometheus/basic.py
@@ -11,6 +11,11 @@ class Status(Enum):
     FAILURE = 'failure'
 
 
+class CycleResult(Enum):
+    SUCCESS = 'success'
+    ERROR = 'error'
+
+
 BUILD_INFO = Info(
     'build',
     'Build info',
@@ -88,3 +93,32 @@ TRANSACTIONS_COUNT = Counter(
     ['status'],
     namespace=PROMETHEUS_PREFIX,
 )
+
+CYCLE_COUNT = Counter(
+    'cycle_count',
+    'Total count of oracle cycles',
+    ['result'],
+    namespace=PROMETHEUS_PREFIX,
+)
+
+LAST_CYCLE_TIMESTAMP = Gauge(
+    'last_cycle_timestamp',
+    'Unix timestamp of the last completed oracle cycle',
+    ['result'],
+    namespace=PROMETHEUS_PREFIX,
+)
+
+
+def init_basic_metrics() -> None:
+    """
+    Initialize metrics with all expected label combinations.
+
+    This ensures metrics exist with value 0 even before first use,
+    preventing issues with Prometheus queries like increase() returning
+    nothing (instead of 0) after service restarts.
+    """
+    for status in Status:
+        TRANSACTIONS_COUNT.labels(status=status.value)
+    for result in CycleResult:
+        CYCLE_COUNT.labels(result=result.value)
+        LAST_CYCLE_TIMESTAMP.labels(result=result.value)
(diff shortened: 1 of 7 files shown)


src/modules/submodules/oracle_module.py
```diff
@@ -62,12 +68,12 @@ def cycle_handler(self):
...
 
     @timeout(variables.MAX_CYCLE_LIFETIME_IN_SECONDS)
-    def _cycle(self):
+    def _cycle(self):  # noqa: C901
         """
         Main cycle logic: fetch the last finalized slot, refresh contracts if necessary,
         and execute the module's business logic.
         """
-        # pylint: disable=too-many-branches
+        cycle_error = True
```
F4ever: maybe just paste `CycleResult.SUCCESS` here?

src/metrics/prometheus/basic.py
```diff
@@ -88,3 +94,32 @@ class Status(Enum):
...
+
+def init_basic_metrics() -> None:
+    """
+    Initialize metrics with all expected label combinations.
+
+    This ensures metrics exist with value 0 even before first use,
+    preventing issues with Prometheus queries like increase() returning
+    nothing (instead of 0) after service restarts.
+    """
+    for status in Status:
```
F4ever, line 122: initialize also `ACCOUNT_BALANCE`

src/metrics/prometheus/basic.py
```diff
@@ -88,3 +94,32 @@ class Status(Enum):
...
+)
+
+
+def init_basic_metrics() -> None:
+    """
+    Initialize metrics with all expected label combinations.
+
+    This ensures metrics exist with value 0 even before first use,
+    preventing issues with Prometheus queries like increase() returning
+    nothing (instead of 0) after service restarts.
```
hweawer: In general it is handled by queries. Like

```
sum(increase(my_counter_total[5m])) OR vector(0)
```

Instrumentation is usually a source of truth and PromQL is a presentation of the data.

src/modules/submodules/oracle_module.py
chasingrainbows: done

src/metrics/prometheus/basic.py
chasingrainbows: done

src/metrics/prometheus/basic.py
chasingrainbows: Holder said, if the service restarts for any reason then the metric is not available, leading to false positives. I will ask him to use `OR vector(0)`, but anyway, init_basic_metrics will make things a little bit easer for them.

F4ever (approved):

hweawer (approved):

PR #876, opened 2026-03-25
feat: cl metrics

hweawer:
# Summary
                                                                                                                                       
  - Integrates HTTPSessionManagerProxy from web3-multi-provider into ConsensusClient to record per-request CL metrics (latency, request counts, CL path labels) via Prometheus                                                                                              
  - Adds a _make_get_request hook to HTTPProvider so the integration is scoped to ConsensusClient only — KAPI and other providers are unaffected                                                                                                                           
  - Updates web3_multi_provider type stubs to expose HTTPSessionManagerProxy and _timed_call
                                                                                                                                       
 # How it works    
                                                                                                                                       
  ConsensusClient._make_get_request routes requests through session_manager._timed_call(self.session.get, url, **kwargs):              
  - Metrics come from HTTPSessionManagerProxy._timed_call — records http_rpc_requests_total, http_rpc_response_seconds, and per-endpoint rpc_request_total with CL path normalization (e.g. eth/v1/beacon/states/{state_id}/...)                                 
  - Retry strategy is preserved — self.session.get (the requests.Session with urllib3.Retry) is passed directly as the callable, not the session manager's own session                                                                                                    
                                                                                                                                       
  ConsensusClientModule (multi-host variant) calls _init_session_manager(hosts[0]) after its own init, using host as the metric label URI.                                                                                                                    
                  
 # Test plan                                                                                                                            
                  
  - test_session_manager_init_params — verifies chain_id=1, network='beacon', layer='cl' labels                                        
  - test_make_get_request_uses_session_manager_timed_call — verifies _timed_call is invoked
  - test_make_get_request_uses_provider_session — verifies self.session.get is the function called (retry strategy preserved)          
  - test_make_get_request_not_overridden_in_kapi — verifies KAPI is unaffected                                                         
  - test_session_manager_metrics_recorded — initializes real Prometheus metrics in a private registry, makes a request, asserts        
  http_rpc_requests_total, http_rpc_response_seconds_count, and rpc_request_total with method="/eth/v1/beacon/genesis" are present in  output


src/providers/consensus/client.py
```diff
@@ -68,6 +69,22 @@ class ConsensusClient(HTTPProvider):
...
 
+    def __init__(self, hosts: list[str], timeout: int, retry_total: int = 3, retry_backoff_factor: int = 3) -> None:
+        super().__init__(
+            hosts, request_timeout=timeout, retry_total=retry_total, retry_backoff_factor=retry_backoff_factor
+        )
+        self._init_session_managers(hosts)
+
+    def _init_session_managers(self, hosts: list[str]) -> None:
+        self._session_managers = {
+            host: HTTPSessionManagerProxy(chain_id=1, uri=host, network='beacon', layer='cl')
```
F4ever: Propagate chain id from el node

hweawer: Done

F4ever (approved):

src/providers/consensus/client.py
```diff
@@ -68,6 +71,23 @@ class ConsensusClient(HTTPProvider):
     API_GET_GENESIS = 'eth/v1/beacon/genesis'
     API_GET_VALIDATOR = 'eth/v1/beacon/states/{}/validators/{}'
 
+    def __init__(
+        self, hosts: list[str], timeout: int, retry_total: int = 3, retry_backoff_factor: int = 3, chain_id: int = 1
```
F4ever, line 75: I think we can fetch chain id directly from CL node. check `get_config_spec`

hweawer: Circular dependency — session managers need chain_id to be created, but fetching chain_id from CL requires session managers. Any solution requires either double-init, a reentrancy flag, or side effects in unexpected places — all worse than the current code.

PR #894, opened 2026-04-13
fix: Include pending topups into cl pending balance

hweawer:
# Summary
                                                                                                                                         
  - Bug fix: _get_cl_pending_validators_balance was only counting pending deposits for new validators (keys not yet activated on the CL).
   Pending top-up deposits for existing active validators were silently omitted.                                                         
   
#  Problem                                                                                                                                
                  
  With Pectra/EIP-7251, additional deposits can be sent to existing validators as top-ups. These land in the CL's pending_deposits queue and are attached to LidoValidator.pending_topups, but _get_cl_pending_validators_balance only iterated over `get_pending_lido_validators`, which filters exclusively for keys not yet in the active validator set.                                   
                  
  This created an accounting gap across frame boundaries: if a top-up deposit was queued in frame N but not yet processed by the CL at frame N+1's reference slot, it would be invisible to the contract's accounting — absent from both `clPendingBalanceAtLastReport` (which stores the oracle's reported pending balance, excluding top-ups) and `depositedForCurrentReport`(which no longer covers it since it was deposited in a prior frame).

#  Fix

  `_get_cl_pending_validators_balance` now sums two components:                                                                            
  1. Pending deposits for new validators awaiting activation (existing behavior)
  2. Pending top-up amounts for existing active validators (validator.pending_topups)                                                    
                                                                                     
  get_active_lido_validators is already @lru_cache'd, so this adds no extra RPC calls.

### DIFF
--- a/src/modules/oracles/accounting/accounting.py
+++ b/src/modules/oracles/accounting/accounting.py
@@ -251,11 +251,23 @@ class Accounting(OracleModule[Web3]):
         return validator_balance_sum
 
     def _get_cl_pending_validators_balance(self, blockstamp: ReferenceBlockStamp) -> Gwei:
-        """Calculate the total balance of all pending Lido validators on the Consensus Layer."""
+        """Calculate the total balance of all pending Lido validators on the Consensus Layer.
+
+        Includes both new validators awaiting activation and pending top-up deposits for
+        existing active validators. Top-ups must be included because they are not yet reflected
+        in validator.balance on the CL; if they remain unprocessed across a frame boundary,
+        they would otherwise be invisible to the contract's accounting (absent from both
+        clPendingBalanceAtLastReport and depositedForCurrentReport).
+        """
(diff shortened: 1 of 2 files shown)


src/modules/oracles/accounting/accounting.py
```diff
@@ -251,11 +251,23 @@ def _get_cl_validators_balance(self, blockstamp: ReferenceBlockStamp) -> Gwei:
         return validator_balance_sum
 
     def _get_cl_pending_validators_balance(self, blockstamp: ReferenceBlockStamp) -> Gwei:
-        """Calculate the total balance of all pending Lido validators on the Consensus Layer."""
+        """Calculate the total balance of all pending Lido validators on the Consensus Layer.
```
F4ever:
```suggestion
        """Calculate the total pending balance on the Consensus Layer.
```

F4ever (approved):

PR #911, opened 2026-04-27
fix: fix abnormal rebase

hweawer:
## Problem                                    
                                                                                                                                    
  `_calculate_cl_rebase_between_blocks` subtracted only newly-appeared validators (counted as 32 ETH each) from the raw CL balance delta. After EIP-7251, Lido can also send **top-up deposits** to already-active validators. These top-ups inflate `raw_cl_rebase` without being subtracted, making the bunker check see an artificially high rebase and potentially fail to trigger when it should.                                                          
                                                                                                                                    
  ## Solution                                                                                                                       
                                                                                                                                    
  Replace `calculate_validators_count_diff_in_gwei` with `_calculate_injected_capital` that handles both cases:
                                                                                                                                    
  **Pre-v4 (before EIP-7251 upgrade):** delegates to the existing validator-count  approach — `(len(ref) - len(prev)) × 32 ETH`. Behaviour is unchanged.
                                                                                                                                    
  **Post-v4:** uses the pending-queue conservation identity to account for all  capital injected (new validators + top-ups) without simulating the deposit queue epoch-by-epoch:                                  
                                                                                                                                    
  injected = deposited_in_window + pending(T1) − pending(T2)                                                                        
                                                                                                                                    
  where `deposited_in_window = depositedForCurrentReport(T2) − depositedForCurrentReport(T1)`. This holds regardless of queue depth or churn rate.
                                                                                                                                    
  The v4 gate is checked at the **last oracle report's block** (not the current frame endpoints). In the first frame after upgrade, `depositedForCurrentReport` is initialised to 0 at upgrade time rather than at the report boundary, so the formula would under-count — falling back to the pre-v4 path for that frame avoids a false negative.                                                                                                                         
                                                                                                                                    
  ## Changes                                                                                                                        
                            
  - `LidoContract`: added `get_contract_version()` and `get_deposited_for_current_report()`                                         
    (reads `depositedForCurrentReport` — index 3 — from `getBalanceStats()`)
  - `AbnormalClRebase`:                                                                                                             
    - `_calculate_injected_capital` — new method replacing the inline validator-count subtraction                                   
    - `_lido_version_at_block` — version check with a class-level "earliest confirmed upgrade block" cache to minimise RPC calls    
    - `_is_negative_specific_cl_rebase` — simplified: explicit same-block check removed, deduplication handled via `dict` keyed by `block_number`                                                                                                                    
  - Tests: `reset_upgrade_cache` autouse fixture added; v4 branch covered by `test_calculate_injected_capital__v4__correct_wei_to_gwei_conversion`


src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -191,15 +188,14 @@ def _calculate_cl_rebase_between_blocks(
...
         raw_cl_rebase = ref_balance_with_vault - prev_balance_with_vault
 
-        # We should account for validators who have been appeared between blocks
-        validators_count_diff_in_gwei = AbnormalClRebase.calculate_validators_count_diff_in_gwei(
-            prev_lido_validators, self.lido_validators
-        )
         # And withdrawals from WithdrawalVault
         withdrawn_from_vault = self._get_withdrawn_from_vault_between_blocks(prev_blockstamp, ref_blockstamp)
 
+        # Capital injected via deposits: new validators (pre-v4) or new validators + top-ups (post-v4)
```
F4ever:
```suggestion
        # Capital injected via deposits: new validators (pre-v4) 
        # or new validators + top-ups (post-v4)
```

src/providers/execution/contracts/lido.py
```diff
@@ -73,6 +73,31 @@ def get_beacon_stat(self, block_identifier: BlockIdentifier = 'latest') -> Beaco
         )
         return response
 
+    def get_contract_version(self, block_identifier: BlockIdentifier = 'latest') -> int:
```
F4ever: Let's remove 'latest'. We are trying to get rid of the default value to make sure we didn't lose the block_identifier arg.

src/providers/execution/contracts/lido.py
```diff
@@ -73,6 +73,31 @@ def get_beacon_stat(self, block_identifier: BlockIdentifier = 'latest') -> Beaco
...
+        response = self.functions.getBalanceStats().call(block_identifier=block_identifier)
+        logger.info(
+            {
+                'msg': 'Call `getBalanceStats()`.',
+                'value': response,
+                'block_identifier': repr(block_identifier),
+                'to': self.address,
+            }
+        )
+        return Wei(response[3])  # depositedForCurrentReport is the 4th return value
```
F4ever: Useless comment. Better add link to struct in core.

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -273,6 +271,75 @@ def _get_eth_distributed_events(self, from_block: BlockNumber, to_block: BlockNu
             )
         )
 
+    def _lido_version_at_block(self, block_number: int) -> int:
```
F4ever:
```suggestion
    def _lido_version_at_block(self, block_number: BlockNumber) -> int:
```

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -273,6 +271,75 @@ def _get_eth_distributed_events(self, from_block: BlockNumber, to_block: BlockNu
...
+        """Lido contract version at `block_number`, with an earliest-confirmed-upgrade cache.
+
+        Caches the lowest block number at which v4 is confirmed so that calls for
+        later blocks skip the RPC, while calls for earlier blocks still probe.
```
F4ever: Why do we need such cache? Maybe just wrap `get_contract_version` with lru_cache and remove this function?

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -273,6 +271,75 @@ def _get_eth_distributed_events(self, from_block: BlockNumber, to_block: BlockNu
...
+
+    def _calculate_injected_capital(
+        self,
+        prev_blockstamp: BlockStamp,
+        ref_blockstamp: ReferenceBlockStamp,
+        prev_lido_validators: list[LidoValidator],
+    ) -> Gwei:
+        """ETH injected via Lido deposits in the window [prev_blockstamp, ref_blockstamp].
+
+        Pre-v4 (no top-ups): counts new validators by actual balance, matching the
```
F4ever: counts new validators only as 32 ETH deposits

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -273,6 +271,75 @@ def _get_eth_distributed_events(self, from_block: BlockNumber, to_block: BlockNu
...
+        conservative behaviour of calculate_validators_count_diff_in_gwei.
+        Post-v4 (EIP-7251): uses the pending-queue conservation identity so top-ups
+        to existing validators are also accounted for.
+        """
```
F4ever, line 283: Please, simplify this comment and make it more human-like

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -273,6 +271,75 @@ def _get_eth_distributed_events(self, from_block: BlockNumber, to_block: BlockNu
...
+        if self._lido_version_at_block(last_report_blockstamp.block_number) < _LIDO_V4:
+            return AbnormalClRebase.calculate_validators_count_diff_in_gwei(
+                prev_lido_validators, self.lido_validators
+            )
+
+        lido_pubkeys = {key.key for key in self.lido_keys}
+        lido_wc_list = self.w3.lido_validators.get_lido_wc_list(ref_blockstamp)
+        genesis_fork_version = hex_str_to_bytes(self.w3.cc.get_genesis().genesis_fork_version)
+
+        def valid_lido_pending(blockstamp: BlockStamp) -> Gwei:
```
F4ever: rename into smth like `sum_of_lido_valid_deposits`

F4ever: Why this function declared here, let's move it on the method level

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -273,6 +271,75 @@ def _get_eth_distributed_events(self, from_block: BlockNumber, to_block: BlockNu
...
+                if d.pubkey in lido_pubkeys
+                and d.withdrawal_credentials in lido_wc_list
+                and is_valid_deposit_signature(
+                    pubkey=hex_str_to_bytes(d.pubkey),
+                    withdrawal_credentials=hex_str_to_bytes(d.withdrawal_credentials),
+                    amount=d.amount,
+                    signature=hex_str_to_bytes(d.signature),
+                    genesis_fork_version=genesis_fork_version,
+                )
+            ))
```
F4ever: Not sure logic is valid. Try to reuse logic implemented in lido_validators.
```
1. Get all pending_deposits
2. If validator already exist on provided block. sum this deposit without additional checks
3. else check if this key is in KAPI
4. find the first deposit for each key that is_valid
5. sum all deposits after valid for key if first deposit has lido wc else ignore
```

src/services/bunker_cases/abnormal_cl_rebase.py
hweawer: This cache preevents checking this every 24h

src/services/bunker_cases/abnormal_cl_rebase.py
hweawer: done

src/services/bunker_cases/abnormal_cl_rebase.py
hweawer: done

src/providers/execution/contracts/lido.py
hweawer: done

src/providers/execution/contracts/lido.py
hweawer: done

src/services/bunker_cases/abnormal_cl_rebase.py
hweawer: done

src/web3py/extensions/lido_validators.py
```diff
@@ -231,69 +231,80 @@ def get_lido_wc_list(self, blockstamp: BlockStamp) -> list[HexStr]:
...
+
+    @lru_cache(maxsize=1)
+    def get_pending_lido_validators(
+        self,
+        blockstamp: BlockStamp,
+    ) -> dict[HexStr, PendingValidator]:
+        """
+        Return the list of Lido keys that already have pending deposits on the CL.
+
+        Validates all BLS signatures and ensures there are no front-running attacks.
```
F4ever: We are not validating front-run, but filtering invalid deposits

hweawer: Changed the comment

src/services/bunker_cases/abnormal_cl_rebase.py
hweawer: removed - discussed offline

src/web3py/extensions/lido_validators.py
```diff
@@ -231,69 +231,80 @@ def get_lido_wc_list(self, blockstamp: BlockStamp) -> list[HexStr]:
...
-        Return the list of Lido keys that already have pending deposits on the CL.
-
-        Validates all BLS signatures and ensures there are no front-running attacks.
+    @staticmethod
+    def _collect_valid_pending_deposits(
+        pending_deposits: list[PendingDeposit],
+        filter_pubkeys: set[str],
+        lido_wc_list: list[HexStr],
+        genesis_fork_version: bytes,
+    ) -> dict[str, list[PendingDeposit]]:
```
F4ever, line 264:
```suggestion
    ) -> dict[HexStr, list[PendingDeposit]]:
```

src/web3py/extensions/lido_validators.py
```diff
@@ -231,69 +231,80 @@ def get_lido_wc_list(self, blockstamp: BlockStamp) -> list[HexStr]:
...
+        """
+        Return the list of Lido keys that already have pending deposits on the CL.
+
+        Validates BLS signatures and filters out deposits with non-Lido withdrawal credentials.
+        """
+        lido_wc_list = self.get_lido_wc_list(blockstamp)
+        genesis_config = self.w3.cc.get_genesis()
+        pending_deposits = self.w3.cc.get_pending_deposits(blockstamp)
+        pending_lido_keys = self._get_lido_validators_with_keys(blockstamp)[1]
+        pending_keys: dict[str, LidoKey] = {key.key: key for key in pending_lido_keys}
```
F4ever:
```suggestion
        pending_keys: dict[HexStr, LidoKey] = {key.key: key for key in pending_lido_keys}
```

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -273,6 +270,76 @@ def _get_eth_distributed_events(self, from_block: BlockNumber, to_block: BlockNu
...
+        deposit has Lido withdrawal credentials.
+        """
+        pending_deposits = self.w3.cc.get_pending_deposits(blockstamp)
+        total = Gwei(0)
+
+        for d in pending_deposits:
+            if d.pubkey in existing_pubkeys:
+                total = Gwei(total + d.amount)
+
+        valid = LidoValidatorsProvider._collect_valid_pending_deposits(
```
F4ever, line 343: Seems here we could have double accounting.
1. We calcualte all topups
2. We caluclate all topups + new deposits

We should filter validators on CL from lido_pubkeys.

Look into `self._get_lido_validators_with_keys(blockstamp)[1]`

src/web3py/extensions/lido_validators.py
```diff
@@ -231,69 +231,80 @@ def get_lido_wc_list(self, blockstamp: BlockStamp) -> list[HexStr]:
...
-    def get_pending_lido_validators(
-        self,
-        blockstamp: BlockStamp,
-    ) -> dict[HexStr, PendingValidator]:
-        """
-        Return the list of Lido keys that already have pending deposits on the CL.
-
-        Validates all BLS signatures and ensures there are no front-running attacks.
+    @staticmethod
+    def _collect_valid_pending_deposits(
```
F4ever, line 259: Can you change method order in class. Since one is calling another
`def get_pending_lido_validators: ...`
`def _collect_valid_pending_deposits: ..`

F4ever (dismissed):
Check linters pls

F4ever (approved):

PR #915, opened 2026-05-07
fix: external validator weight calculation

F4ever:
* In `_process_group` in `exit_order_iterator.py`, the assignment for `weight` of external node operators is changed from incrementing (`+=`) to setting (`=`), ensuring each operator's weight is exactly the share of `internal_weight` divided among them.

### DIFF
--- a/src/services/exit_order_iterator.py
+++ b/src/services/exit_order_iterator.py
@@ -371,7 +371,7 @@ class ValidatorExitIterator:
 
         for gid in external_gids:
             self.node_operators_stats[gid].total_stake += internal_balance / len(external_gids)
-            self.node_operators_stats[gid].weight += internal_weight / len(external_gids)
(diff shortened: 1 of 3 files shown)


src/services/exit_order_iterator.py
```diff
@@ -371,7 +371,7 @@ def _process_group(self, group: OperatorGroup, cm_v1: CMv1, cm_v2: CMv2) -> None
 
         for gid in external_gids:
             self.node_operators_stats[gid].total_stake += internal_balance / len(external_gids)
-            self.node_operators_stats[gid].weight += internal_weight / len(external_gids)
```
chasingrainbows, line 374: Comment "why?" needed here

chasingrainbows (approved):

madlabman (approved):

PR #926, opened 2026-05-21
EIP-7732: correct sweep calculation

hweawer:
## Summary

Fixes sweep delay calculation for EIP-7732 (Glamsterdam/ePBS) to be safe against manipulation and always request at least as many validator exits as needed.

**Problem:** The previous implementation included `pending_partial_withdrawals` in the sweep calculation. An attacker could queue future-dated EIP-7002 partial withdrawals to inflate `epochs_to_sweep`, which inflates projected `future_rewards`, causing the ejector to request too few exits (see Immunefi #79957).

**Changes:**
- Remove `pending_partial_withdrawals` from both the numerator (withdrawals count) and denominator (available slots per payload). Pending partials are externally manipulable via EIP-7002 and are not needed for a conservative ejection estimate.
- Remove `get_builders_sweep_withdrawals` — builder sweep was also dynamic state, excluded for the same reason.
- Remove `SweepPrediction` dataclass, replaced with a plain tuple.
- Post-ePBS: `available_per_payload = MAX_WITHDRAWALS_PER_PAYLOAD − builder_pending_per_block`. Builder pending withdrawals are a real protocol constraint (~1 slot, not EIP-7002-manipulable), so they are retained.
- Pre-ePBS: `available_per_payload = MAX_WITHDRAWALS_PER_PAYLOAD = 16` (unchanged effective behaviour).

**Safety invariant:** `estimated_delay ≤ real_delay` in all states. Real validator sweep capacity = `16 − builder_pending − actual_partials − builder_sweep`. Our estimate omits `actual_partials` and `builder_sweep` from the subtraction, so our available ≥ real available, our delay ≤ real delay, our projected future rewards ≤ real future rewards → ejector requests ≥ exits needed.


src/modules/oracles/ejector/sweep.py
```diff
@@ -25,78 +24,54 @@ class Withdrawal:
...
+        builder_pending_per_block = min(len(state.builder_pending_withdrawals), MAX_WITHDRAWALS_PER_PAYLOAD - 1)
+        available_per_payload = MAX_WITHDRAWALS_PER_PAYLOAD - builder_pending_per_block
+    else:
+        available_per_payload = MAX_WITHDRAWALS_PER_PAYLOAD
 
-    return withdrawals_number
+    return len(validators_withdrawals), available_per_payload
 
 
 def get_pending_partial_withdrawals(state: BeaconStateView) -> list[Withdrawal]:
```
chasingrainbows, line 118: get_pending_partial_withdrawals not used anywhere except tests anymore. Should be deleted?

src/modules/oracles/ejector/sweep.py
```diff
@@ -25,78 +24,54 @@ class Withdrawal:
...
 
 
-def get_sweep_delay_in_epochs(state: BeaconStateView, spec: ChainConfig) -> int:
-    """
-    This method predicts the average withdrawal delay in epochs.
-    It is assumed that on average, a validator sweep is achieved in half the time of a full sweep cycle.
+def get_sweep_delay_in_epochs(
+    state: BeaconStateView,
+    spec: ChainConfig,
+    is_epbs_active: bool = False,
```
chasingrainbows: cant find any place except tests where is_epbs_active=True will be used, why?

src/modules/oracles/ejector/sweep.py
```diff
@@ -25,78 +24,54 @@ class Withdrawal:
...
+    Post-ePBS: builder_pending_withdrawals are a real protocol constraint (~1 slot) and
+    are subtracted from available_per_payload. They are not externally manipulable via
+    EIP-7002, so including them does not open an attack surface.
 
-    pending_partial_withdrawals_number_in_cycle = min(
-        pending_partial_withdrawals_number, pending_partial_withdrawals_max_number_in_cycle
-    )
+    Assumption: MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP is never reached.
+    """
+    validators_withdrawals = get_validators_withdrawals(state, [], slots_per_epoch)
```
chasingrainbows: there is no point in partial_withdrawals in get_validators_withdrawals parameters now, because it always empty list now

chasingrainbows (changes requested):

src/modules/oracles/ejector/sweep.py
hweawer: removed

src/modules/oracles/ejector/sweep.py
hweawer: Before activating the fork ethereum foundation puts the fork epoch in the config of the chain, based on this epoch we write oracle logic - this is the feature flag for our changes. See [electra](https://github.com/ethereum/consensus-specs/blob/master/specs/electra/fork.md#configuration)

For gloas epoch is TBD

src/modules/oracles/ejector/sweep.py
hweawer: removed

src/modules/oracles/ejector/sweep.py
chasingrainbows: Sure, but where and when this flag will be used? I dont see any conditions at startup or ENV.

chasingrainbows (approved):

src/modules/oracles/ejector/sweep.py
```diff
@@ -25,127 +21,68 @@ class Withdrawal:
...
 
 
-def get_sweep_delay_in_epochs(state: BeaconStateView, spec: ChainConfig) -> int:
-    """
-    This method predicts the average withdrawal delay in epochs.
-    It is assumed that on average, a validator sweep is achieved in half the time of a full sweep cycle.
+def get_sweep_delay_in_epochs(
+    state: BeaconStateView,
+    spec: ChainConfig,
+    is_epbs_active: bool = False,  # feature flag: set to True once glamsterdam slot is known
```
F4ever: Maybe rename to
`is_gloas_activated`

F4ever: Should be True if state ref_epoch passed or equal gloas emoch

src/modules/oracles/ejector/sweep.py
```diff
@@ -25,127 +21,68 @@ class Withdrawal:
...
-                Withdrawal(
-                    validator_index=index,
-                    amount=withdrawable_balance,
-                )
-            )
+    validators_withdrawals = get_validators_withdrawals(state, slots_per_epoch)
 
-    return withdrawals
+    if is_epbs_active:
+        builder_pending_per_block = min(len(state.builder_pending_withdrawals), MAX_WITHDRAWALS_PER_PAYLOAD - 1)
```
F4ever: Usually builder_pending_withdrawals is 1 almost always
available_per_payload = MAX_WITHDRAWALS_PER_PAYLOAD - USUAL_BUILDER_WITHDRAWALS (16 - 1)

hweawer: Done

src/modules/oracles/ejector/sweep.py
hweawer: Done

hweawer:
@F4ever @chasingrainbows as discussed offline - now we completely ignore builder_pending_withdrawals, updated implementation

PR #928, opened 2026-05-26
feat: Correct CL balance for EIP-7732

hweawer:
# Problem   

##  Under EIP-7732 (Gloas fork), Lido's TVL identity breaks silently:
  
  TVL = clValidatorsBalance + withdrawalVaultBalance

  `process_withdrawals` in the beacon block for slot N reduces CL validator balances and populates
  `state.payload_expected_withdrawals`, but the matching ETH only arrives in the withdrawal vault when the execution payload envelope is processed at slot N+1. Meanwhile, lido-oracle anchors its EL reads to `state.latest_block_hash` (the last revealed EL block, at most N-1), so the vault has not yet received those amounts.

  Result on every ref-slot with active Lido withdrawals:
  clValidatorsBalance = real_balance − W   (CL already deducted)
  withdrawalVaultBalance = real_vault − W  (EL hasn't credited yet)
  TVL = real_TVL − 2W                      → underreported

  This causes two downstream failures:
  1. AccountingOracle reports an incorrect TVL and share rate, producing wrong finalization batches.
  2. Bunker mode detection (abnormal_cl_rebase.py) sees a spurious negative rebase on frames with large withdrawal batches at the ref-slot → false bunker trigger.
 
  The issue is explicitly acknowledged in the Gloas spec (process_withdrawals note) and tracked in consensus-specs 
  PR #5128, added in response to Lido feedback.
  
 ## Fix

  Add `payload_expected_withdrawals` back to `clValidatorsBalance` when ePBS is active. Only Lido validators'
  withdrawals are included.

 ## Notes

  - All changes are gated on EPBS_FORK_EPOCH — zero behaviour change on any network until the Gloas fork is
  activated 
  - EPBS_FORK_EPOCH constant name is a placeholder; needs to be confirmed against the final spec before the fork is scheduled


src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -216,9 +216,26 @@ def _get_lido_validators_balance_with_vault(
...
+        if isinstance(blockstamp, ReferenceBlockStamp):
+            spec = self.w3.cc.get_config_spec()
+            if blockstamp.ref_epoch >= spec.EPBS_FORK_EPOCH:
+                state = self.w3.cc.get_state_view(blockstamp)
+                lido_indices = {v.index for v in lido_validators}
+                correction = Gwei(sum(
+                    w.amount for w in state.payload_expected_withdrawals
+                    if w.validator_index in lido_indices
+                ))
+                real_cl_balance = Gwei(real_cl_balance + correction)
```
chasingrainbows: Looks like logic dublicated from _get_cl_validators_balance upper. Could we bring this logic into some common method/etc?

src/modules/oracles/accounting/accounting.py
```diff
@@ -248,8 +248,24 @@ def _get_cl_validators_balance(self, blockstamp: ReferenceBlockStamp) -> Gwei:
         validator_balance_sum = Gwei(sum(validator.balance for validator in lido_validators))
         logger.info({'msg': 'Calculate active balance.', 'value': validator_balance_sum})
 
+        if self._is_epbs_active(blockstamp):
```
chasingrainbows: The ePBS correction is applied to the total CL balance in _get_cl_validators_balance, but in _get_balances_by_modules still sums raw validator.balance.

Both feed the same ReportData. Should we apply the same correction per module?

chasingrainbows (changes requested):

src/services/bunker_cases/abnormal_cl_rebase.py
hweawer: Done

src/modules/oracles/accounting/accounting.py
hweawer: Done

chasingrainbows (approved):

PR #935, opened 2026-06-09
feat(orc-329): remove curl from production

chasingrainbows:
## Description
This pull request updates the health check mechanism for several Docker services by replacing direct `curl` commands with a new Python-based health check script. The main goal is to remove curl dependency from production because of flaky versioning in debian repo which leads to necessity frequent updates of Dockerfile.

## Related Issue/Task

## How Has This Been Tested?

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ ARG SOURCE_DATE_EPOCH
 RUN apt-get update && apt-get install -y --no-install-recommends -qq \
     libffi-dev=3.4.8-2 \
     g++=4:14.2.0-1 \
-    curl=8.14.1-2+deb13u3 \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* \
  && rm -rf /var/cache/* \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -52,6 +51,7 @@ RUN pip install --no-cache-dir poetry==${POETRY_VERSION}
 RUN apt-get update && apt-get install -y --no-install-recommends -qq \
     git=1:2.47.3-0+deb13u1 \
     htop=3.4.1-5 \
+    curl=8.14.1-2+deb13u3 \
```
hweawer, line 57: Does it solve the problem of failing pipeline? If not maybe other options can work:
- download foundry with python
- copy foundry from the foundry image

chasingrainbows: > Does it solve the problem of failing pipeline?

For now, yep, after fix building can be failed on dev Docker stage which used on dev environment only for now.

F4ever (approved):

PR #942, opened 2026-06-23
feat(orc-112): add final tests before image publishing

chasingrainbows:
## Description
This pull request updates the CI/CD workflows and Dockerfile to improve the reliability and consistency of testing across development, staging, and production environments. The main changes introduce a dedicated Docker build stage for testing, and update the workflows to run tests inside this standardized environment before deploying.

## Related Issue/Task

## How Has This Been Tested?
Runned locally because u need to prepare release to test it fully on CI
- [] Not tested (explain why)

## Checklist

### DIFF
--- a/.github/workflows/ci-dev.yml
+++ b/.github/workflows/ci-dev.yml
@@ -9,17 +9,60 @@ on:
       - ".github/**"
       - "README.md"
 
-permissions: {}
+permissions:
+  contents: read
 
 jobs:
-  # test:
-  #   ...
-
   deploy:
     runs-on: ubuntu-latest
-    # needs: test
     name: Build and deploy
     steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Build test image
+        run: docker build --target test -t oracle:test .
+
+      - name: Test the image with pytest
+        timeout-minutes: 60
+        run: >
+          docker run --rm
+          -e EXECUTION_CLIENT_URI
+          -e CONSENSUS_CLIENT_URI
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-dev.yml
```diff
@@ -9,17 +9,61 @@ on:
...
 jobs:
-  # test:
-  #   ...
-
   deploy:
     runs-on: ubuntu-latest
-    # needs: test
     name: Build and deploy
     steps:
+      - name: Checkout
```
hweawer, line 26: Maybe lets move 

steps:
      - name: Checkout
      - name: Set up Docker Buildx
      - name: Build test image
      - name: Test the image with pytest

to a separate yml, and then reuse it 

```
jobs:
  test:
    uses: ./.github/workflows/_test-image.yml
    secrets: inherit
```

We also have tests.yml - we can run from the image here as well, but ignore this pipeline on pushes to master/develop

WDYT?

chasingrainbows: 1. The idea was to run tests inline in the deploy job, right before the deploy step - not as a separate "gate" job before it.
2. A separate _test-image.yml can't do that: GitHub Actions only allows calling a reusable workflow as a separate job via needs:. That breaks atomicity - the deploy job is scheduled separately and may start an unbounded time after the tests pass, so we'd deploy against stale results. Inline steps run back-to-back on the same runner: what was tested is what gets deployed, immediately.
3. But, now i moved shared action in one file - a local composite action (.github/actions/test-image) called as a regular step from each deploy job and from tests.yml. The only duplication left is the secrets env: block per call site, since composite actions can't read the secrets context.

PR #948, opened 2026-07-08
fix: fix abnormal cl rebase calculation

hweawer:
## Summary

Two separate fixes to `AbnormalClRebase` (Lido v4+ / EIP-7251 abnormal CL rebase / bunker-mode check):

1. **Formula fix** — `_calculate_injected_capital` was silently under/over-counting deposits netted out of the intraframe CL rebase delta.
2. **Validation** — `rebase_check_distant_epoch_distance` can be misconfigured to point outside the current reporting frame; this is now rejected explicitly instead of silently producing a wrong result.

## Fix 1: injected-capital formula

**Problem.** The old code computed injected capital as:

```python
deposited_in_window = wei_to_gwei(Wei(
    self.w3.lido_contracts.lido.get_deposited_for_current_report(ref_blockstamp.block_hash)
    - self.w3.lido_contracts.lido.get_deposited_for_current_report(prev_blockstamp.block_hash)
))
```

`get_deposited_for_current_report` (`depositedForCurrentReport` on-chain) re-derives its value against the **current, still-open** HashConsensus frame on every call. Reading it at `prev_blockstamp` and `ref_blockstamp` within the same open frame always returns the identical number, so the subtraction always canceled out to `0` — deposit capital was never netted out of the raw CL rebase delta, regardless of how much was actually deposited.

**Fix.** Replaced with `LidoContract.get_balance_stats()` (new `BalanceStats` dataclass, `src/modules/oracles/accounting/types.py`), and combine two fields from it:

```python
ref_balance_stats = self.w3.lido_contracts.lido.get_balance_stats(ref_blockstamp.block_hash)
prev_balance_stats = self.w3.lido_contracts.lido.get_balance_stats(prev_blockstamp.block_hash)
deposited_in_window = wei_to_gwei(
    Wei(
        (ref_balance_stats.deposited_since_last_report - ref_balance_stats.deposited_since_last_report)
        - (prev_balance_stats.deposited_since_last_report - prev_balance_stats.deposited_for_current_report)
    )
)
```

`deposited_since_last_report` (`depositedPostReport`) only resets when a report is actually processed on-chain, so a naive diff of two reads is close — except if an old, delayed report settles and resets the counter somewhere between `prev_blockstamp` and `ref_blockstamp`, a plain diff would under-count by the "backlog" amount that got reset away. `prev_balance_stats.deposited_since_last_report - prev_balance_stats.deposited_for_current_report` recovers that backlog using only `prev`'s own self-consistent state, so subtracting it back out cancels the reset regardless of whether one occurred in between.

## Fix 2: validate `distant_slot` stays within the current frame

`_calculate_injected_capital`'s accounting (fix 1) assumes at most one report can settle between `distant_slot`/`nearest_slot` and `ref_slot`. That assumption breaks if `rebase_check_distant_epoch_distance` is configured large enough to reach back into an already-processed frame. Added `_validate_distant_slot_within_ref_frame`, which raises if `distant_slot` falls before the start of the reporting frame that `ref_slot` belongs to (using the existing `Web3Converter.get_frame_by_slot`/`get_frame_first_slot`, same mechanism already used by `MidtermSlashingPenalty`). `AbnormalClRebase` now takes `frame_config: FrameConfig` to build this check.

## Note

This diff also includes an unrelated `docker-compose.yml` change (a security-warning comment on the unauthenticated `performance-web` port). That's not part of this fix — it came in from `develop` via the merge of #949 (`fix/performance-web-api-fix`), authored by @chasingrainbows.


src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -161,6 +164,18 @@ def _get_nearest_and_distant_blockstamps(
 
         return nearest_blockstamp, distant_blockstamp
 
+    def _validate_distant_slot_within_ref_frame(self, distant_slot: SlotNumber, ref_slot: SlotNumber) -> None:
+        """distant_slot must stay within the same reporting frame as ref_slot: _calculate_injected_capital's
+        deposit accounting assumes at most one report can settle between distant/nearest_slot and ref_slot,
+        which no longer holds once distant_slot reaches back into an already-processed frame.
+        """
+        frame_first_slot = self._web3_converter.get_frame_first_slot(self._web3_converter.get_frame_by_slot(ref_slot))
+        if distant_slot < frame_first_slot:
```
F4ever: I suppose we can validate `nearest_slot` as well

hweawer: This happens in `validate_slot_distance`

src/providers/execution/contracts/lido.py
```diff
@@ -85,9 +85,21 @@ def get_contract_version(self, block_identifier: BlockIdentifier) -> int:
...
-        """Lido v4+: ETH deposited by Lido since the last oracle report, in wei."""
+    def get_balance_stats(self, block_identifier: BlockIdentifier) -> BalanceStats:
+        """Lido v4+: balance/deposit accounting state. See
+        https://github.com/lidofinance/core/blob/c2872dc75eae824a9959bb4a5f21caef792de1a1/contracts/0.4.24/Lido.sol#L734
+
+        `deposited_since_last_report` (`depositedPostReport`) only resets when a report is
+        actually processed on-chain, so diffing two readings taken within the same (still
+        unsettled) reporting frame gives the deposits made strictly between those two blocks.
+
+        Do not use `deposited_for_current_report` for that purpose: it is re-derived against
```
F4ever: better add param description to the struct itself, instead case-only description

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -302,13 +317,21 @@ def _calculate_injected_capital(
             lido_wc_list=lido_wc_list,
             genesis_fork_version=genesis_fork_version,
         )
-
-        deposited_in_window = wei_to_gwei(
-            Wei(
-                self.w3.lido_contracts.lido.get_deposited_for_current_report(ref_blockstamp.block_hash)
-                - self.w3.lido_contracts.lido.get_deposited_for_current_report(prev_blockstamp.block_hash)
-            )
+        ref_balance_stats = self.w3.lido_contracts.lido.get_balance_stats(ref_blockstamp.block_hash)
```
F4ever, line 323: let's move `_validate_distant_slot_within_ref_frame` to this func

hweawer: Done

src/providers/execution/contracts/lido.py
hweawer: Done

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -151,6 +153,7 @@ def _get_nearest_and_distant_blockstamps(
         )
 
         AbnormalClRebase.validate_slot_distance(distant_slot, nearest_slot, ref_blockstamp.slot_number)
+        self._validate_distant_slot_within_ref_frame(distant_slot, ref_blockstamp.ref_slot)
```
tamtamchik: Do we need to check here or by `distant_blockstamp.slot_number`?

hweawer: Moved this check later in the stack to `_calculate_injected_capital` as @F4ever proposed https://github.com/lidofinance/lido-oracle/pull/948#discussion_r3551075843

tamtamchik (approved):

chasingrainbows (approved):

PR #963, opened 2026-07-20
feat(gloas): build report blockstamps from ref_slot's child under ePBS

hweawer:
## Glamsterdam (Gloas / EIP-7732) — 1/4: shared BlockStamp infra

Under EIP-7732 a slot's payload, deposits and withdrawals reach the beacon state only when its
**child** block is processed. So post-fork a blockstamp for slot N is built from N's child — the
first non-missed block after N — and the execution anchor is the last payload that child's state has
applied, which the child's own payload bid carries as `parent_block_hash`.

Pre-fork behaviour is unchanged: the last non-missed block at or before N, anchored on its embedded
payload.

### What changes
- `get_blockstamp` / `get_reference_blockstamp` resolve the anchor block: `get_prev_non_missed_slot`
  pre-fork, `get_next_non_missed_slot` post-fork. Which side of the fork a slot falls on comes from
  `GLOAS_FORK_EPOCH` in the beacon config (`ConsensusClient.is_gloas` / `is_gloas_slot`); the block
  shape is a cross-check, and a disagreement raises `InconsistentData` rather than resolving to the
  wrong block.
- The execution anchor comes from the block body's `signed_execution_payload_bid`, so no blockstamp
  downloads a beacon state. `process_execution_payload_bid` asserts
  `bid.parent_block_hash == state.latest_block_hash` unconditionally, and the spec assigns
  `latest_block_hash` in exactly one place, before that assert — so it holds whether the parent
  payload was revealed or withheld.
- New consensus types: optional `execution_payload` and `signed_execution_payload_bid` on the block
  body, `payload_expected_withdrawals` and `GLOAS_FORK_EPOCH` on the config/state views.
  `get_config_spec` is now cached, like `get_genesis`.
- `get_next_non_missed_slot` + `ChildSlotNotFinalized`: a report waits and retries, it does not crash.
- Every daemon now takes an execution client, the performance collector included — see below.

### Operator-visible
The `performance_collector` sidecar now requires `EXECUTION_CLIENT_URI` and refuses to start without
it. Post-fork an execution client is the only way to turn an anchor hash into a block number and
timestamp; the alternative was placeholder EL fields, which read as block 0 in the sidecar's
`ORACLE_BLOCK_NUMBER` metric and left a trap for the next reader. README and `docker-compose.yml`
updated; worth a release-note line.

### Consequences
`BlockStamp` no longer describes one block. It holds three points, documented with a diagram on
`ReferenceBlockStamp`:
- `ref_slot` / `ref_epoch` — the on-chain label, verbatim from `getCurrentFrame()`;
- `slot_number` / `state_root` — the consensus anchor, ref_slot's **child**, so `slot_number` is now
  greater than `ref_slot`, inverting the pre-fork invariant;
- `block_hash` / `block_number` / `block_timestamp` — ref_slot's execution block, or an older one if
  its payload was withheld. Not the child's.

`epoch_of(slot_number)` is normally `ref_epoch + 1`: code that needs the report's epoch must read
`ref_epoch`. The report's CL view is one epoch transition past ref_slot — deliberate; the only CL→EL
flow in between is withdrawals, covered by the add-back in #964. A report waits for one extra
finalized slot.

### Verified on a Gloas node
glamsterdam-devnet-8 (Lighthouse v8.2.2):
- the bid/state equality at slots 137088, 137086 and 137051 — the last two after a withheld parent
  payload, 137051 also a self-build (`builder_index = 2**64-1`);
- `GLOAS_FORK_EPOCH = 1536` is served in `/eth/v1/config/spec`, and `is_gloas_slot` flips exactly
  between slots 49151 and 49152;
- the real resolvers: ref_slot 137087 → child 137088, missed ref_slot 137077 → child 137078,
  ref_slot 137085 → the payload of 137084 rather than its own withheld one, ref_slot 49151 → back to
  the pre-fork block at 49149.

### Consumers to fix before release
Both are consumer bugs, not blockstamp infra, so they belong in their own PRs — but this one must not
reach a release branch ahead of them:
- `get_events_in_past` computes `ref_slot_shift = ref_slot - slot_number`, which goes negative
  post-fork and widens the event window instead of narrowing it.
- The staking-vaults IPFS report derives `timestamp` from `slot_number`, publishing the child's
  timestamp under a `refSlot`-labelled artifact.

### Testing
`tests/utils/test_blockstamp_gloas.py` (anchor, fork boundary with a missed ref slot, both
config/shape disagreements), child-finalization gate tests in `tests/modules/submodules/consensus`,
and a skip-marked `tests/integration/test_gloas_blockstamp.py` holding the devnet invariants above.
Full unit suite green, `ruff` + `pyright` clean.

Base: `feat/oracle-v9`. This PR is the base for the accounting and ejector PRs.


src/modules/common/daemon_module.py
```diff
@@ -33,11 +35,19 @@ class DaemonModule(ABC):
     - Timeout management
     """
 
-    def __init__(self, cc: ConsensusClient, **kwargs) -> None:
+    def __init__(self, cc: ConsensusClient, el: Eth | None = None, **kwargs) -> None:
```
F4ever: Let's mark this mandatory and provide el for performance module also.

docs/glamsterdam-devnet-testplan.md
```diff
@@ -0,0 +1,165 @@
+# Glamsterdam (Gloas) Oracle — Devnet Test Plan
```
F4ever: Remove testplan from docs

F4ever:
I propose to modify Blockstamp struct. `ref_slot` -> refernce slot for report and `build_slot` -> slot used to build repor

docs/glamsterdam-devnet-testplan.md
hweawer: Done

src/modules/common/daemon_module.py
hweawer: Done

hweawer:
> I propose to modify Blockstamp struct. `ref_slot` -> refernce slot for report and `build_slot` -> slot used to build repor

we have this slot in the bs:

- ref_slot / ref_epoch — the on-chain label, stored verbatim from getCurrentFrame()
- slot_number + state_root — the CL anchor, i.e. ref_slot's child
- block_hash / block_number / block_timestamp — the EL anchor, read from state.latest_block_hash, which is ref_slot's execution block, not the child's

src/types.py
```diff
@@ -76,7 +76,9 @@ class BlockStamp:
 
 @dataclass(frozen=True)
 class ReferenceBlockStamp(BlockStamp):
-    # Ref slot could differ from slot_number if ref_slot was missed slot_number will be previous first non-missed slot
+    # ref_slot labels the report on-chain; slot_number is the block it is built from. Under
+    # EIP-7732 that block is ref_slot's child, so slot_number can be greater than ref_slot, and the
+    # block_* fields belong to ref_slot rather than to slot_number. See BlockstampBuilder.
```
F4ever: It isn’t clear from the description how it works. Let’s add a diagram here—something along these lines, but more polished.

Let’s also remove other mentions of this algorithm to avoid duplicating the same information.

```
ref_slot and ref_block is mark for onchain report
Report builds from next validated slot and from ref_block or closes before it (on which slot is pointing

                    ref_slot
     1         2       3         4       5
cl [X]      [ ]     [X]      [ ]      [X]
     1                                       2
el [X]      [ ]     [ ]        [ ]     [X]

So ref slot is 3, but slot is 5 and block is 1.
```

src/utils/slot.py
```diff
@@ -115,29 +119,31 @@ def get_prev_non_missed_slot(
...
     slot: SlotNumber,
     last_finalized_slot_number: SlotNumber,
-):
-    """Get first non-missed slot header and generates blockstamp for it"""
-    logger.info({'msg': f'Get Blockstamp for slot: {slot}'})
-    existed_slot = get_prev_non_missed_slot(cc, slot, last_finalized_slot_number)
-    logger.info({'msg': f'Resolved to slot: {existed_slot.message.slot}'})
-    return build_blockstamp(existed_slot)
+) -> BlockDetailsResponse:
+    """Get the first non-missed block strictly after `slot` (its child).
```
F4ever: Clear from function name. Let's remove redundant comments

src/services/bunker_cases/abnormal_cl_rebase.py
```diff
@@ -39,6 +40,10 @@ def __init__(self, w3: Web3, c_conf: ChainConfig, b_conf: BunkerConfig, frame_co
         self.b_conf = b_conf
         self._web3_converter = Web3Converter(c_conf, frame_config)
 
+    @cached_property
+    def _blockstamp_builder(self) -> BlockstampBuilder:
```
F4ever: why we need cached property if we call it two times in one place.
let's init it in method itself

src/utils/blockstamp.py
```diff
@@ -1,53 +1,167 @@
...
+def _el_fields_from_hash(el: Eth, el_block_hash: BlockHash) -> dict:
+    block = el.get_block(el_block_hash)
+    return {
+        "block_number": BlockNumber(block["number"]),  # type: ignore[typeddict-item]
+        "block_hash": add_0x_prefix(el_block_hash),
+        "block_timestamp": Timestamp(block["timestamp"]),  # type: ignore[typeddict-item]
+    }
+
+
+class BlockstampBuilder:
```
F4ever: Maybe use it as plain function as before. Just expand functions with additional param `el`. Seem this class doesn't need any state

tests/utils/test_fingerprint_mainnet_fixtures.py
```diff
@@ -20,7 +20,11 @@
 FIXTURES = pathlib.Path(__file__).parent / 'fixtures'
 
 # Mainnet, slot 14921600, first 100 entries of each list.
-BEACON_STATE_DIGEST = '0xb3f502f6d5cffc8af70622c453b83399b48d29da56e26fed5ed9a1c5c63c8bd0'
+# Repinned when BeaconStateView gained payload_expected_withdrawals and latest_block_hash: the
+# fixture predates Gloas so both take their defaults, but they are fingerprinted like every other
+# field, which moves the digest. Members running across this release boundary therefore cannot
+# compare beacon-state digests with each other.
```
F4ever: Seems this should go into commit message. Not related to code

src/modules/oracles/common/consensus.py
```diff
@@ -237,7 +242,8 @@ def get_blockstamp_for_report(self, last_finalized_blockstamp: BlockStamp) -> Re
 
         logger.info({'msg': 'Fetch member info.', 'value': member_info})
 
-        # Check if the current slot is higher than the member slot
+        # Necessary but not sufficient under EIP-7732, where the report is built on ref_slot's
+        # child: the resolver below raises ChildSlotNotFinalized for the other half.
```
F4ever: let's than change it and check if `last_finalized_blockstamp.slot_number < member_info.current_frame_ref_slot + 1` is True.

F4ever: or remove this check, since we have try catch later

src/services/safe_border.py
```diff
@@ -243,7 +248,9 @@ def _get_last_finalized_withdrawal_request_epoch(self) -> EpochNumber:
         return self.get_epoch_by_timestamp(last_finalized_request_data.timestamp)
 
     def _get_blockstamp(self, last_slot_in_frame: SlotNumber):
-        return get_blockstamp(self.w3.cc, last_slot_in_frame, self.blockstamp.ref_slot)
+        # Both slots are known finalized; under EIP-7732 the report's own block is the later one.
```
F4ever: We are not sure. If this is pectra yet, next slot could be not finalized

src/services/safe_border.py
```diff
@@ -243,7 +248,9 @@ def _get_last_finalized_withdrawal_request_epoch(self) -> EpochNumber:
...
+        last_finalized_slot = max(self.blockstamp.ref_slot, self.blockstamp.slot_number)
+        return self._blockstamp_builder.get_blockstamp(last_slot_in_frame, last_finalized_slot)
```
F4ever: I think we can remove this modifications. `last_slot_in_frame` will always be ref_slot, not higher

src/providers/consensus/types.py
```diff
@@ -129,6 +129,29 @@ class ExecutionPayload(Nested, FromResponse):
...
+class ExecutionPayloadBid(Nested, FromResponse):
+    """
+    EIP-7732 builder bid header (subset we need).
+
+    `parent_block_hash` is the execution block the builder builds on top of, and so the block's
+    execution-layer anchor: `process_execution_payload_bid` asserts
+    `bid.parent_block_hash == state.latest_block_hash`.
+
+    The `message.parent_block_hash` path is per the consensus-specs container; confirm against the
+    final beacon-APIs serialization before mainnet activation.
```
F4ever: propose to remove part of comment, since it's not relate to this part of code

src/providers/consensus/types.py
```diff
@@ -248,6 +284,22 @@ class PendingConsolidation(Nested):
...
+@dataclass
+class ExpectedWithdrawal(Nested, FromResponse):
+    """
+    A single entry in BeaconState.payload_expected_withdrawals (EIP-7732).
+
+    `process_withdrawals` deducts these amounts from CL validator balances before the matching
+    execution payload credits the withdrawal vault, so consumers that need CL/EL balance
+    consistency add them back.
+
+    FromResponse ignores the extra fields (index, address) the API returns alongside these.
```
F4ever: Let's remove this line. `FromResponse` is common class we use everywhere. Don't need to comment how it works in every place.

src/providers/consensus/types.py
```diff
@@ -248,6 +284,22 @@ class PendingConsolidation(Nested):
...
 
 
+@dataclass
+class ExpectedWithdrawal(Nested, FromResponse):
+    """
+    A single entry in BeaconState.payload_expected_withdrawals (EIP-7732).
+
+    `process_withdrawals` deducts these amounts from CL validator balances before the matching
+    execution payload credits the withdrawal vault, so consumers that need CL/EL balance
+    consistency add them back.
```
F4ever: Ask AI to rewrite using B2 level English

src/utils/slot.py
```diff
@@ -115,29 +119,31 @@ def get_prev_non_missed_slot(
...
 
+    Raises ChildSlotNotFinalized when no finalized block exists after `slot` yet.
+    """
+    if slot >= last_finalized_slot_number:
+        raise ChildSlotNotFinalized(
```
F4ever, line 128: We have such check inside `get_non_missed_slot_header`

src/utils/slot.py
```diff
@@ -115,29 +119,31 @@ def get_prev_non_missed_slot(
...
+            f'No finalized child block after slot [{slot}]; last finalized is [{last_finalized_slot_number}].'
+        )
+    _, existing_header = get_non_missed_slot_header(cc, SlotNumber(slot + 1), last_finalized_slot_number)
+
+    # Unlike the pre-fork resolver, this one reaches at slots the caller has not itself checked, so
+    # the finalized-only invariant is asserted rather than left to the scan bound.
+    child_slot = existing_header.data.header.message.slot
+    if child_slot > last_finalized_slot_number:
```
F4ever: `_check_block_header` already checks finalization?

src/utils/blockstamp.py
```diff
@@ -1,53 +1,167 @@
...
+    post-fork, where the child's state is the earliest one that has N's execution payload, deposits
+    and withdrawals applied.
+
+    The execution anchor is the anchor block's `state.latest_block_hash` — N's own execution block
+    when N's payload was revealed, an earlier one when it was withheld. Reports pay nothing extra
+    for it: they read that same state for validators and pending_deposits, and
+    `ConsensusClient.get_state_view` keys its cache on `(state_root, slot)` so both reads hit one
+    entry. Liveness blockstamps need only a block hash for `eth_call`s, so rather than download a
+    multi-gigabyte state every daemon cycle they take the identical value from the block's own bid:
+    `process_execution_payload_bid` asserts `bid.parent_block_hash == state.latest_block_hash`.
```
F4ever: Please, simplify this

src/utils/blockstamp.py
```diff
@@ -1,53 +1,167 @@
...
+        The fork is detected from the block shape — an embedded execution payload means
+        pre-EIP-7732 — rather than from a beacon-config fork epoch, whose key would silently select
+        the wrong branch if it were missing or renamed.
+        """
+        block = get_prev_non_missed_slot(self.cc, slot, last_finalized_slot_number)
+        if block.message.body.execution_payload is not None:
+            return block
+        return get_next_non_missed_slot(self.cc, slot, last_finalized_slot_number)
+
+    def _base_fields(self, slot_details: BlockDetailsResponse, read_anchor_from_state: bool) -> dict:
```
F4ever: Please, re-check method names in this file. These must be verbs. `_get_base_fields`

src/utils/blockstamp.py
```diff
@@ -1,53 +1,167 @@
...
+
+    def _anchor_hash(self, slot_details: BlockDetailsResponse, read_from_state: bool) -> BlockHash:
+        """The post-EIP-7732 execution-layer anchor of `slot_details`. See the class docstring."""
+        slot = slot_details.message.slot
+
+        if read_from_state:
+            anchor = self.cc.get_state_view((slot_details.message.state_root, slot)).latest_block_hash
+            if not anchor:
+                raise MissingExecutionAnchor(f'State at slot [{slot}] has no latest_block_hash.')
+            return anchor
```
F4ever: Why we need this branch?

F4ever: Why we need this branch?

F4ever (changes requested):

src/types.py
hweawer: Done

src/utils/slot.py
hweawer: Done

src/services/bunker_cases/abnormal_cl_rebase.py
hweawer: Done

src/utils/blockstamp.py
hweawer: Done

tests/utils/test_fingerprint_mainnet_fixtures.py
hweawer: Done

src/modules/oracles/common/consensus.py
hweawer: `+1` breaks pre-fork: when the finalized block sits exactly on ref_slot the report
is already buildable, so we'd lose a cycle on every frame. And post-fork `+1` isn't
enough anyway — with missed slots the child can be at ref_slot + 5.

Removing it breaks pre-fork too: there ChildSlotNotFinalized is never raised, we get
ValueError from get_non_missed_slot_header → "Unexpected error" → cycle counted as
ERROR

src/providers/consensus/types.py
hweawer: Done

src/providers/consensus/types.py
hweawer: Done

src/providers/consensus/types.py
hweawer: Done

src/utils/slot.py
hweawer: Right, the condition is the same one — get_non_missed_slot_header is called with
slot + 1, and `slot + 1 > last_finalized` is `slot >= last_finalized`.

It's here only for the exception type: get_non_missed_slot_header raises ValueError,
which exception_handler logs as "Unexpected error" and counts the cycle as ERROR,
while a report caller needs "wait and retry". Added a comment saying exactly that.

src/utils/slot.py
hweawer: Removed

src/utils/blockstamp.py
hweawer: Removed

src/utils/blockstamp.py
hweawer: Done

src/utils/blockstamp.py
hweawer: Branch is gone entirely — there's one anchor source now, the block's payload bid

hweawer: Branch is gone entirely — there's one anchor source now, the block's payload bid

src/utils/blockstamp.py
```diff
@@ -1,53 +1,167 @@
...
+
+        The fork is detected from the block shape — an embedded execution payload means
+        pre-EIP-7732 — rather than from a beacon-config fork epoch, whose key would silently select
+        the wrong branch if it were missing or renamed.
+        """
+        block = get_prev_non_missed_slot(self.cc, slot, last_finalized_slot_number)
+        if block.message.body.execution_payload is not None:
+            return block
+        return get_next_non_missed_slot(self.cc, slot, last_finalized_slot_number)
+
```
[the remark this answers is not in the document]
hweawer: Done

src/providers/consensus/client.py
```diff
@@ -89,11 +89,20 @@ def _make_get_request(self, host: str, endpoint: str, **kwargs) -> Response:
         session_manager = self._session_managers.get(host) or next(iter(self._session_managers.values()))
         return session_manager._timed_call(self.session.get, self._urljoin(host, endpoint), **kwargs)
 
+    @lru_cache(maxsize=1)
```
F4ever: Remove cache. This will cache it forever, but the response can be changed after hardfork

src/providers/consensus/client.py
```diff
@@ -89,11 +89,20 @@ def _make_get_request(self, host: str, endpoint: str, **kwargs) -> Response:
...
     def get_config_spec(self) -> BeaconSpecResponse:
         """Spec: https://ethereum.github.io/beacon-APIs/#/Config/getSpec"""
         data, _ = self._get(self.API_GET_SPEC, validate_response=data_is_dict)
         return BeaconSpecResponse.from_response(**data)
 
+    def is_gloas(self, epoch: EpochNumber) -> bool:
```
F4ever: Since we have `is_gloas_slot`, maybe rename to `is_gloas_epoch`?

src/modules/sidecars/performance/collector/entrypoint.py
```diff
@@ -14,9 +18,19 @@ def _build_consensus_client() -> ConsensusClient:
     )
 
 
+def _build_execution_client() -> Eth:
```
F4ever, line 20: Maybe just reuse default web3 buillder and reuse it?

src/providers/consensus/client.py
hweawer: doen

src/providers/consensus/client.py
hweawer: done

F4ever (approved):
LGTM!

One thing: I don't see any reason to split the CL and EL clients and pass them as separate args in functions. I'd prefer to see just the web3 used everywhere (e.g., in `get_blockstamp` or the `OracleModule` class).

PR #964, opened 2026-07-20
feat(gloas): in-flight withdrawal correction (accounting + ejector)

hweawer:
## Glamsterdam (Gloas / EIP-7732) — 2/4: in-flight withdrawal correction (accounting + ejector)

Stacked on #963 (`feat/gloas-blockstamp-infra`). Review after T1.

Under EIP-7732 `process_withdrawals` reduces CL validator balances while the matching execution
payload — the one that actually credits the withdrawal vault — is revealed only afterwards. With the
child-anchored blockstamp from #963, the report block always sits before that reveal, so at
`blockstamp.block_hash` the vault has never received the block's own withdrawals.

The Gloas spec calls this out explicitly and prescribes the fix:

> `state.balances` reflects the withdrawal deduction before the corresponding execution payload is
> confirmed, creating a transient asymmetry with the EL state at `state.latest_block_hash`.
> Off-chain consumers that require CL/EL balance consistency can reconstruct pre-deduction balances
> by adding back `state.payload_expected_withdrawals`.

Every place that pairs CL balances with an EL-side balance therefore has to add them back — on the
CL side **only**, so the same ETH is not counted twice once the payload lands.

The correction is **unconditional**, driven purely by the data: `payload_expected_withdrawals` does
not exist in pre-Gloas states, so the add-back is zero before the fork and no fork gate is needed.

### Accounting

- `gloas_balance_correction` helper (also excludes builder-registry indices ≥ `2**40`).
- Add-back at four call sites: total CL balance, per-module breakdown (attributed per validator so
  module sums still equal the total), bunker-mode balance, and staking-vault total value.
- The bunker add-back applies to **both ends** of an abnormal-rebase interval, not just the
  reference side. Every blockstamp carries the same CL/EL asymmetry now, so correcting only the ref
  side would overstate each rebase by the previous side's in-flight withdrawals. The previous
  blockstamp's state is fetched once and reused for both its validators and its expected
  withdrawals, so this costs no extra beacon-state download.

### Ejector

The ejector pairs CL balances with EL vault balances exactly like accounting does, so it carries the
same asymmetry — both `validator.balance` and `_get_total_el_balance` are short by the Lido-owned
in-flight batch.

The sharpest case is a **full withdrawal**: it leaves `validator.balance` at zero, which fails the
`balance > 0` arm of `is_fully_withdrawable_validator`, so the whole payout drops out of the
predicted liquidity rather than shrinking by a rounding error. Under-predicted liquidity makes the
ejector request validator exits that the committed withdrawals would have covered — and an exit
cannot be undone.

Corrected at the two terms that turn CL balances into expected EL liquidity:
`_get_withdrawable_lido_validators_balance` and the going-to-withdraw sum. Both are needed:
correcting one alone leaves the other short, and both read pre-deduction balances afterwards, which
is what the spec prescribes.

Deliberately left uncorrected, each for its own reason:

| Site | Why no correction |
|---|---|
| `_get_total_el_balance` | EL side — correcting it too would double-count once the payload lands |
| `_get_total_active_balance` | sums *effective* balances, which withdrawals do not touch until the epoch boundary |
| sweep projection (`predict_withdrawals_number_in_sweep_cycle`) | models the protocol's own view and must keep post-deduction balances |
| ejection loop (`get_predictable_inbound_balance`) | caps at the max effective balance, so an in-flight partial sweep cannot move it |
| consolidation sources | skipped by the surrounding sum, so their in-flight amounts are not part of it |

### Duplicate validator indices

The staking-vault correction map was built with a dict comprehension over
`payload_expected_withdrawals`, so a validator appearing twice in one payload kept only the last
amount and the vault total value was under-corrected.

Duplicates are reachable per spec:

- `get_pending_partial_withdrawals` emits one entry per queued EIP-7002 request, and a validator may
  have several queued;
- `get_validators_sweep_withdrawals` does **not** skip a validator already served earlier in the
  same payload — it only recomputes the balance through `get_balance_after_withdrawals`.

Impact was vault-only: the dropped amount is missing from that vault's Total Value, and therefore
from its Merkle leaf and the fee derived from it. The global and per-module totals stay correct
because they sum the list, so the "module sums == corrected total" check cannot catch it.

Per-index aggregation now lives in `gloas_correction_by_index`. The vault path is the only site that
needs the correction per validator rather than as a single sum, so it is the only one that builds a
map at all — the other three call `gloas_balance_correction`, which sums the list and is
structurally immune to this bug class.

### Testing

- `tests/modules/accounting/test_gloas_correction.py` — in-flight withdrawal added back;
  foreign-validator and builder-index exclusion; pre-fork state is a no-op; per-module sum ==
  corrected total; duplicate indices summed.
- `tests/modules/accounting/staking_vault/test_total_values.py` — the correction end-to-end through
  `get_vaults_total_values`, including two entries for one validator summing rather than
  overwriting. This is the site where a dropped duplicate changes a reported number.
- `tests/modules/ejector/test_ejector_gloas_correction.py` — in-flight full withdrawal restored;
  in-flight partial sweep restores the excess; duplicate indices summed; foreign and builder entries
  ignored; consolidation source not corrected; pre-fork state is a no-op.

### DIFF
--- a/src/modules/oracles/accounting/accounting.py
+++ b/src/modules/oracles/accounting/accounting.py
@@ -32,7 +32,6 @@ from src.modules.oracles.common.consensus import (
     InitialEpochIsYetToArriveRevert,
 )
 from src.modules.oracles.common.oracle_module import OracleModule
-from src.providers.consensus.types import ExpectedWithdrawal
 from src.providers.execution.contracts.accounting_oracle import AccountingOracleContract
 from src.services.bunker import BunkerService
 from src.services.staking_vaults import StakingVaultsService
@@ -48,7 +47,6 @@ from src.types import (
 from src.utils.apr import calculate_gross_core_apr
 from src.utils.cache import global_lru_cache as lru_cache
 from src.utils.units import gwei_to_wei
-from src.utils.validator_balance import gloas_balance_correction, gloas_correction_by_index
 from src.variables import ALLOW_REPORTING_IN_BUNKER_MODE
 from src.web3py.types import Web3
 
@@ -242,9 +240,6 @@ class Accounting(OracleModule[Web3]):
         self._update_metrics(report_data)
         return report_data
 
-    def _expected_withdrawals(self, blockstamp: ReferenceBlockStamp) -> list[ExpectedWithdrawal]:
-        return self.w3.cc.get_state_view(blockstamp).payload_expected_withdrawals
-
     def _get_cl_validators_balance(self, blockstamp: ReferenceBlockStamp) -> Gwei:
         lido_validators = self.w3.lido_validators.get_active_lido_validators(blockstamp)
         logger.info({'msg': 'Get lido validators.', 'value': len(lido_validators)})
@@ -253,9 +248,9 @@ class Accounting(OracleModule[Web3]):
         logger.info({'msg': 'Calculate active balance.', 'value': validator_balance_sum})
 
         lido_indices = {validator.index for validator in lido_validators}
-        correction = gloas_balance_correction(self._expected_withdrawals(blockstamp), lido_indices)
+        correction = self.w3.cc.get_state_view(blockstamp).in_flight_withdrawal_sum(lido_indices)
         if correction:
-            logger.info({'msg': 'Gloas in-flight withdrawal correction.', 'value': correction})
+            logger.info({'msg': 'In-flight withdrawals added back.', 'value': correction})
 
         return Gwei(validator_balance_sum + correction)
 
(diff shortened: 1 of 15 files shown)


src/modules/oracles/accounting/accounting.py
```diff
@@ -247,7 +247,12 @@ def _get_cl_validators_balance(self, blockstamp: ReferenceBlockStamp) -> Gwei:
         validator_balance_sum = Gwei(sum(validator.balance for validator in lido_validators))
         logger.info({'msg': 'Calculate active balance.', 'value': validator_balance_sum})
 
-        return validator_balance_sum
+        lido_indices = {validator.index for validator in lido_validators}
+        correction = self.w3.cc.get_state_view(blockstamp).in_flight_withdrawal_sum(lido_indices)
+        if correction:
+            logger.info({'msg': 'In-flight withdrawals added back.', 'value': correction})
+
+        return Gwei(validator_balance_sum + correction)
```
F4ever: Maybe better do it once in `def indexed_validators`

F4ever: I propose making changes only in `indexed_validators` and modifying all balances there. That way, we don't need to modify or add logic anywhere else.

PR #965, opened 2026-07-20
feat(gloas): EIP-8061 exit churn limit + EIP-7732 sweep-delay (ejector)

hweawer:
## Glamsterdam — 3/4: ejector sweep-delay (EIP-7732) + exit churn limit (EIP-8061)

Stacked on #963 (`feat/gloas-blockstamp-infra`). Review after T1.

Two Glamsterdam changes to the Validator Exit Bus Oracle's withdrawal-timing model, both gated on
the Gloas fork:

- **EIP-8061:** uncapped `get_exit_churn_limit` (`CHURN_LIMIT_QUOTIENT_GLOAS = 2**15`) used in
  `compute_exit_epoch_and_update_churn` once Gloas is active. The pre-fork capped formula
  overestimates the withdrawal epoch post-fork, which would make the ejector under-request exits
  (the unsafe direction). At ~40M ETH stake the new limit is ~1220 ETH/epoch, ~5× the capped value.
- **EIP-7732:** the externally-triggerable `pending_partial_withdrawals` queue is excluded from the
  sweep-delay projection once Gloas is active (its future size is not derivable from current
  protocol state, and it now competes for withdrawal slots at higher priority than the sweep). This
  keeps the estimated delay ≤ the real delay (over-ejection is the safe direction).

`GLOAS_FORK_EPOCH` and `ConsensusClient.is_gloas` live in this PR: churn and sweep are the only
places that change behaviour at the fork with nothing in the observed data to key off. Blockstamp
construction (#963) reads the fork off the block shape instead, so it does not depend on the config
key name being right.

### Testing
- New unit tests: `tests/modules/ejector/test_gloas_ejector.py` (churn ≈ 1220 ETH/epoch and ~5× the
  capped value; sweep excludes partials post-fork; pre-fork regression).

Note: `fix/fix-churn-calc` (effective-balance quantization) is a separate change and is **not**
included here.

---

Follow-up: the churn parameters are moved out of `src/constants.py` and read from
`/eth/v1/config/spec` in a separate PR stacked on this one.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -31,6 +31,8 @@ CHURN_LIMIT_QUOTIENT = 2**16
 # https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md#validator-cycle
 MIN_PER_EPOCH_CHURN_LIMIT_ELECTRA = Gwei(2**7 * 10**9)
 MAX_PER_EPOCH_ACTIVATION_EXIT_CHURN_LIMIT = Gwei(2**8 * 10**9)
+# EIP-8061 (Glamsterdam): the exit churn limit loses its cap and the quotient is halved.
+CHURN_LIMIT_QUOTIENT_GLOAS = 2**15
 # https://github.com/ethereum/consensus-specs/blob/master/specs/phase0/beacon-chain.md#time-parameters
 SLOTS_PER_HISTORICAL_ROOT = 2**13  # 8192
 # https://github.com/ethereum/consensus-specs/blob/master/specs/altair/beacon-chain.md#sync-committee
--- a/src/constants.py
+++ b/src/constants.py
@@ -31,7 +31,7 @@ CHURN_LIMIT_QUOTIENT = 2**16
 # https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md#validator-cycle
 MIN_PER_EPOCH_CHURN_LIMIT_ELECTRA = Gwei(2**7 * 10**9)
 MAX_PER_EPOCH_ACTIVATION_EXIT_CHURN_LIMIT = Gwei(2**8 * 10**9)
-# EIP-8061 (Glamsterdam): the exit churn limit loses its cap and the quotient is halved.
+# https://eips.ethereum.org/EIPS/eip-8061
 CHURN_LIMIT_QUOTIENT_GLOAS = 2**15
 # https://github.com/ethereum/consensus-specs/blob/master/specs/phase0/beacon-chain.md#time-parameters
 SLOTS_PER_HISTORICAL_ROOT = 2**13  # 8192
--- a/src/modules/oracles/ejector/ejector.py
+++ b/src/modules/oracles/ejector/ejector.py
@@ -39,6 +39,7 @@ from src.utils.validator_balance import (
 from src.utils.validator_state import (
     compute_activation_exit_epoch,
     get_activation_exit_churn_limit,
+    get_exit_churn_limit,
     is_active_validator,
     is_fully_withdrawable_validator,
 )
@@ -316,7 +317,14 @@ class Ejector(OracleModule[Web3]):
         https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md#new-compute_exit_epoch_and_update_churn
         """
         earliest_exit_epoch = max(state.earliest_exit_epoch, compute_activation_exit_epoch(blockstamp.ref_epoch))
-        per_epoch_churn = get_activation_exit_churn_limit(self._get_total_active_balance(blockstamp))
+        total_active_balance = self._get_total_active_balance(blockstamp)
+        if self.w3.cc.is_gloas(blockstamp.ref_epoch):
+            # EIP-8061 removes the cap on the exit churn limit and halves the quotient. Using the old
+            # capped formula post-fork overestimates withdrawal_epoch, which makes the ejector
+            # under-request exits (the unsafe direction).
+            per_epoch_churn = get_exit_churn_limit(total_active_balance)
+        else:
+            per_epoch_churn = get_activation_exit_churn_limit(total_active_balance)
         # New epoch for exits.
         if state.earliest_exit_epoch < earliest_exit_epoch:
             exit_balance_to_consume = per_epoch_churn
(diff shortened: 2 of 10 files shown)


src/providers/consensus/types.py
```diff
@@ -29,6 +29,10 @@ class BeaconSpecResponse(Nested, FromResponse):
     SLOTS_PER_HISTORICAL_ROOT: int
     SLOT_DURATION_MS: int = 0
     SECONDS_PER_SLOT: int = 0
+    # Glamsterdam/EIP-7732 (Gloas) activation epoch. Defaults to "never" so the fork-gated
+    # ejector paths stay inactive on pre-fork networks. Exact spec constant name to be confirmed
```
F4ever: why ejector only?
better add ToDo instead of comment

src/modules/oracles/ejector/ejector.py
```diff
@@ -316,7 +317,14 @@ def compute_exit_epoch_and_update_churn(
         https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md#new-compute_exit_epoch_and_update_churn
         """
         earliest_exit_epoch = max(state.earliest_exit_epoch, compute_activation_exit_epoch(blockstamp.ref_epoch))
-        per_epoch_churn = get_activation_exit_churn_limit(self._get_total_active_balance(blockstamp))
+        total_active_balance = self._get_total_active_balance(blockstamp)
+        if self.w3.cc.is_gloas(blockstamp.ref_epoch):
```
F4ever: we should use not ref_epoch, because ref_epoch could be pectra yet. Use real epoch from slot we used for building blockstamp

F4ever (changes requested):

src/modules/oracles/ejector/ejector.py
hweawer: Good catch — fixed in 0d54a2eb.

Both gates now key off `_state_epoch(blockstamp)` = epoch of `blockstamp.slot_number`, i.e. the block the state view is actually read from, instead of `ref_epoch`. Under EIP-7732 that block is `ref_slot`'s child (see `BlockstampBuilder._resolve_anchor_block`), so it can sit in a later epoch than `ref_epoch`; a missed `ref_slot` puts it in an earlier one. Either way the fork the state was produced under is what matters.

Covered by `TestForkGateEpoch` in `tests/modules/ejector/test_gloas_ejector.py` — the churn test fails if the gate is switched back to `ref_epoch`. `TestPredictedWithdrawableEpoch` is now explicitly pinned pre-Gloas.

src/providers/consensus/types.py
hweawer: Reworded in 0d54a2eb — the field comment no longer claims it is ejector-specific, and the open question is now a `TODO`:

```python
# Glamsterdam/EIP-7732 (Gloas) activation epoch. Defaults to "never" so fork-gated paths stay
# inactive on networks whose config does not announce the fork yet.
# TODO: confirm the constant name against the final consensus-specs Glamsterdam config before
#  mainnet activation — a renamed key would silently keep every fork-gated path disabled.
```

On "why ejector only": nothing restricts the field itself — the churn and sweep projections are just the only current call sites, because they are the only places that change behaviour at the fork with nothing in the observed data to key off. Everything else (blockstamp construction) detects the fork from the block shape, which is why a wrong/renamed config key cannot break it. That rationale lives in the `ConsensusClient.is_gloas` docstring.

src/modules/oracles/ejector/ejector.py
```diff
@@ -336,7 +344,11 @@ def _get_sweep_delay_in_epochs(self, blockstamp: ReferenceBlockStamp) -> int:
         """Returns the number of epochs that will take to sweep all validators in the chain."""
         chain_config = self.get_chain_config(blockstamp)
         state = self.w3.cc.get_state_view(blockstamp)
-        return get_sweep_delay_in_epochs(state, chain_config)
+        return get_sweep_delay_in_epochs(state, chain_config, self.w3.cc.is_gloas_epoch(self._state_epoch(blockstamp)))
+
+    def _state_epoch(self, blockstamp: ReferenceBlockStamp) -> EpochNumber:
```
F4ever: I suppose we can add `epoch_number` to the ReferenceBlockStamp struct, seems it could be usefull here and in future

src/utils/validator_state.py
```diff
@@ -132,6 +133,12 @@ def get_balance_churn_limit(total_active_balance: Gwei) -> Gwei:
     return Gwei(churn - churn % EFFECTIVE_BALANCE_INCREMENT)
 
 
+# @see https://eips.ethereum.org/EIPS/eip-8061
```
F4ever: add link to https://github.com/ethereum/consensus-specs/blob/81749c877c36760a9a5fca1e11bfd83ab6427bcc/specs/gloas/beacon-chain.md?plain=1#L1456

src/modules/oracles/ejector/sweep.py
```diff
@@ -25,21 +25,25 @@ class Withdrawal:
...
     full_sweep_cycle_in_epochs = math.ceil(
         withdrawals_number_in_sweep_cycle / MAX_WITHDRAWALS_PER_PAYLOAD / spec.slots_per_epoch
     )
 
     return full_sweep_cycle_in_epochs // 2
 
 
-def predict_withdrawals_number_in_sweep_cycle(state: BeaconStateView, slots_per_epoch: int) -> int:
+def predict_withdrawals_number_in_sweep_cycle(
+    state: BeaconStateView, slots_per_epoch: int, is_gloas_active: bool = False
```
F4ever:
```suggestion
    state: BeaconStateView, 
    slots_per_epoch: int, 
    is_gloas_active: bool,
```

src/modules/oracles/ejector/sweep.py
```diff
@@ -56,6 +60,12 @@ def predict_withdrawals_number_in_sweep_cycle(state: BeaconStateView, slots_per_
     in any group of 16,384 MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP consecutive validators is less than 1%.
     This makes such an event extremely unlikely. More details can be found in the research: https://hackmd.io/@lido/HyrhJeLOJe.
     """
+    if is_gloas_active:
+        # The partials queue is externally triggerable, so its post-EIP-7732 size is not derivable
+        # from current state. Dropping it keeps the estimate below the real delay, and a too-short
+        # delay only makes the ejector request more exits than needed.
```
F4ever: Not sure `EIP-7732` is the cause here. I'd propose just mentioning that we're simplifying this, and that in some cases it will lead to a few more exits than in prev version (not `than needed` because the 7002 requests we see now will only be executed once the validator we're requesting passes its withdrawable epoch).

src/constants.py
```diff
@@ -31,6 +31,8 @@
 # https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md#validator-cycle
 MIN_PER_EPOCH_CHURN_LIMIT_ELECTRA = Gwei(2**7 * 10**9)
 MAX_PER_EPOCH_ACTIVATION_EXIT_CHURN_LIMIT = Gwei(2**8 * 10**9)
+# https://eips.ethereum.org/EIPS/eip-8061
+CHURN_LIMIT_QUOTIENT_GLOAS = 2**15
```
F4ever: add link to ethereum spec repo

PR #966, opened 2026-07-20
feat(gloas): proposer-duties v2 endpoint with relaxed v1 fallback

hweawer:
## Glamsterdam — 4/4: proposer-duties v2 endpoint with relaxed v1 fallback (EIP-7917)

Independent of the other three PRs (based directly on `feat/oracle-v9`).

EIP-7917 moves proposer-lookahead one epoch earlier and changes the `dependent_root` the duties
endpoint returns. Post-Fulu the deprecated v1 endpoint reports it inconsistently across clients,
which the oracle's reorg-safety check hard-fails on.

### What changes
- Prefer `GET /eth/v2/validator/duties/proposer` (fork-invariant `dependent_root`); validated
  strictly — a mismatch is fatal.
- Fall back to v1 only on a 404 for v2 (e.g. Nimbus); on that path a `dependent_root` mismatch is a
  warning, not fatal.
- Performance collector computes both dependent roots: v1 = last slot of epoch-1, v2 = last slot of
  epoch-2.

### Testing
- New unit tests in `tests/providers/consensus/test_consensus_client.py` (v2 preferred; 404 →
  relaxed v1 fallback; non-404 propagates; v1 mismatch does not raise) and updated checkpoint tests.

### Note for reviewers
The v2 `dependent_root` uses **last slot of epoch-2** (the correct one-epoch EIP-7917 shift from
v1's last-slot-of-epoch-1). Confirm against a node that has shipped beacon-APIs#563.

Config compatibility (`SLOT_DURATION_MS`) is already present on the dev line — no code here.

Base: `feat/oracle-v9`.

### DIFF
--- a/src/providers/consensus/client.py
+++ b/src/providers/consensus/client.py
@@ -66,6 +66,7 @@ class ConsensusClient(HTTPProvider):
     API_GET_ATTESTATION_COMMITTEES = 'eth/v1/beacon/states/{}/committees'
     API_GET_SYNC_COMMITTEE = 'eth/v1/beacon/states/{}/sync_committees'
     API_GET_PROPOSER_DUTIES = 'eth/v1/validator/duties/proposer/{}'
+    API_GET_PROPOSER_DUTIES_V2 = 'eth/v2/validator/duties/proposer/{}'
     API_GET_STATE = 'eth/v2/debug/beacon/states/{}'
     API_GET_SPEC = 'eth/v1/config/spec'
     API_GET_GENESIS = 'eth/v1/beacon/genesis'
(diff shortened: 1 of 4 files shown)


src/providers/consensus/client.py
```diff
@@ -210,10 +211,35 @@ def get_sync_committee(self, blockstamp: BlockStamp, epoch: EpochNumber) -> Sync
...
+        except ConsensusClientError as error:
+            if error.status != HTTPStatus.NOT_FOUND:
+                raise
+            logger.info(
+                {
+                    'msg': f'{self.API_GET_PROPOSER_DUTIES_V2} returned 404, '
+                    f'falling back to {self.API_GET_PROPOSER_DUTIES} for epoch {epoch}.'
+                }
+            )
+            return self._get_proposer_duties(self.API_GET_PROPOSER_DUTIES, epoch, expected_dependent_root_v1)
```
F4ever: Do we need fallback here? Seems already all nodes supports v2 endpoint

PR #967, opened 2026-07-22
feat(edf): delegation + soft key rotation

F4ever:
This pull request introduces several important updates related to delegation contract support, key rotation, and reproducible builds. The main changes include updating the delegation contract ABI, introducing support for a second member key (for key rotation and delegation), and improving the reproducibility of Docker builds. There are also various documentation updates and minor dependency bumps.


tests/fork/test_delegation.py
```diff
@@ -20,10 +20,10 @@ def finalized_slots(real_finalized_slot):
 
 @pytest.fixture()
 def delegatee_account(accounts_from_fork, monkeypatch):
-    addresses, private_keys = accounts_from_fork
+    _, private_keys = accounts_from_fork
     account = Account.from_key(private_keys[1])
     monkeypatch.setattr(variables, 'ACCOUNT', account)
-    return addresses[1], private_keys[1]
+    return account, private_keys[1]
```
hweawer: Seems we don't need to return `private_keys[1]`

src/modules/oracles/common/consensus.py
```diff
@@ -162,9 +161,11 @@ def get_member_info(self, blockstamp: BlockStamp) -> MemberInfo:
...
         current_frame_consensus_report = current_frame_member_report = ZERO_HASH
 
-        if variables.ACCOUNT:
-            ACCOUNT_BALANCE.labels(str(variables.ACCOUNT.address)).set(
-                self.w3.eth.get_balance(variables.ACCOUNT.address)
+        hash_consensus_member_address = self.w3.signer.hash_consensus_member_address
+
+        if hash_consensus_member_address:
+            ACCOUNT_BALANCE.labels(str(hash_consensus_member_address)).set(
+                self.w3.eth.get_balance(hash_consensus_member_address)
```
hweawer: This is the address of delegation contract, right? But gas will be payed from the delegate. In the current implementation this metric doesn't make sense

src/providers/execution/contracts/delegation_contract.py
```diff
@@ -15,43 +14,42 @@
...
         logger.info(
-            {'msg': 'Call admin()', 'value': response, 'block_identifier': repr(block_identifier), 'to': self.address}
+            {'msg': 'Call owner()', 'value': response, 'block_identifier': repr(block_identifier), 'to': self.address}
         )
         return Web3.to_checksum_address(response)
 
-    def get_delegatee(self, block_identifier: BlockIdentifier) -> ChecksumAddress:
-        """Get current delegatee address"""
-        response = self.functions.delegatee().call(block_identifier=block_identifier)
+    def get_delegate(self, block_identifier: BlockIdentifier = 'latest') -> ChecksumAddress:
```
hweawer, line 47: We are always using getDelegate even if it was terminated - should we check for `isTerminated`?

hweawer:
Fix docs `delegation.md`

src/web3py/extensions/signer.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+
+    active_signer: LocalAccount | None
+    is_delegated: bool
+    delegation_contract: DelegationContract | None
+
+    def __init__(
+        self,
+        w3: Web3,
+        account_1: LocalAccount | None,
```
hweawer: maybe pass list here?

hweawer (commented):
Update abi please

tests/fork/test_delegation.py
F4ever: removed

src/modules/oracles/common/consensus.py
F4ever: fixed

src/providers/execution/contracts/delegation_contract.py
F4ever: It will return `None` if the contract is terminated. I don't think we need to additionally handle the termination case.

src/modules/oracles/common/consensus.py
```diff
@@ -185,23 +186,25 @@ def get_member_info(self, blockstamp: BlockStamp) -> MemberInfo:
...
-                raise IsNotMemberException(
-                    'Provided Account is not part of Oracle\'s members and has no submit role. '
-                    'For dry mode remove MEMBER_PRIV_KEY from variables.'
+                logger.warning(
+                    {
+                        'msg': 'Provided Account is not part of Oracle\'s members and has no submit role. '
+                        'This is expected while a key rotation is pending; the oracle will keep '
+                        'polling and pick it up automatically once enacted on-chain.',
+                    }
                 )
```
chasingrainbows, line 207: `IsNotMemberException` was removed. If the account is misconfigured (wrong `MEMBER_PRIV_KEY`, key removed from the committee, wrong contract), the oracle now runs forever without submitting reports. It only logs a warning that says the state is expected during key rotation. So a permanent misconfiguration looks the same as a normal rotation — no crash, no alert, no metric.

Suggestion: make these states distinguishable. For example, add a Prometheus gauge (e.g. `oracle_member_active`) or raise the log level to ERROR after N cycles in this state.

src/web3py/extensions/signer.py
```diff
@@ -0,0 +1,139 @@
...
+    ):
+        super().__init__(w3)
+        self.account_1 = account_1
+        self.account_2 = account_2
+
+        self.active_signer = None
+        self.is_delegated = False
+        self.delegation_contract = None
+
+        if delegation_contract_address:
```
chasingrainbows, line 55: The old DelegationModule validated the setup at startup - it raised NotConfiguredError / DelegateMismatchError when the delegate was unset or pointed to another key, so a broken deploy crashed right away. SignerModule.__init__ does no such check, so a deploy with DELEGATION_CONTRACT_ADDRESS set but no delegate assigned starts cleanly and just idles in dry mode forever. 

Also docs/delegation.md still describes the old errors and the restart-based rotation flow. Worth re-adding some fail-fast startup check (or a readiness signal an operator can alert on) and updating the doc.

src/web3py/extensions/tx_utils.py
```diff
@@ -23,29 +23,30 @@
...
 
         return None
 
-    def _manual_tx_processing(self, transaction, params: TxParams, account: LocalAccount):
+    def _manual_tx_processing(self, transaction, params: TxParams, account: LocalAccount) -> TxReceipt | None:
         logger.warning({'msg': 'Send transaction in manual mode.'})
         msg = f'\nGoing to send transaction to blockchain: \nTx args:\n{transaction.args}\nTx params:\n{params}\n'
         if prompt(f'{msg}Should we send this TX? [y/n]: '):
-            self._send_transaction(transaction, params, account)
+            return self._send_transaction(transaction, params, account)
```
chasingrainbows, line 48: Previously the receipt from an accepted manual tx was discarded and check_and_send_transaction returned None; this change fixes that. But test_manual_tx_processing only asserts _send_transaction was called, not the return value, so the fix has no test coverage. Worth one assertion on the returned receipt.

src/variables.py
```diff
@@ -28,13 +28,16 @@
 # - Account -
 MEMBER_PRIV_KEY = from_file_or_env('MEMBER_PRIV_KEY')
 ACCOUNT = Account.from_key(MEMBER_PRIV_KEY) if MEMBER_PRIV_KEY else None  # pylint: disable=no-value-for-parameter
+MEMBER_PRIV_KEY_2 = from_file_or_env('MEMBER_PRIV_KEY_2')
+ACCOUNT_2 = Account.from_key(MEMBER_PRIV_KEY_2) if MEMBER_PRIV_KEY_2 else None  # pylint: disable=no-value-for-parameter
```
chasingrainbows, line 32: Nothing validates that ACCOUNT and ACCOUNT_2 are different. If both env vars point to the same key, two-key rotation silently degrades to a single key. A startup check that ACCOUNT.address != ACCOUNT_2.address (when both are set) would fail fast.

src/web3py/extensions/signer.py
F4ever: done

src/modules/oracles/common/consensus.py
F4ever: It was done for new members - Oracle should not fail if it will be active in future only after voting.
If account miss-cofigured  we will see it instantly in telemetry

src/modules/oracles/common/consensus.py
```diff
@@ -487,12 +494,14 @@ def _get_slot_delay_before_data_submit(self, blockstamp: BlockStamp) -> int:
...
-        if member.is_submit_member or variables.ACCOUNT is None:
+        report_address = self.w3.signer.hash_consensus_member_address
+
+        if member.is_submit_member or report_address is None:
             return 0
 
         members, _ = self._get_consensus_contract_members(blockstamp)
 
-        mem_position = members.index(variables.ACCOUNT.address)
+        mem_position = members.index(report_address)
```
chasingrainbows, line 503: members.index(report_address) can now raise ValueError. Before this PR the path was not reachable: get_member_info raised IsNotMemberException one line earlier. Now, if the signer resolved on the latest block but is not yet in members on this blockstamp (possible mid-rotation), index() throws, and the cycle fails with a generic "Unexpected error" log without any details about the rotation. Suggestion: check report_address in members first and return 0 (or log a clear message) instead.

tests/web3py/test_signer_module.py
```diff
@@ -0,0 +1,255 @@
...
+                make_w3(),
+                [make_account(ACCOUNT_1_ADDRESS), make_account(ACCOUNT_1_ADDRESS)],
+                None,
+            )
+
+
+@pytest.mark.unit
+class TestProcessMembers:
+    def test_process_members__account_1_is_plain_member__activates_account_1(self):
+        account_1 = make_account(ACCOUNT_1_ADDRESS)
```
chasingrainbows, line 71: All single-key tests here configure the key in the first slot - SignerModule(w3, account_1, None, ...). There is no test where only account_2 is set: SignerModule(w3, None, account_2, ...). One such test would be useful.

src/web3py/extensions/signer.py
```diff
@@ -0,0 +1,139 @@
...
+            self.delegation_contract = cast(
+                DelegationContract,
+                self.w3.eth.contract(
+                    address=Web3.to_checksum_address(delegation_contract_address),
+                    ContractFactoryClass=DelegationContract,
+                    decode_tuples=True,
+                ),
+            )
+
+    def process_members(self, members: list[ChecksumAddress]) -> None:
```
chasingrainbows, line 66: The contract interface uses "delegate" (getDelegate / assignDelegate), but this docstring says "delegatee". Same in README.md line 257 - "a candidate delegatee of DELEGATION_CONTRACT_ADDRESS". Better to use one term everywhere.

hweawer (approved):

src/web3py/extensions/signer.py
F4ever: Don't think this require changing to old behaviour

src/web3py/extensions/tx_utils.py
F4ever: test added

src/variables.py
F4ever: fixed

src/modules/oracles/common/consensus.py
F4ever: restart will fix issue immediately. The error itself tells us enough to identify the issue. Not sure we need here: 
try: catch: raise

tests/web3py/test_signer_module.py
F4ever: this is outdated

src/web3py/extensions/signer.py
F4ever: fixed

chasingrainbows (approved):

PR #969, opened 2026-07-24
feat: replace validation lib with c compiled

F4ever:
Migrate from `py_ecc` to [blst](https://github.com/supranational/blst) library


.github/workflows/tests.yml
```diff
@@ -30,6 +32,12 @@ jobs:
         run: |
           poetry install --no-interaction --with=dev
 
+      - name: Install swig
+        run: sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends -qq swig
```
hweawer: swig pinned in Docker, not pinned in CI

tests/services/test_deposit_signature_verification.py
```diff
@@ -200,13 +211,116 @@ def test_is_valid_deposit_signature_with_explicit_genesis_validators_root():
 @pytest.mark.unit
 def test_is_valid_deposit_signature_different_fork_versions_produce_different_signing_roots():
     # Two calls with different fork versions must produce different signing roots
-    with patch('src.services.deposit_signature_verification.BLSVerifier') as mock_bls:
-        mock_bls.Verify.return_value = True
+    with patch('src.services.deposit_signature_verification.blst') as mock_blst:
```
hweawer: honestly dont see value in unit tests if we have real tests here

hweawer (dismissed):
LGTM, with nits

tests/services/test_deposit_signature_verification.py
F4ever: 100% - removed

.github/workflows/tests.yml
F4ever: Pinned

hweawer (dismissed):

PR #970, opened 2026-07-25
ORC-790 feat: log input fingerprints for report divergence debugging

hweawer:
Draft. Base is `feat/oracle-v8.1`.

## Why

When members submit different report hashes they disagree about an *input*. The two inputs big enough to hide a disagreement — the beacon state (2.3M validators on mainnet) and the Keys API used-key set (487k keys, a 216 MB response) — cannot be logged as-is. On 2026-07-25 that left a one-deposit divergence effectively unattributable after the fact.

This logs one digest per response instead. Two members compare a line each and learn which dependency they disagree about, without either operator sharing any data.

## What is logged

| `msg` | Covers |
|---|---|
| `Beacon state fingerprint.` | The whole `BeaconStateView` — validators, balances, slashings, the pending deposit queue, pending partial withdrawals and consolidations. With `state_root` and `slot`. |
| `Keys API used keys fingerprint.` | The whole `v1/keys?used=true` response. |
| `Keys API module operators keys fingerprint.` | The whole `v1/modules/{}/operators/keys?used=true` response — keys, module and operator records. |

**Equal digests mean both members read the same inputs. That is the whole claim.** The digest covers every field the oracle *parses*, in the order it arrived — a projection, not the response: `BeaconStateView` keeps nine fields of a `BeaconState` (a 956 MB body on mainnet) and `LidoKey` drops `vetted`, with `from_response` discarding the rest before the digest sees them. That is the right scope, since a report can only diverge over something it was computed from, but it is not a hash of the wire bytes and could not be — see below. It deliberately does not say *which* entry differs: naming one needs the data itself, and the data is still there — the Keys API instances still disagree today, and `_kapi_sanity_check_by_operator` on the base already names the operator whose key indexes are missing. Pre-staging a per-entry answer in the logs would cost orders of magnitude more volume every cycle.

Three log lines, under 200 bytes each. No blobs.

## One decision worth reviewing

**The digest covers the parsed response, not the bytes on the wire.** Measured, not assumed: the same state at slot 14921600 from the two nodes above came back as 956,168,**799** bytes and 956,168,**798** bytes, with different sha256. One byte, in the envelope — `"execution_optimistic":false` against `"execution_optimistic":true`, because one node's execution client was lagging. Hashing the raw body would report a divergence between two members holding identical state, every cycle, for as long as one node is behind. Across different clients you would additionally get key order, whitespace and optional-field differences. The oracle also never holds those bytes: at 956 MB the state is consumed as a stream.

The cost of parsing first is that the digest covers `BeaconStateView` — the fields the oracle actually reads — rather than the full `BeaconState`.

Within that, the response is fingerprinted exactly as it arrived, list order included. For the beacon state that is required — the pending-deposit filter keeps the first deposit seen per pubkey, so a reordering alone changes the report. It also means a Keys API instance serving the same rows in a different order reads as a divergence; that is a difference in what we were handed, and one contract is easier to reason about than a per-provider switch. JSON objects are keyed in sorted order, since unlike a list their field order carries no meaning.

## Also

- `total_amount_gwei` on `Get pending lido validators.` — half of what `clPendingBalanceGwei` is summed from.
- Dropped the duplicate `Get pending lido validators.` line in `accounting.py`, now covered with more detail in `lido_validators.py`.

## Verified against live mainnet

`scripts/fingerprint_e2e.py` checks the property the whole diagnostic rests on, which no unit test can: **two independent nodes handed the same slot must produce the same digest.** Run against mainnet slot 14921600 (2,329,639 validators, 38,666 pending deposits) on two Lighthouse nodes running *different versions* — v8.0.1 and v8.1.3:

| | digest | fetch | digest cost |
|---|---|---|---|
| `cl-3` (Lighthouse v8.1.3) | `0x9c8c2177…` | 53.9 s | 8.2 s |
| `cl-1` (Lighthouse v8.0.1) | `0x9c8c2177…` | 53.4 s | 8.0 s |
| Keys API, 487,299 keys | `0x7b907b04…` | 43.8 s | 1.2 s |
| Keys API, same call a minute later | `0x7b907b04…` | | |

Both consensus nodes agree, and two Keys API calls a minute apart agree — so the Keys API's row order is stable in practice, which is what makes the order-sensitive contract above safe.

One run caught a real case worth knowing about: asked for a slot beyond its head, a lagging node (`is_syncing`, 5392 slots behind) **silently served its own head state** rather than returning 404. The digests differed, and the `slot` field on the log line said why — 14925567 against 14921663. That is the diagnostic doing its job, and it is why `slot` is logged next to the digest.

## Cost

Once per report cycle each, against fetches of ~54 s and ~44 s. Bunker mode fetches two additional historical states and pays the state cost again for each; they are distinguishable by `state_root`.

The first implementation cost 16.6 s for the state; a fast path for lists of same-type records halved it to 8.2 s. The output is unchanged — both paths share one scalar encoder, two tests assert equality by disabling the fast path and re-digesting, and the full-state digest is identical before and after.

## Safety

Every addition is a diagnostic: `log_fingerprint` logs and continues, and never fails a report.

## Merge with the base

The base branch has since grown `Keys API response.` (the `elBlockSnapshot` per request) and `_kapi_sanity_check_by_operator`, which requires every key index in `[0, total_deposited_validators)` per operator and names the gap. Both supersede this branch's earlier key-set conservation checks, so those and the Keys API snapshot plumbing they needed are gone.

## Docs

A "Report divergence" section in `README.md` — which lines to compare, in which order, and what a differing digest does and does not tell you. No separate doc: at three log lines it would be one more thing to keep in sync with the messages.

## Test

`ruff` and `pyright` clean, unit suite passes.

- `tests/utils/test_fingerprint.py` parametrises over `dataclasses.fields(BeaconStateView)` and `fields(LidoKey)`, so a field added later without being covered by the digest fails the suite.
- `tests/utils/test_fingerprint_mainnet_fixtures.py` replays 2000-entry slices of the real mainnet responses above (767 KB committed) and pins their digests. Nothing else would catch an encoding change that is self-consistent but differs from the previous release — every digest would simply move together — and that change would silently stop two members on different oracle versions from being comparable. It needs no network, so it runs in the normal CI job.

CI's `tests` job is red on `test_get_validators` / `test_get_state_view`, which is not this branch: re-running `feat/oracle-v8.1`'s last green run unmodified fails the same test today. The shared hoodi node read-times-out on `/eth/v2/debug/beacon/states/` (10 s, 3 retries), and that request completes before any code here runs.


src/utils/fingerprint.py
```diff
@@ -0,0 +1,86 @@
...
+needs per-slice detail costing orders of magnitude more log volume than these three fields,
+for a case that has not yet occurred. How to read them: `docs/report-divergence-logs.md`.
+"""
+
+import logging
+from collections.abc import Iterable
+from dataclasses import dataclass
+from typing import Any
+
+from eth_hash.auto import keccak
```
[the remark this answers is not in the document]
hweawer: This isn't right — `eth_hash.auto.keccak` is a `Keccak256` instance, not a bare function, and it does expose `.new()`. On the pinned version:

```
eth-hash version : 0.7.1
keccak type      : <class 'eth_hash.main.Keccak256'>
has .new()       : True
streaming digest : 0x4e03657aea45a94fc7d47ba826c8d667c0d1e6e33a64a036ec44f58fa12d6c45
one-shot digest  : 0x4e03657aea45a94fc7d47ba826c8d667c0d1e6e33a64a036ec44f58fa12d6c45
match            : True
```

That value is the keccak-256 of `abc`, so the streaming path is both available and correct.

It is also exercised well beyond unit tests: `scripts/fingerprint_e2e.py` ran this code over a 956 MB mainnet beacon state and a 216 MB Keys API response, and `tests/utils/test_fingerprint_mainnet_fixtures.py` pins the resulting digests.

Importing PyCryptodome's keccak directly would be a step backwards — `eth_hash.auto` is what picks an available backend, and hardcoding one drops that fallback.

src/web3py/extensions/lido_validators.py
```diff
@@ -367,6 +418,78 @@ def _get_lido_validators_with_keys(self, blockstamp: BlockStamp) -> tuple[list[L
...
+            # Cache hit: same response the keys themselves came from.
+            snapshot = self.w3.kac.get_used_lido_keys_snapshot(blockstamp)
+
+            # `get_beacon_stat` is lru_cache(maxsize=1) on the block identifier, so a second
+            # block evicts the entry `_kapi_sanity_check` just cached. Skip when equal.
+            block = (
+                blockstamp.block_hash
+                if snapshot['blockNumber'] == blockstamp.block_number
+                else HexStr(snapshot['blockHash'])
+            )
```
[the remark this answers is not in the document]
hweawer: Fair point, but it is now moot: `_log_key_set_conservation` has been removed from the branch entirely.

The base branch (`feat/oracle-v8.1`) has since grown `_kapi_sanity_check_by_operator`, which does this better — it requires the Keys API to return every key index in `[0, total_deposited_validators)` per operator and names the missing indexes, rather than comparing aggregate counts across two blocks. Keeping a weaker duplicate of it here made no sense, so the check and the Keys API snapshot plumbing it needed are both gone.

The PR is now three fingerprint log lines and nothing else in that file.

src/modules/oracles/common/consensus.py
```diff
@@ -233,28 +233,28 @@ def get_blockstamp_for_report(self, last_finalized_blockstamp: BlockStamp) -> Re
         latest_blockstamp, member_info = self._get_latest_data()
 
         # Check if the contract is currently reportable
-        # if not self.is_contract_reportable(latest_blockstamp):
-        #     logger.info({'msg': 'Contract is not reportable.'})
-        #     return None
+        if not self.is_contract_reportable(latest_blockstamp):
+            logger.info({'msg': 'Contract is not reportable.'})
+            return None
```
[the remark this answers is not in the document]
hweawer: Agreed, and this is resolved — `src/modules/oracles/common/consensus.py` is no longer touched by this PR at all.

To be clear about what that hunk was: it was not a proposed behavioural change. The branch had temporarily commented out the reportability/finalization/deadline gating and pinned `ref_slot` to `14846399` in order to replay the 2026-07-25 divergence locally; the commit under review was *restoring* it. Merging `feat/oracle-v8.1` in took the file back to the base branch's version, so it is identical to base now and there is nothing to split out.

Current file list for the PR:

```
README.md
scripts/fingerprint_e2e.py
src/modules/oracles/accounting/accounting.py
src/providers/consensus/client.py
src/providers/keys/client.py
src/utils/fingerprint.py
src/web3py/extensions/lido_validators.py
tests/utils/fixtures/*.json.gz
tests/utils/test_fingerprint.py
tests/utils/test_fingerprint_mainnet_fixtures.py
```

src/providers/consensus/client.py
```diff
@@ -273,7 +274,9 @@ def get_state_view_no_cache(self, blockstamp: BlockStamp) -> BeaconStateView:
             else:
                 raise
 
-        return BeaconStateView.from_response(**data)
+        state = BeaconStateView.from_response(**data)
+        log_fingerprint(logger, 'Beacon state', state, state_root=blockstamp.state_root, slot=state.slot)
```
chasingrainbows, line 278: `log_fingerprint` sits in `get_state_view_no_cache`, so it also fires from `get_validators_no_cache` - `AbnormalClRebase._calculate_lido_normal_cl_rebase` calls it every accounting cycle for the last report state. So there are at least two `Beacon state fingerprint.` lines per cycle, not one as the docstring and README say, and each extra line costs ~8 s of digest. Maybe move the call into the cached `get_state_view`, or note in the README that lines must be matched by `slot` / `state_root`.

src/utils/fingerprint.py
```diff
@@ -0,0 +1,171 @@
...
+
+def log_fingerprint(logger: logging.Logger, subject: str, value: Any, **context: Any) -> None:
+    """Log a one-line digest of `value`.
+
+    Never raises: this is a diagnostic, and no report may fail over one.
+    """
+    try:
+        digest = digest_of(value)
+    except Exception as error:  # pylint: disable=broad-except
+        logger.warning({'msg': f'{subject} fingerprint failed.', 'error': repr(error)})
```
chasingrainbows: Maybe add `**context` to the warning too?
```suggestion
        logger.warning({'msg': f'{subject} fingerprint failed.', 'error': repr(error), **context})
```

src/utils/fingerprint.py
```diff
@@ -0,0 +1,171 @@
+"""One comparable digest per large response an oracle report is built from.
```
chasingrainbows, line 1: The module docstring is too long - the code documents itself, and a docstring like this helps neither people nor LLMs, it is just one more thing to keep in sync. Some points repeat (parsed vs wire bytes is stated twice), and the size/timing numbers will go stale. I would cut it down to a few lines: what the digest covers, what equal digests mean, and a pointer to the README section.

src/providers/keys/client.py
```diff
@@ -87,6 +88,7 @@ def get_used_lido_keys(self, blockstamp: BlockStamp) -> list[LidoKey]:
         """Docs: https://keys-api.lido.fi/api/static/index.html#/keys/KeysController_get"""
         data = [LidoKey.from_response(**x) for x in self._get_with_blockstamp(self.USED_KEYS, blockstamp)]
         self._check_used_keys(data)
+        log_fingerprint(logger, 'Keys API used keys', data, endpoint=self.USED_KEYS)
```
chasingrainbows, line 91: None of the three call sites is covered by a test - if a refactor drops a `log_fingerprint` call, CI stays green. Cheap to close: `patch('src.providers.keys.client.log_fingerprint')` (same for the consensus client) and `assert_called_once_with(...)` per call site.

tests/utils/test_fingerprint_mainnet_fixtures.py
```diff
@@ -0,0 +1,87 @@
+"""Golden-master digests over real mainnet responses, captured once and replayed offline.
```
chasingrainbows, line 1: Same here - the module docstring is too long for what it says. The one thing worth keeping is the warning that regenerating the fixtures is a breaking change for digest comparability; the rest restates what the tests and fixture names already show. I would keep it to two or three lines.

README.md
```diff
@@ -338,6 +338,35 @@ Check out our [delegation guide](docs/delegation.md) for setting up oracle with
 
 Check out our [alerting guide](docs/alerts.md) for Prometheus Alertmanager configuration examples.
 
+### Report divergence
```
chasingrainbows, line 341: Suggest moving this section to a separate file under `docs/` (e.g. `docs/report-divergence.md`) with a one-line link from the README. It is mostly a debug guide, and the README does not need to grow by this much for it.

tests/utils/test_fingerprint_mainnet_fixtures.py
chasingrainbows, line 1: Committing binary files to git is not great, and these gzips cannot be read from the sources - a reviewer cannot tell what is inside or whether the tests are actually valid. The tests only need a deterministic input and a pinned digest constant, so the input does not have to be a real mainnet slice. The same protection works with synthetic data built in code - the `_lido_key` / `_state` builders from `test_fingerprint.py`, scaled up:

```python
PINNED_STATE_DIGEST = '0x...'
PINNED_KEYS_DIGEST = '0x...'

def test_digest_of__synthetic_state__matches_pinned_digest():
    assert digest_of(_state(validators=2000, deposits=500)) == PINNED_STATE_DIGEST

def test_digest_of__synthetic_keys__matches_pinned_digest():
    assert digest_of([_lido_key(i) for i in range(2000)]) == PINNED_KEYS_DIGEST
```

If real values matter, a small plain-JSON slice (50-100 entries) would be readable and enough.

src/utils/fingerprint.py
hweawer: done

src/providers/consensus/client.py
hweawer: > Maybe move the call into the cached get_state_view

how does it change anything?
added in readme - obviously hash for the same slot cant be equal to hash of another

tests/utils/test_fingerprint_mainnet_fixtures.py
hweawer: I don't see a problem with storing gzips in repo but see a problem with the approach above - from_response bug = silence in tests

src/providers/keys/client.py
hweawer: Fair, and fixed in 62c1568 — one test per call site:

- `test_get_state_view_no_cache__state_fetched__logs_fingerprint_of_returned_state`
- `test_get_used_lido_keys__response__logs_fingerprint_of_returned_keys`
- `test_get_used_module_operators_keys__response__logs_fingerprint_of_returned_data`

I went with the log line rather than `patch(...)` + `assert_called_once_with`. Same coverage of "the call is still there", but it also survives a rename of the helper, and it asserts `line['digest'] == digest_of(<what the method returned>)` — so it fails not only on a dropped call but on a fingerprint taken over the wrong value. The consensus one mocks `session.get`, so it goes through the real parse path.

Checked they actually bite: deleting all three `log_fingerprint` calls gives 3 failed, restoring gives 25 passed.

README.md
hweawer: Agreed, done in 5cdf819 — `docs/report-divergence.md`, with a one-line link from the README and an entry in the docs list in `agents.md`.

You are right on convention and I was wrong to fold it in earlier: the neighbouring sections (`### Delegation`, `### Alerts`) are one-line pointers into `docs/`, and this one had grown to 34 lines. My stated reason — "one more file to keep in sync with the log messages" — does not hold up, since the guide goes stale identically wherever it lives, just less visibly in the README.

src/utils/fingerprint.py
hweawer: Cut to six lines in 5cdf819:

```python
"""One comparable digest per large response an oracle report is built from.

Equal digests mean two members read the same inputs — that is the whole claim. The digest
covers the parsed value, not the raw body: the state response envelope carries a per-node
`execution_optimistic`, so raw bytes differ between two correct members whenever one node's
execution client lags. Which lines to compare: `docs/report-divergence.md`.
"""
```

The repetition and the volumes were fair — both gone. I kept the parsed-vs-raw sentence, because that one does not follow from the code: hashing the response body is the obvious thing to do, and without the reason someone will "fix" it that way. Everything else moved to `docs/report-divergence.md`.

tests/utils/test_fingerprint_mainnet_fixtures.py
hweawer: Cut from 15 lines to six in 5cdf819:

```python
"""Golden-master digests over real mainnet responses, captured once and replayed offline.

Nothing else catches an encoding change that is self-consistent but differs from the previous
release, which would silently stop members on different versions from comparing. Repin only
alongside a deliberate encoding change. Captured from mainnet slot 14921600 with
`scripts/fingerprint_e2e.py`, each list truncated to its first 2000 entries.
"""
```

I kept the "why this file exists" sentence alongside the breaking-change warning, rather than the warning alone. Without it the file reads as a duplicate of `test_fingerprint.py` and invites deletion — the point that no other test catches an encoding change where every digest moves together is the least obvious thing here. The paragraph about the slices is gone.

chasingrainbows: `from_response` stays covered if the raw dicts are built in code and parsed the same way the tests do it now after `json.load`:

```python
keys = [LidoKey.from_response(**_raw_key(i)) for i in range(2000)]
state = BeaconStateView.from_response(**_raw_state(validators=2000))
```

The gzips also feed `from_response` with already-projected values, not the raw provider response, so the parsing coverage is the same.

Why I would not store gzips in the repo:
- they cannot be read in review or from the sources, so nothing in the repo shows what the pinned digests are actually computed from - the constants can only be trusted, not checked
- git history bloats - every regeneration stores a full new blob forever, and every clone carries all of them; a few files like this and the repo gets much heavier for no real reason

hweawer: Agreed on both counts, done in 68bf2a3 — plain JSON, 100 entries per list.

You were right that the gzips bought no extra parsing coverage: the slices hold already-projected values, so `from_response` sees the same shape either way. And my first pass at this only made your second bullet worse — plain JSON at 2000 entries was 2.8 MB, and GitHub will not render a 44k-line diff, so it would have bought the readability in name only.

At 100 entries it is 205 KB total (159 KB state, 46 KB keys), both render in the diff, and the pins can now be checked against what they are computed over instead of trusted. Digests repinned accordingly; no encoding changed.

One thing the trim did cost, so flagging it rather than leaving it to be discovered: `digest_of` flushes to keccak every `_CHUNK_SIZE = 4096` pieces, and 100 keys is 802 pieces — a fifth of one buffer, so the multi-flush path lost its only coverage (the state slice still spans 4.6 chunks, so it kept its own). Sizing the fixture around that would put us back where we started, so I tested it directly in `test_fingerprint.py` instead: patch `_CHUNK_SIZE` down to 4 over 500 keys, assert the digest does not move. Checked it bites — deleting the `hasher.update(...)` in the flush branch fails it. That is better than what the big fixture gave us, since it tests the boundary on purpose rather than incidentally via response size.

Also had to move the two mutation targets, since the old indexes are past the end now: `validators[1337]` → `[3]`, and `used_keys[999]` → `[42]`. Worth knowing that 42 was the obvious pick for the validator one but has `effective_balance: 0` in the first 100, so the downward mutation would have gone negative and failed parsing for the wrong reason.

chasingrainbows (approved):

PR #978, opened 2026-07-26
feat: tight kapi check

F4ever:
## Description
Tight KAPI check. Check kapi returned more or equal used keys per operator than deposited on ref slot

## How Has This Been Tested?
- [X] Local tests (e.g., `pytest`)

## Checklist
- [X] New tests added (if applicable)


src/web3py/extensions/lido_validators.py
```diff
@@ -385,6 +386,36 @@ def _kapi_sanity_check(self, keys_count_received: int, blockstamp: BlockStamp):
                 f'than amount of deposited validators ({stats.deposited_validators}) returned from Staking Router'
             )
 
+    def _kapi_sanity_check_by_operator(self, lido_keys: list[LidoKey], blockstamp: BlockStamp) -> None:
```
avsetsin, line 422: We could make it more strict so that the indexes returned by KAPI are also checked. Smth like:

```python
required = set(range(operator.total_deposited_validators))
received = {
    key.index
    for key in lido_keys
    if key.module_address == module_address
    and key.operator_index == operator.id
}

missing = required - received
if missing:
    raise CountOfKeysDiffersException(...)
```

src/web3py/extensions/lido_validators.py
avsetsin, line 1: Do we have a check against Lido validators on CL? That the number of validators (including pending ones) we see on CL equal the number of deposited keys

src/web3py/extensions/lido_validators.py
```diff
@@ -380,11 +386,93 @@ def _kapi_sanity_check(self, keys_count_received: int, blockstamp: BlockStamp):
         # Make sure that used keys fetched from Keys API are >= total number of
         # deposited validators from Staking Router.
         if keys_count_received < stats.deposited_validators:
+            logger.error(
+                {
+                    'msg': 'Keys API sanity check failed: fewer used keys than deposited validators. '
+                    'Check el_block_snapshot in the adjacent `Keys API response.`',
+                    'keys_count_received': keys_count_received,
+                    'deposited_validators': stats.deposited_validators,
+                }
```
dgusakov, line 415: How does this log extend the raised error below?

src/web3py/extensions/lido_validators.py
```diff
@@ -240,6 +240,10 @@ def get_active_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidat
             for lido_validator in lido_validators
         ]
         logger.info({'msg': 'Get active lido validators.', 'value': len(result)})
+
+        pending_validators = self.get_pending_lido_validators(blockstamp)
+        self._validate_total_validators_count(len(result), len(pending_validators), blockstamp)
```
dgusakov: Why this check is here and not when you actually call `get_active_lido_validators` and `get_pending_lido_validators`?

src/web3py/extensions/lido_validators.py
```diff
@@ -356,6 +360,8 @@ def _get_lido_validators_with_keys(self, blockstamp: BlockStamp) -> tuple[list[L
 
         validators = self.w3.cc.get_validators(blockstamp)
         self._kapi_sanity_check(len(lido_keys), blockstamp)
+        self._kapi_sanity_check_by_operator(lido_keys, blockstamp)
+        self._kapi_sanity_check_pending_deposits(lido_keys, blockstamp)
 
         lido_validators, pending_lido_keys = self.compute_lido_validators(lido_keys, validators)
```
avsetsin, line 386: here `pending_lido_keys` is the keys from the response from kapi that is not in the validator registry yet. so they could be in pending_deposits as pending validators, or not yet (if these deposits were done after ref slot).

now it could be misleading (can be treated as pending keys from pending deposits on cl on ref slot), consider a better naming

src/web3py/extensions/lido_validators.py
```diff
@@ -380,11 +386,93 @@ def _kapi_sanity_check(self, keys_count_received: int, blockstamp: BlockStamp):
...
+        if orphaned_pubkeys:
+            logger.warning(
+                {
+                    'msg': 'Pending deposits with Lido WC are not matched by any used key from Keys API.',
+                    'value': len(orphaned_pubkeys),
+                    'pubkeys': sorted(orphaned_pubkeys)[:10],
+                }
+            )
+
+    def _validate_total_validators_count(self, active_count: int, pending_count: int, blockstamp: BlockStamp) -> None:
```
mkurayan, line 482: Sorry i confused a little bit what active_count means? 

Because you compare  ` active_count + pending_count` with lido.deposited_validators which include all initial deposits ever made even if validator already exited.

mkurayan: Ahh it actually all validators active + exited. 
Maybe rename?

src/web3py/extensions/lido_validators.py
avsetsin, line 305: The name of the method is a bit misleading. It collects the deposits that will lead to create a lido validator based on KAPI pending keys (that has no validators in the registry yet). so top ups for existing validators in the registry technically are valid deposits (even with invalid signatures)

src/web3py/extensions/lido_validators.py
```diff
@@ -380,11 +386,93 @@ def _kapi_sanity_check(self, keys_count_received: int, blockstamp: BlockStamp):
...
+        """
+        Every deposited validator must be accounted for as either active (already visible on CL)
+        or pending (deposited but not yet processed by CL) — no more, no less.
+        """
+        stats = self.w3.lido_contracts.lido.get_beacon_stat(blockstamp.block_hash)
+        total_count = active_count + pending_count
+
+        if total_count != stats.deposited_validators:
```
avsetsin: Looks like there could be an edge case here

Oracle Frame X: let's say an attacker donate 32 ether and depositing on one of our key using valid signature from the module contract (or it could be a node operator, in this case he can deposit only 1 eth). In this case we will NOT unvet the key and we can deposit later with 32 ether (if i recall correctly how council daemon works)

Oracle Frame X + 1 (after refslot but before finalization): the protocol deposits on this key full initial deposit (32 eth). On the refslot on EL the deposited_validators are Y, but KAPI returns Y+1 and on CL we see Y+1, so the check will not pass

hweawer (approved):

src/web3py/extensions/lido_validators.py
F4ever: fixed

src/web3py/extensions/lido_validators.py
F4ever: Created a ticket for next release

src/web3py/extensions/lido_validators.py
F4ever: We call those methods in several places. To not duplicated code decided to store sanity check in the end of function we checking

src/web3py/extensions/lido_validators.py
F4ever: created a ticket

src/web3py/extensions/lido_validators.py
F4ever: created a ticket

PR #983, opened 2026-07-27
fix: auditor comments on the front-run and KAPI checks

hweawer:
## Description

Auditor comments on the front-run / KAPI checks that landed in #978 and #980.

Three items ship here; a fourth was implemented and then reverted on request — kept in the history and described below so the trade-off is on record.

### ~~1. `FrontRunAttackError` raised at the decision point, not in the helper~~ — reverted (`bb748dfc`)

Done in `cdcae935`, then reverted on request. `FrontRunAttackError` is raised from `_collect_valid_pending_deposits` again, as on `develop`.

Recorded so it is not rediscovered later: the historical caller `AbnormalClRebase._sum_valid_lido_pending` therefore raises too. It runs over `prev_blockstamp` to measure a rebase, so the bunker check fails on a front run anywhere in the previous frame — **including one already resolved by the reference slot**, where the validator has been created, the pubkey has left the not-yet-on-CL filter, and the ref-slot paths no longer see it. Reporting is then blocked over a closed incident.

The bunker must-not-raise test and the two helper tuple-contract tests went with the revert.

### 2. The log-only KAPI check runs before the raising ones (`8c4cac7f`)

`_kapi_sanity_check_pending_deposits` only warns; `_kapi_sanity_check` and `_kapi_sanity_check_by_operator` raise. Ordered behind them, its output was unreachable in exactly the incident where you want it — the run aborts and the log ends at the raise, with no record of which pending deposits sat on Lido credentials without a matching used key.

Demonstrated by restoring the old order under the new test: the log then contains only the `_kapi_sanity_check` error and nothing about pending deposits.

Cost is nil in practice — the check reads pending deposits and the WC list, and every caller of `_get_lido_validators_with_keys` needs the state view anyway, so the `lru_cache`d fetch is not duplicated. What changes is that a grossly stale Keys API no longer bails out before that read, which is the point: that read is the diagnostic.

### 3. Discarded-deposit test now asserts the refusal (`c34d8b7b`)

A test left deliberately red in #980 asked whether a deposit the CL threw away should block reporting. Decision: it should — the oracle does not write such a loss off on its own. The test now pins `CountOfKeysDiffersException` with the counts in the message, and the suite is green.

Deliberately `CountOfKeysDiffersException` and not `FrontRunAttackError`: nobody captured the ether here, so there is no counterparty to name and no key to escalate — only a count that no longer adds up. The two states look alike from the outside, so the test docstring says why they raise differently.

The consequence is recorded in `_validate_total_validators_count`: the refusal can be **permanent**. The CL drops a deposit whose signature does not verify (Electra `apply_pending_deposit`) without leaving a validator or a queue entry, `depositedValidators` never decreases, so the key is missing from both sets for good and every later frame raises too. Written down because it reads like a bug otherwise.

### 4. The private active/pending getters were running twice per report (`a5d930ef`)

`get_active_lido_validators` and `get_pending_lido_validators` are both `lru_cache`d, but each cross-checks active against pending, so each calls both private halves — and those were uncached. Accounting calls both wrappers, so every report ran `_get_active_lido_validators` and `_get_pending_lido_validators` twice, the latter BLS-verifying the whole Lido pending-deposit queue both times.

Measured on a read-only rebuild of the mainnet report at ref slot 14860799 (23,316 pending Lido deposits), warm state cache, one run each:

| | develop | with caches |
|---|---|---|
| `_get_active_lido_validators` runs | 2 | 1 |
| `_collect_valid_pending_deposits` runs | 2 | 1 |
| CL balance trace | 73.9 s | 52.4 s |
| full report build | 23.0 s | 23.5 s |
| **total** | **96.9 s** | **75.9 s** |

Report hash unchanged: `0x4c766af8b5843360d319a0f52879f8ab4f3941d73cdcdaa5e2a61db3a473e054` — the same hash the first oracle member submitted on chain for that slot.

Caveat on the timings: how the saving splits between the two phases depends on the debug tool's call order, so read the total rather than the split. The run counts are the tool-independent part.

No extra memory — the wrapper returns the very object the private method built, so the two `maxsize=1` entries point at the same list/dict. No behaviour change either: `maxsize=1` evicts on a different blockstamp exactly as the wrappers already did, which is what `safe_border`'s historical lookups rely on.

## How Has This Been Tested?

- [X] Local tests (`pytest`)
- [X] Manual testing — read-only mainnet report rebuild, see item 4
- [ ] Not tested

**1524 passed, 0 failed.** `ruff format`, `ruff check` and `pyright src` clean.

Each behavioural change is mutation-checked, so none of the new tests are vacuous:

| change | mutation | result |
|---|---|---|
| log-only check moved first | restore the old order | diagnostics missing from the log |
| discarded deposit refused | disable `_validate_total_validators_count` | `DID NOT RAISE` |

New tests:

- `test_get_lido_validators_with_keys__raising_check_fails__pending_deposit_diagnostics_still_logged` — pins the check order;
- `test_report__cl_discarded_the_deposit__reporting_is_refused` — replaces the write-off test.

## Checklist

- [X] New tests added (if applicable)

### DIFF
--- a/src/web3py/extensions/lido_validators.py
+++ b/src/web3py/extensions/lido_validators.py
@@ -545,9 +545,7 @@ class LidoValidatorsProvider(Module):
         an active CL validator before Lido's own deposit() call increments the ref-slot-pinned
         deposited_validators counter).
 
-        If the counts do not add up, we do not submit the report. That can be permanent: the CL drops
-        a deposit with an invalid signature without leaving a validator or a queue entry, and
-        deposited_validators never decreases. Governance accounts for lost ether, not the oracle.
+If the total number of deposits and validators on the CL side is lower than what we see on the EL side, the oracle will not produce a report because the data sources are out of sync.
         """
         stats = self.w3.lido_contracts.lido.get_beacon_stat(blockstamp.block_hash)
         total_count = active_count + pending_count
--- a/src/web3py/extensions/lido_validators.py
+++ b/src/web3py/extensions/lido_validators.py
@@ -191,7 +191,6 @@ type PendingValidator = tuple[LidoKey, list[PendingDeposit]]
 class LidoValidatorsProvider(Module):
     w3: Web3
 
-    @lru_cache(maxsize=1)
     def get_active_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
         """Both public getters cross-check active against pending, so both call both private
         halves. The caches on the private methods are what keeps that from doing the work twice --
--- a/src/web3py/extensions/lido_validators.py
+++ b/src/web3py/extensions/lido_validators.py
@@ -193,11 +193,17 @@ class LidoValidatorsProvider(Module):
 
     @lru_cache(maxsize=1)
     def get_active_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
+        """Both public getters cross-check active against pending, so both call both private
+        halves. The caches on the private methods are what keeps that from doing the work twice --
+        in particular `_get_pending_lido_validators` BLS-verifies the whole Lido pending-deposit
+        queue, which is the most expensive thing in the report.
+        """
         result = self._get_active_lido_validators(blockstamp)
         pending_validators = self._get_pending_lido_validators(blockstamp)
         self._validate_total_validators_count(len(result), len(pending_validators), blockstamp)
         return result
 
+    @lru_cache(maxsize=1)
     def _get_active_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
         pending_deposits = self.w3.cc.get_pending_deposits(blockstamp)
         deposits_by_pubkey: dict[str, list[PendingDeposit]] = {}
@@ -272,6 +278,7 @@ class LidoValidatorsProvider(Module):
         self._validate_total_validators_count(len(active_validators), len(pending_validators), blockstamp)
         return pending_validators
 
+    @lru_cache(maxsize=1)
     def _get_pending_lido_validators(
         self,
         blockstamp: BlockStamp,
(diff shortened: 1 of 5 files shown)


src/web3py/extensions/lido_validators.py
```diff
@@ -191,13 +191,18 @@ class FrontRunAttackError(Exception):
 class LidoValidatorsProvider(Module):
     w3: Web3
 
-    @lru_cache(maxsize=1)
     def get_active_lido_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
+        """Both public getters cross-check active against pending, so both call both private
+        halves. The caches on the private methods are what keeps that from doing the work twice --
+        in particular `_get_pending_lido_validators` BLS-verifies the whole Lido pending-deposit
+        queue, which is the most expensive thing in the report.
+        """
```
F4ever: let's remove it

src/web3py/extensions/lido_validators.py
```diff
@@ -539,6 +544,10 @@ def _validate_total_validators_count(self, active_count: int, pending_count: int
         Contract using one of Lido's vetted-but-not-yet-protocol-deposited keys — the key becomes
         an active CL validator before Lido's own deposit() call increments the ref-slot-pinned
         deposited_validators counter).
+
+        If the counts do not add up, we do not submit the report. That can be permanent: the CL drops
+        a deposit with an invalid signature without leaving a validator or a queue entry, and
+        deposited_validators never decreases. Governance accounts for lost ether, not the oracle.
```
F4ever:
```suggestion
If the total number of deposits and validators on the CL side is lower than what we see on the EL side, the oracle will not produce a report because the data sources are out of sync.
```

F4ever (dismissed):

F4ever (approved):

PR #989, opened 2026-07-28
feat: Build script for report

F4ever:


scripts/run_oracle_at_slot.py
```diff
@@ -0,0 +1,154 @@
...
+    args = parse_args()
+
+    prometheus_port = args.prometheus_port or find_free_port()
+    healthcheck_port = args.healthcheck_port or find_free_port()
+
+    # Must be set before `src.variables` (or anything importing it) is loaded,
+    # since those values are read from the environment at import time.
+    os.environ['DAEMON'] = 'False'
+    # Force dry mode (no on-chain tx signing/sending) regardless of the caller's shell env.
+    os.environ.pop('MEMBER_PRIV_KEY', None)
```
hweawer, line 93: nit: I found that our dry_mode is still sending telemetry transactions, so might need to pop TELEMETRY_PRIV_KEY/TELEMETRY_PRIV_KEY_FILE

hweawer (approved):
LGTM to me with minor comments - there are side effects. Our DRY_RUN excludes sending to IPFS and telemetry sending.

I propose to fix it - now collecting report like this can send both (and actually REQUIRES IPFS). But it might be in the scope of other PR.

PR #997, opened 2026-08-07
Feat/telemetry tx retry

F4ever:
## Description
Describe what this pull request does. Explain the purpose of the changes and the problem they solve.

## Related Issue/Task

## How Has This Been Tested?

## Checklist


src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +122,53 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
         payload = json.dumps(message, default=str).encode('utf-8')
 
         tx = self._contract.send_message(event_id.value, payload)
-        params = build_transaction_params(self._data_bus_w3, tx, variables.TELEMETRY_ACCOUNT)
-        tx_hash = sign_and_send_transaction(self._data_bus_w3, tx, params, variables.TELEMETRY_ACCOUNT)
+        try:
+            tx_hash = self._send_with_retry(tx, self._data_bus_w3, variables.TELEMETRY_ACCOUNT)
+        except SendTimeoutError:
```
hweawer: It shouldn't be caught, right? caller will not catch it and put incorrect ` self._last_sent_report_hash`

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -18,17 +22,24 @@
 
 logger = logging.getLogger(__name__)
 
+# Interval between transaction status polls while waiting for inclusion or a nonce change.
+_POLL_INTERVAL_SECONDS = 12
```
hweawer: nit: in general would be good to have random jitter, wdyt?

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +122,53 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
         self.update_telemetry_account_balance_metric()
+
+    def _send_with_retry(self, tx: ContractFunction, w3: Web3, account: LocalAccount) -> bytes:
+        deadline = time.monotonic() + variables.TELEMETRY_TX_SEND_TIMEOUT_SECONDS
+        attempt = 0
+        tx_hash: bytes | None = None
+        nonce: int | None = None
+
+        while time.monotonic() < deadline:
+            attempt += 1
```
hweawer: Maybe we can just assume that mempool inclusion passed? And then retry the sending itself without nonce checks?
something like this:
```
while True:
        attempt += 1
        try:
            params = build_transaction_params(w3, tx, account)
            return sign_and_send_transaction(w3, tx, params, account)
        except (ValueError, Web3Exception, RequestException) as error:
            delay = _retry_delay(attempt)
            if time.monotonic() + delay >= deadline:
                raise SendTimeoutError(f"Gave up sending DataBus telemetry after {attempt} attempts.") from error
            logger.warning({
                'msg': 'Failed to send DataBus telemetry transaction. Will retry.',
                'attempt': attempt,
                'retry_in_seconds': round(delay, 1),
                'error': str(error),
            })
            time.sleep(delay)
```

src/web3py/extensions/telemetry_data_bus.py
F4ever: Since transactions can only be included once every 12 seconds, I don’t see a reason to poll the node at random intervals.

src/web3py/extensions/telemetry_data_bus.py
F4ever: In this case, the node will throw an error that the gas price for the same nonce must be higher and won’t accept the transaction at all.

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +121,50 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+            try:
+                if tx_hash is not None:
+                    sent_tx = w3.eth.get_transaction(HexBytes(tx_hash))
```
chasingrainbows: If the sent tx is dropped from the mempool, `get_transaction` raises `TransactionNotFound` instead of returning a pending tx with `blockNumber: None`. The `except` below only logs and sleeps, and `tx_hash` is not reset - so the loop keeps polling the same dead hash until the deadline and never re-sends. Consider catching `TransactionNotFound` around this call and resetting `tx_hash`/`nonce`, so the next iteration rebuilds and re-sends the transaction.

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +121,50 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+                        return tx_hash
+
+                    current_nonce = w3.eth.get_transaction_count(account.address)
+                    if current_nonce == nonce:
+                        remaining = deadline - time.monotonic()
+                        time.sleep(min(_POLL_INTERVAL_SECONDS, remaining))
+                        continue
+
+                params = build_transaction_params(w3, tx, account)
+                nonce = params.get('nonce')
```
chasingrainbows: `nonce` is updated before `sign_and_send_transaction` completes. If the send fails after `build_transaction_params` succeeded, `nonce` already holds the new value while `tx_hash` still points to the previous tx. On the next iteration `current_nonce == nonce` compares the on-chain nonce with a tx that was never sent, so the loop waits for the old tx (which can no longer be mined) until the deadline instead of retrying the send. Consider assigning `nonce` together with `tx_hash` only after `sign_and_send_transaction` returns.

src/variables.py
```diff
@@ -154,6 +154,7 @@
 TELEMETRY_DATA_BUS_RPC: Final = os.getenv('TELEMETRY_DATA_BUS_RPC', '')
 DATA_BUS_ADDRESS: Final = os.getenv('DATA_BUS_ADDRESS', '0x37De961D6bb5865867aDd416be07189D2Dd960e6')
 TELEMETRY_DIAGNOSTIC_INTERVAL_SECONDS: Final = int(os.getenv('TELEMETRY_DIAGNOSTIC_INTERVAL_SECONDS', 3600))
+TELEMETRY_TX_SEND_TIMEOUT_SECONDS: Final = int(os.getenv('TELEMETRY_TX_SEND_TIMEOUT_SECONDS', 4 * 60))
```
chasingrainbows, line 157: `send_telemetry` runs inside `process_report`'s `finally` block, so on RPC trouble it blocks the main cycle for up to `TELEMETRY_TX_SEND_TIMEOUT_SECONDS` (4 min by default). Consider reducing the timeout to ~30 seconds, or removing the env variable and keeping it as a constant - so an operator cannot set a value that breaks reporting (`MAX_CYCLE_LIFETIME_IN_SECONDS`).

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +121,50 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+                        'attempt': attempt,
+                        'remaining_seconds': max(remaining, 0),
+                        'error': str(error),
+                    }
+                )
+                if remaining <= 0:
+                    break
+                time.sleep(min(_POLL_INTERVAL_SECONDS, remaining))
+
+        raise SendTimeoutError(
```
chasingrainbows: Consider `raise SendTimeoutError(...) from error` (last caught exception) - otherwise the root cause is lost in the callers' logs.

README.md
```diff
@@ -321,6 +321,7 @@ In manual mode all sleeps are disabled and `ALLOW_REPORTING_IN_BUNKER_MODE` is T
...
 | `TELEMETRY_DATA_BUS_RPC`                                 | EL RPC endpoint for the DataBus telemetry contract                                                                                                                          | False               | `http://localhost:8545`                      |
 | `DATA_BUS_ADDRESS`                                       | Address of the DataBus contract for telemetry reporting                                                                                                                     | False               | `0x37De961D6bb5865867aDd416be07189D2Dd960e6` |
+| `TELEMETRY_TX_SEND_TIMEOUT_SECONDS`                      | Total time budget to retry sending a telemetry transaction (refetching nonce and fee each attempt) before giving up. Default 120                                            | False               | `240`                                        |
```
chasingrainbows: The description says "Default 120", but the actual default is 240 (`4 * 60` in `variables.py`).

tests/web3py/test_telemetry_data_bus.py
```diff
@@ -117,6 +128,324 @@ def test_send_telemetry__configured__sends_transaction(
         assert 'DataBus telemetry sent.' in caplog.text
         mock_data_bus_w3.eth.get_balance.assert_called_once_with(variables.TELEMETRY_ACCOUNT.address)
 
+    @patch('src.web3py.extensions.telemetry_data_bus.time.sleep')
+    @patch('src.web3py.extensions.telemetry_data_bus.sign_and_send_transaction')
+    @patch('src.web3py.extensions.telemetry_data_bus.build_transaction_params')
+    @patch.object(TelemetryDataBus, '_validate')
+    @patch.object(TelemetryDataBus, '_create_web3')
+    def test_send_telemetry__transient_failure__retries_and_succeeds(
```
chasingrainbows: These 5 retry tests at the `send_telemetry` level look like 1:1 duplicates of the `_send_with_retry` tests below. The wiring is already covered by `test_send_telemetry__configured__sends_transaction`, so they could probably be removed (~180 lines) without losing coverage. Could you check?

chasingrainbows (changes requested):

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +117,35 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+    ) -> bytes:
+        start_time = time.monotonic()
+
+        while True:
+            try:
+                params = build_transaction_params(w3, tx, account)
+                tx_hash = sign_and_send_transaction(w3, tx, params, account)
+            except Exception as error:
+                logger.warning({'msg': 'DataBus telemetry send failed.', 'error': error, 'module': self._module_name})
+                continue
```
chasingrainbows: This `continue` jumps back to `while True` and skips both the deadline check and the sleep. If `build_transaction_params` or `sign_and_send_transaction` fails on every attempt (RPC down, insufficient funds, replacement underpriced), the loop ignores `timeout` and spins forever without backoff - and since this runs in the main thread, startup or the report cycle would hang. Could you add a unit test that proves the loop cannot run forever - e.g. persistent exceptions on every attempt with a small timeout, asserting `_send_with_retry` still returns?

chasingrainbows: Replace `while` with `for _ in range(RETRY_ATTEMPTS)`.

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +117,35 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+            else:
+                sent_tx = w3.eth.get_transaction(HexBytes(tx_hash))
```
chasingrainbows: `get_transaction` sits in the `else` clause, outside the `try` above, so `TransactionNotFound` is not caught here. Right after broadcast the node may not have indexed the tx yet (common with load-balanced RPC), and this exception would abort the whole retry loop even though the tx was already sent. Maybe wrap this call too and treat `TransactionNotFound` as "still pending"?

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +117,35 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+                if sent_tx.get('blockNumber') is not None:
+                    return tx_hash
+
+            if start_time + timeout < time.monotonic():
+                return tx_hash
```
chasingrainbows: On timeout this returns the same `tx_hash` as the mined case, so the caller cannot tell "included in a block" from "gave up while still pending" - it logs `DataBus telemetry sent.` and `_try_send_telemetry` marks the report hash as sent either way. Maybe log a different message here (or return None) so a stuck tx is visible in logs?

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +117,35 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+        timeout: float,
+    ) -> bytes:
+        start_time = time.monotonic()
+
+        while True:
+            try:
+                params = build_transaction_params(w3, tx, account)
+                tx_hash = sign_and_send_transaction(w3, tx, params, account)
+            except Exception as error:
+                logger.warning({'msg': 'DataBus telemetry send failed.', 'error': error, 'module': self._module_name})
```
chasingrainbows: `error` here is the raw exception object. `JsonFormatter` calls `json.dumps` without `default=str`, so this line raises `TypeError` inside the formatter and the record is lost (logging prints a raw traceback to stderr instead). `str(error)` would fix it - same pattern as in `src/utils/transaction.py`.

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +117,40 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+            else:
+                sent_tx = w3.eth.get_transaction(HexBytes(tx_hash))
+                if sent_tx.get('blockNumber') is not None:
+                    return tx_hash
+
+            if start_time + timeout < time.monotonic():
+                if tx_hash:
+                    return tx_hash
```
chasingrainbows, line 155: When the timeout expires with the tx sent but still pending, this returns the same `tx_hash` as the mined case - the caller logs `DataBus telemetry sent.` and `_try_send_telemetry` marks the report hash as sent, even though the tx may never land. Maybe log a different message here so a stuck tx is visible in logs?

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +117,40 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+                else:
+                    raise Exception('DataBus telemetry send failed.') from error
```
chasingrainbows: Python deletes the `except ... as error` name when the except block exits, so the `error = None` init above does not help - when every attempt fails and the timeout expires, this line raises `UnboundLocalError: cannot access local variable 'error'` instead of the intended exception (reproduced by calling `_send_with_retry` with a failing `build_transaction_params`). Storing it under a different name, e.g. `except Exception as exc:` + `last_error = exc`, would fix it.

chasingrainbows (approved):

PR #999, opened 2026-08-10
fix: correct the swept-rewards and deposit-reserve terms in the ejector's balance estimate

### LINKED ISSUE #993: CL rewards double accounting in ejector
## Summary

`Ejector._get_predicted_el_balance` may double-count validator sweep amounts against the predicted rewards rate, causing `predictable_el_balance` to be overestimated. In practice this appears to be mitigated by the deposit lock deduction in the same formula.

## Where

`src/modules/oracles/ejector/ejector.py`, `_get_predicted_el_balance` (currently lines 218-269):

```python
rewards_speed_per_epoch = self.prediction_service.get_rewards_per_epoch(blockstamp, chain_config)
...
future_rewards = time_to_last_withdrawal_in_epoch * rewards_speed_per_epoch
...
future_withdrawals = self._get_withdrawable_lido_validators_balance(withdrawal_epoch, blockstamp)
...
return Wei(
    future_rewards
    + future_withdrawals
    + total_available_balance
    + gwei_to_wei(going_to_withdraw_balance_gwei)
    - deposit_lock,
)
```

`rewards_speed_per_epoch` comes from `RewardsPredictionService.get_rewards_per_epoch` (`src/services/prediction.py:28-80`), which derives a per-epoch rate from historical `TokenRebased`/`ETHDistributed` events:

```python
total_rewards += (
    event['postCLBalance']
    + event['withdrawalsWithdrawn']
    - event['preCLBalance']
    + event['executionLayerRewardsWithdrawn']
)
```

## Issue

`rewards_speed_per_epoch` is derived by adding back `withdrawalsWithdrawn` (and EL rewards) on top of the net CL balance change, so it already reflects the combined historical flow of new CL rewards **and** validator sweep withdrawals into the Withdrawal Vault. When this rate is projected forward (`future_rewards = time_to_last_withdrawal_in_epoch * rewards_speed_per_epoch`), it already represents expected future EL inflow from the ongoing sweep process.

`future_withdrawals` then separately sums, for every active Lido validator, the currently unswept balance above the effective-balance cap (`get_predictable_inbound_sweep`) or full balance for validators about to become withdrawable (`_get_withdrawable_lido_validators_balance`, `src/modules/oracles/ejector/ejector.py:271-286`), and adds it on top of `future_rewards`.

Since `rewards_speed_per_epoch` already carries a sweep component, adding the explicit `future_withdrawals` sweep sum on top of `future_rewards` risks accounting for the same expected sweep inflow twice, inflating `predictable_el_balance`. An inflated `predictable_el_balance` would cause the ejector to under-select validators for exit relative to the actual ETH needed to cover the withdrawal queue (`get_validators_to_eject`, `src/modules/oracles/ejector/ejector.py:168-216`).

## Mitigating factor

The deposit lock deduction (`_get_deposit_lock_amount`, `src/modules/oracles/ejector/ejector.py:352-381`, subtracted at line 268) currently offsets this. Deposit lock reserved per day is larger than rewards accrued per day, so in practice the `- deposit_lock` term outweighs the potential over-count from `future_rewards` + `future_withdrawals`, keeping `predictable_el_balance` conservative rather than inflated. This masks the double-accounting from having a real operational impact today, but the formula should be reviewed/decoupled from this offsetting effect, since deposit lock and reward prediction are unrelated in intent and the mitigation could break if either side of the formula changes independently (e.g. deposit lock parameters change, or the rewards prediction window/methodology changes).

hweawer:
Fixes #993.

Two terms of `_get_predicted_el_balance` were wrong in opposite directions.

```diff
  future_rewards
- + future_withdrawals      # capped balance of exiting validators + everything above the cap
+ + withdrawable_principal  # capped balance of exiting validators only
  + total_available_balance
  + going_to_withdraw
- - deposit_lock
+ - deposit_lock // 2
```

## 1. Swept rewards were counted twice

`future_withdrawals` added the balance validators hold above their effective-balance cap, but
`future_rewards` already covers it: over a horizon `H` the sweep reaches `H/cycle` of the set and
each validator it reaches hands over a full cycle of accrual, summing to `rate * H`. Verified on
mainnet state at ref slot 14924159 — the measured pile of 2 796 ETH over 273 472 Lido validators
equals `rate * epochs_to_sweep` to within 0.2%, where the rate comes independently from TVL and APR.

It is not kept as a lower bound either: a pile is a stock and the term is a flow. It also would not
help where it looks like it should — a negative rebase only zeroes the rate once losses outrun the
rewards of the whole 7-day prediction window, and a milder one leaves `rate * H` larger anyway.

Direction: the old formula over-estimated available ETH, so the ejector requested **too few** exits.

## 2. The deposit reserve was charged in full

The reserve is refilled by new stake, and that inflow is not a term in this formula at all, so
charging the full lock bills the withdrawal queue twice for the same ETH.

Measured on mainnet, 360 days / 361 AO frames / 118 255 `Submitted` events, against
`reserve_per_frame = min(getDepositsReserveTarget()=1500, MAX_STETH_WITHDRAWAL_AMOUNT=1000) = 1000 ETH`:

| daily new stake | min | p1 | p5 | median |
|---|---|---|---|---|
| ETH | 823 | 1 132 | 1 553 | 7 548 |

New stake covered the reserve in **360 of 361 frames** (99.95%). `// 2` corresponds to a daily stake
of 500 ETH — 1.65x below the worst day observed in a year, so it is a deliberately conservative
choice rather than a fitted one. **Open for review:** the data supports a smaller divisor or dropping
the term; `// 2` is a judgment call and I have no objection to changing it.

Direction: opposite to (1) — this raises the estimate.

## Net effect

At ref slot 14924159 the two changes largely offset: −2 796 ETH from (1), +2 000 ETH from (2), net
−797 ETH against an unfinalized queue of 8 458 ETH. Exit counts need a re-run to state precisely.

## Consensus-breaking

The report hash changes on any frame where either term was binding. Quorum is 5 of 9 — needs a
coordinated release, not a rolling upgrade.

## Testing

Regression tests pin both decisions and were mutation-checked (reintroducing the pile, and
`// 1`, `// 3`, `* 0` for the divisor — all caught). Unit suite green, `ruff` clean.

### DIFF
--- a/src/modules/oracles/ejector/ejector.py
+++ b/src/modules/oracles/ejector/ejector.py
@@ -253,57 +253,18 @@ class Ejector(OracleModule[Web3]):
         future_rewards = time_to_last_withdrawal_in_epoch * rewards_speed_per_epoch
         logger.info({'msg': 'Calculate future rewards.', 'value': future_rewards})
 
-        # The balance validators hold above their effective-balance cap — rewards earned but not yet
-        # swept — is deliberately NOT a term here, neither added nor used as a floor.
-        #
-        # `future_rewards` already covers it exactly. Over a horizon H, the sweep reaches a fraction
-        # H/cycle of the set, and each validator it reaches hands over a whole cycle's worth of
-        # accrual (what it was holding, plus what it accrues while waiting its turn). Those cancel to
-        # rate * H regardless of where each validator sits in the sweep order — so in steady state
-        # rate * H is the exact inflow, not an approximation. Adding the unswept pile on top counts
-        # the same ETH twice, which is what issue #993 was about.
-        #
-        # It is not kept as a lower bound either: a pile of ETH is a stock and this term is a flow,
-        # so `max()` between them compares different units. It also would not have helped where it
-        # looks like it should. A negative rebase only zeroes the rate when losses outrun the rewards
-        # of the whole prediction window; a milder one leaves `rate * H` the larger of the two, so
-        # `max()` would keep picking the stale event-based average and discard the fresh measurement.
-        #
-        # Consequence: when the rate prediction does degenerate — no `ETHDistributed` in the window,
-        # or losses large enough to trip the clamp in `RewardsPredictionService` — the estimated
-        # inflow is 0 instead of the pile's size. The estimate then reads low and the ejector requests
-        # more exits than needed. Those exits are irreversible, unlike the delay in the opposite
-        # direction, so the degenerate cases belong in the rate prediction, not behind a floor here.
+        # The balance above the cap is already inside `future_rewards`: the sweep pays H/cycle of the
+        # set a full cycle of accrual each, i.e. rate * H. Adding it double counts (#993); a floor
+        # would compare a stock to a flow. So a degenerate rate over-ejects — fix that in prediction.
         withdrawable_principal = self._get_withdrawable_principal(withdrawal_epoch, blockstamp)
         logger.info({'msg': 'Calculate withdrawable principal.', 'value': withdrawable_principal})
 
         deposit_lock = self._get_deposit_lock_amount(time_to_last_withdrawal_in_epoch, blockstamp)
         logger.info({'msg': 'Calculate deposit lock.', 'value': deposit_lock})
 
-        # Only half of the deposit reserve is charged against the withdrawal queue.
-        #
-        # Why less than all of it: the reserve is refilled by new stake, and fresh deposits are not a
-        # term in this formula at all. While deposits keep up with `getDepositsReserveTarget()` — the
-        # normal case — the reserve is funded out of that inflow, not out of ETH the queue was
-        # counting on. Subtracting the whole lock while omitting the inflow that pays for it charges
-        # the queue twice for the same ETH and biases the estimate low, which over-ejects.
-        #
-        # Why not none of it: "deposits cover the reserve" is an empirical expectation about demand,
-        # not an invariant, and it fails exactly when it matters — stETH demand collapsing, or the
-        # staking router paused. Then the reserve really does come out of the buffer. Halving keeps a
-        # margin for that case instead of betting the whole term on the assumption holding.
-        #
-        # The 1/2 is a deliberate judgment call, not a derived figure: it splits the difference
-        # between modelling both flows and modelling neither. Deposit-inflow telemetry is what would
-        # justify moving it — toward 0 if inflow reliably covers the reserve, back toward the full
-        # lock if it does not.
-        #
-        # Consequence versus charging the full lock: the estimate reads higher by `deposit_lock // 2`,
-        # so the loop in `get_validators_to_eject` stops earlier and requests fewer exits. That is the
-        # under-ejection direction — withdrawal requests stay unfinalized longer. It is a delay, not a
-        # loss: no funds are at risk, and the next frame (~5h) re-estimates from fresh state and ejects
-        # the shortfall once the buffer actually drains. The error is bounded by half of what the lock
-        # contributes, i.e. it grows with the withdrawal horizon and is capped per accounting frame.
+        # Half: new stake refills the reserve, and that inflow is not a term here, so the full lock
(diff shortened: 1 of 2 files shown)


src/modules/oracles/ejector/ejector.py
```diff
@@ -254,34 +253,47 @@ def _get_predicted_el_balance(self, to_exit_gwei: Gwei, blockstamp: ReferenceBlo
...
+        withdrawable_principal = self._get_withdrawable_principal(withdrawal_epoch, blockstamp)
+        logger.info({'msg': 'Calculate withdrawable principal.', 'value': withdrawable_principal})
 
         deposit_lock = self._get_deposit_lock_amount(time_to_last_withdrawal_in_epoch, blockstamp)
         logger.info({'msg': 'Calculate deposit lock.', 'value': deposit_lock})
 
+        # Half: new stake refills the reserve, and that inflow is not a term here, so the full lock
+        # bills the queue twice. Charging none would bet on stETH demand, which fails when it matters.
+        # The 1/2 is a judgment call. Costs at most `lock // 2` of under-ejection, corrected next frame.
+        charged_deposit_lock = deposit_lock // 2
```
F4ever: move to `_get_deposit_lock_amount`

src/modules/oracles/ejector/ejector.py
```diff
@@ -254,34 +253,47 @@ def _get_predicted_el_balance(self, to_exit_gwei: Gwei, blockstamp: ReferenceBlo
         future_rewards = time_to_last_withdrawal_in_epoch * rewards_speed_per_epoch
         logger.info({'msg': 'Calculate future rewards.', 'value': future_rewards})
 
-        future_withdrawals = self._get_withdrawable_lido_validators_balance(withdrawal_epoch, blockstamp)
-        logger.info({'msg': 'Calculate future withdrawals sum.', 'value': future_withdrawals})
+        # The balance above the cap is already inside `future_rewards`: the sweep pays H/cycle of the
```
F4ever: Move details to `_get_withdrawable_principal` method

src/modules/oracles/ejector/ejector.py
```diff
@@ -254,34 +253,47 @@ def _get_predicted_el_balance(self, to_exit_gwei: Gwei, blockstamp: ReferenceBlo
...
             + gwei_to_wei(going_to_withdraw_balance_gwei)
-            - deposit_lock,
+            - charged_deposit_lock,
         )
 
     @lru_cache(maxsize=1)
-    def _get_withdrawable_lido_validators_balance(self, on_epoch: EpochNumber, blockstamp: BlockStamp) -> Wei:
-        lido_validators = self.w3.lido_validators.get_active_lido_validators(blockstamp=blockstamp)
+    def _sweepable_validators(self, blockstamp: BlockStamp) -> list[LidoValidator]:
+        """Active Lido validators the sweep pays out. A consolidation source pays its target, not the vault."""
```
F4ever: remove this

src/modules/oracles/ejector/ejector.py
hweawer: Done in d670854f — moved into the `_get_withdrawable_principal` docstring.

src/modules/oracles/ejector/ejector.py
hweawer: Done in d670854f — the halving now happens inside `_get_deposit_lock_amount` (`ao_frames * reserve_per_frame // 2`) with the rationale in its docstring, so `_get_predicted_el_balance` is just the sum of terms again. The regression tests that pin the `1/2` moved down with it (`test_get_deposit_lock_amount__reserve_locked__charges_half`); the caller-side test now only pins that the charge is subtracted once.

One naming caveat: the method now returns half of the actual on-chain lock, so the name reads slightly wider than what it computes. The docstring leads with that. Happy to rename it to something like `_get_charged_deposit_lock` if you prefer.

src/modules/oracles/ejector/ejector.py
hweawer: Removed in d670854f.

F4ever (approved):

PR #1002, opened 2026-08-12
Feat: Telemetry retry

F4ever:
## Description
Describe what this pull request does. Explain the purpose of the changes and the problem they solve.

## Related Issue/Task

## How Has This Been Tested?

## Checklist


tests/web3py/test_telemetry_data_bus.py
```diff
@@ -117,6 +128,141 @@ def test_send_telemetry__configured__sends_transaction(
...
+        assert mock_build_params.call_count == 2
+        assert mock_sign_and_send.call_count == 2
+        assert w3_mock.eth.get_transaction.call_count == 2
+        w3_mock.eth.get_transaction_count.assert_called_once_with(account.address)
+        mock_sleep.assert_not_called()
+
+    @patch('src.web3py.extensions.telemetry_data_bus.time.sleep')
+    @patch('src.web3py.extensions.telemetry_data_bus.sign_and_send_transaction')
+    @patch('src.web3py.extensions.telemetry_data_bus.build_transaction_params')
+    def test__send_with_retry__build_params_fails_then_succeeds__retries_and_returns_tx_hash(
```
chasingrainbows, line 219: Test coverage gaps in the new `_send_with_retry` tests:

- No test where `get_transaction` / `get_transaction_count` raises (e.g. `TransactionNotFound`) while `tx_hash` is already set. All exception tests fail only in `build_transaction_params` before the first send. This is the most important gap: with a dropped tx the loop never re-sends and returns the stale hash as success.
- No test where `sign_and_send_transaction` raises after `build_transaction_params` succeeds.
- The `if remaining <= 0: break` branch is never exercised - all tests exit via the `while` condition.
- `test_send_telemetry__configured__sends_transaction` does not assert the return value, so dropping `return tx_hash` would pass all tests.

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -18,17 +22,25 @@
...
 
 
-class TelemetryDataBus(Module):
-    class ContractNotDeployedError(Exception):
-        pass
+class ContractNotDeployedError(Exception):
+    pass
+
+
+class SendTimeoutError(Exception):
```
chasingrainbows: After the move to module level these names look too generic: `SendTimeoutError` is easy to confuse with the builtin `TimeoutError`, and `ContractNotDeployedError` could fit any contract validation code elsewhere. Consider domain-prefixed names, e.g. `TelemetrySendTimeoutError` / `DataBusContractNotDeployedError` - it is cheap to rename now while only tests import them.

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +121,52 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+
+        while time.monotonic() < deadline:
+            attempt += 1
+            try:
+                if tx_hash is not None:
+                    sent_tx = w3.eth.get_transaction(HexBytes(tx_hash))
+                    if sent_tx.get('blockNumber') is not None:
+                        return tx_hash
+
+                    current_nonce = w3.eth.get_transaction_count(account.address)
```
chasingrainbows: There is a race between `get_transaction` and `get_transaction_count`: if the tx gets mined between these two calls, `current_nonce` is already incremented, the `current_nonce == nonce` check fails, and the loop re-sends a second tx with the same payload. Both txs get mined - duplicate telemetry event. The window reopens on every poll iteration.

Possible fix: when `current_nonce != nonce`, re-check `get_transaction(tx_hash)` before re-sending and return the hash if the tx is already included.

docker-compose.yml
```diff
@@ -121,6 +121,8 @@ services:
     environment:
       MEMBER_PRIV_KEY: ${MEMBER_PRIV_KEY:-}
       MEMBER_PRIV_KEY_FILE: ${MEMBER_PRIV_KEY_FILE:-}
+      MEMBER_PRIV_KEY_2: ${MEMBER_PRIV_KEY_2:-}
+      MEMBER_PRIV_KEY_2_FILE: ${MEMBER_PRIV_KEY_FILE:-}
```
chasingrainbows: `MEMBER_PRIV_KEY_2_FILE` is mapped to the host's `MEMBER_PRIV_KEY_FILE` instead of `MEMBER_PRIV_KEY_2_FILE` (same in the other 3 services). Since `from_file_or_env` prefers the `_FILE` variant, `ACCOUNT_2` would silently load the same key as `ACCOUNT` when key files are used, even if `MEMBER_PRIV_KEY_2` is set explicitly.

Suggested fix:
```yaml
MEMBER_PRIV_KEY_2_FILE: ${MEMBER_PRIV_KEY_2_FILE:-}
```

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +123,52 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+                    time.sleep(_POLL_INTERVAL_SECONDS)
+                    continue
+
+                # Check if this is telemetry transaction
+                if tx_hash:
+                    sent_tx = w3.eth.get_transaction(HexBytes(tx_hash))
+                    if sent_tx.get('blockNumber') is not None:
+                        return tx_hash
+
+                nonce = params.get('nonce')
```
chasingrainbows: `nonce` is updated before `sign_and_send_transaction` is called. If a resend attempt fails (after a first tx was already sent), `nonce` keeps the new value while `tx_hash` still points to the old unconfirmed tx. On the next iterations the `params.get('nonce') == nonce` shortcut fires, so the loop never re-checks inclusion and never retries the send - it sleeps until the deadline and returns the stale hash.

The existing failure tests only cover a failure on the first attempt (`tx_hash` is still `None`), so this path is untested. It would be good to add a test for the sequence: send succeeds - nonce advances - resend raises - loop goes through the shortcut until deadline. One possible fix is to update `nonce` only after `sign_and_send_transaction` succeeds.

docker-compose.yml
F4ever: fixed

src/web3py/extensions/telemetry_data_bus.py
F4ever: fixed

src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -111,8 +123,52 @@ def send_telemetry(self, event_id: TelemetryEventId, data: dict | None = None) -
...
+                        'error': str(error),
+                    }
+                )
+
+            time.sleep(_POLL_INTERVAL_SECONDS)
+
+        if tx_hash:
+            return tx_hash
+
+        raise TelemetrySendTimeoutError(
```
chasingrainbows, line 172: When the deadline is reached, the original error from the last attempt is lost - it only goes to the warning log as a string. Consider keeping the last caught exception and either re-raising it, or chaining it via `raise TelemetrySendTimeoutError(...) from last_error`. Callers log this with `exc_info=True`, so the chained traceback would show the root cause of the failure.

chasingrainbows (approved):

PR #1006, opened 2026-08-14
Feat: fix zero available eth error

F4ever:


hweawer (approved):

tests/modules/accounting/test_withdrawal_unit.py
```diff
@@ -92,3 +92,16 @@ def test_calculate_finalization_batches(subject: Withdrawal, past_blockstamp):
     result = subject._calculate_finalization_batches(1, SHARE_RATE_PRECISION_E27, past_blockstamp.block_timestamp)
 
     assert result == [2]
+
+
+@pytest.mark.unit
+def test__calculate_finalization_batches__available_eth_zero__does_not_call_contract(
```
chasingrainbows, line 98: The new test checks the case when available_eth is 0 before the loop starts. But the main case for this fix - when the contract returns remaining_eth_budget=0 with finished=False in the middle of the loop - has no test. Maybe add one more test: mock returns a state with remaining_eth_budget=0 and finished=False, then check that the loop stops after one call and returns the collected batches.

chasingrainbows (approved):

PR #1007, opened 2026-08-14
fix(gloas): identify bunker CL rebase samples by state root

hweawer:
## Glamsterdam (Gloas / EIP-7732): bunker CL rebase samples must be identified by CL state, not by execution block

Stacked on #963 (`feat/gloas-blockstamp-infra`), which is where blockstamps start taking their
execution anchor from `state.latest_block_hash`. Independent of #964.

### Problem

`process_withdrawals` returns early when the parent payload is withheld:

```python
# specs/gloas/beacon-chain.md
if state.latest_block_hash != state.latest_execution_payload_bid.block_hash:
    return
```

`state.latest_block_hash` — the blockstamp's execution anchor — does not advance either, while CL
processing continues as normal: attestation rewards, penalties and slashings all still move
balances. So under Gloas several distinct CL states can share one execution block, and the two
identities that coincided before the fork no longer do.

The bunker's intraframe check keyed on the execution block in two places:

- `_calculate_cl_rebase_between_blocks` returned `Gwei(0)` when `prev.block_number == ref.block_number`;
- `_is_negative_specific_cl_rebase` deduplicated the nearest and distant samples by `block_number`.

In a window with no revealed payload, the first reports a zero rebase across balances that actually
moved, and the second collapses two genuinely different samples into one. The negative-rebase
detector then returns a false negative — during exactly the condition it exists to catch. Note the
detector only runs once the frame rebase is already below normal, so the failure is a missed
confirmation of bunker mode, not a spurious one.

### Fix

Both sites now key on `state_root`. Zero is returned only when the two blockstamps are literally the
same CL state — which is still reachable and still correct, since `validate_slot_distance` permits
`rebase_check_nearest_epoch_distance == 0`.

Pre-fork behaviour is unchanged: every beacon block embeds its own payload, so distinct blocks always
have distinct execution blocks and the two keys agree.

`_get_withdrawn_from_vault_between_blocks` needed a guard as well. It builds the event range as
`[prev.block_number + 1, ref.block_number]`, and `get_events_in_range` raises on `l_block > r_block`.
Without the guard, letting the calculation proceed past equal anchors turns a false negative into an
exception. Equal anchors mean no execution block passed, so no `ETHDistributed` could have been
emitted and zero is the correct value.

`_calculate_injected_capital` needs no change — its post-v4 path reads
`get_deposited_for_current_report` at each `block_hash` rather than over a range, so equal anchors
already yield zero.

### Reachability

The trigger needs a whole sampling window without a single revealed payload —
`rebase_check_nearest_epoch_distance` epochs, since execution block numbers advance once per
revealed payload. That is a builder-market collapse, not an ordinary withheld block.

This is filed as robustness rather than an urgent fix: the identity is wrong regardless of how often
it is reachable, the sampling distances are on-chain governance parameters that can change, and the
error direction is the unsafe one.

### Testing

- New `TestClSampleIdentityUnderEip7732` in `tests/modules/accounting/bunker/test_bunker_abnormal_cl_rebase.py`:
  same execution anchor with different state roots counts the CL delta; identical state roots still
  short-circuit to zero; the vault-withdrawal lookup returns zero without querying events; the
  nearest/distant pair sharing an execution anchor is checked twice rather than deduplicated.
- `test_calculate_cl_rebase_between_blocks` now builds its two blockstamps with distinct state roots.
  `ReferenceBlockStampFactory` defaults every blockstamp to one fixed `state_root`, so the samples
  were identical on the field the code now keys on.

### DIFF
--- a/src/services/bunker_cases/abnormal_cl_rebase.py
+++ b/src/services/bunker_cases/abnormal_cl_rebase.py
@@ -129,10 +129,10 @@ class AbnormalClRebase:
 
         nearest_blockstamp, distant_blockstamp = self._get_nearest_and_distant_blockstamps(blockstamp)
 
-        # dict preserves insertion order and deduplicates by block_number
+        # dict preserves insertion order and deduplicates by state_root
         for bs in {
-            nearest_blockstamp.block_number: nearest_blockstamp,
-            distant_blockstamp.block_number: distant_blockstamp,
+            nearest_blockstamp.state_root: nearest_blockstamp,
+            distant_blockstamp.state_root: distant_blockstamp,
         }.values():
             rebase = self._calculate_cl_rebase_between_blocks(bs, blockstamp)
             logger.info({"msg": f"Intraframe sampled CL rebase: {rebase} Gwei"})
@@ -189,8 +189,8 @@ class AbnormalClRebase:
         Check for these events is enough to account for all withdrawals since the protocol assumes that
         the vault can only be withdrawn at the time of the Oracle report between reference slots.
         """
-        if prev_blockstamp.block_number == ref_blockstamp.block_number:
-            # Can't calculate rebase between the same block
+        if prev_blockstamp.state_root == ref_blockstamp.state_root:
+            # Can't calculate rebase between the same CL state
             return Gwei(0)
 
         (prev_lido_validators, _) = LidoValidatorsProvider.compute_lido_validators(
@@ -258,6 +258,10 @@ class AbnormalClRebase:
                 )
             }
         )
+
+        if prev_blockstamp.block_number >= ref_blockstamp.block_number:
+            logger.info({"msg": "No execution blocks between samples. Vault withdrawals: 0 Gwei."})
+            return Gwei(0)
 
         events = self._get_eth_distributed_events(
             # We added +1 to prev block number because withdrawals from vault
--- a/tests/modules/accounting/bunker/test_bunker_abnormal_cl_rebase.py
+++ b/tests/modules/accounting/bunker/test_bunker_abnormal_cl_rebase.py
@@ -1,13 +1,15 @@
 from unittest.mock import Mock
 
 import pytest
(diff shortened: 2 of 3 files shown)


F4ever (approved):

F4ever (changes requested):
please check this
```
  1. src/services/bunker_cases/abnormal_cl_rebase.py:231: after the fork, the bunker check can report a negative rebase that isn't real (High)
  - What changed: After Gloas, a blockstamp reads the beacon state after ref_slot's child block. It reads the execution side at bid.parent_block_hash, which is before that child's payload runs
    (src/utils/blockstamp.py:86-148).
  - Why it breaks: The child block has already taken its payload's withdrawals out of validator balances. That ETH only reaches the withdrawal vault once the payload runs. So _get_lido_validators_balance_with_vault,
    which adds consensus-layer balances to the vault balance, comes up short by Lido's share of payload_expected_withdrawals. The PR adds BeaconStateView.payload_expected_withdrawals, but nothing reads it.
  - Example: The child block sweeps several full Lido exits while the comparison sample swept none. The rebase comes out hundreds of ETH negative and bunker mode turns on. The reported consensus-layer balance is
    short by the same amount.
  - Link to #964: This is the gap PR #964 fills: it adds in-flight withdrawals back on the consensus-layer side, including both ends of the abnormal-rebase check. If #964 is the planned follow-up, this is a problem
    of merge order, not a missing fix. Either land the two together, or make sure #1007 can't ship to a Gloas network on its own.
```

and 

```
  2. scripts/run_oracle_at_slot.py:118: the debug script crashes after the fork when --slot isn't given (Low)
  - Without --slot, the patched get_blockstamp_for_report sets ref_slot to the last finalized slot.
  - After Gloas, get_next_non_missed_slot raises ChildSlotNotFinalized for that slot (src/utils/slot.py:121).
  - consensus.py catches that exception, but the script's patched copy doesn't. 
 ```

PR #1012, opened 2026-08-28
feat(stc-892): cap VEBO exit requests and transaction gas

chasingrainbows:
## Description

Adds two safeguards for large VEBO exit reports:

- caps a single ejector report at 500 validator exit requests, including forced exits, with the remaining demand deferred to subsequent frames;
- replaces the fixed transaction gas cap with a configurable `TX_GAS_LIMIT`, defaulting to 55,000,000 gas.

Updates the public environment variable documentation and adds unit coverage for both limits.

## Related Issue/Task
- Related task: STC-892
- Epic: N/A

## How Has This Been Tested?
- [x] Local tests (`make test ORACLE_TEST_PATH='tests/modules/ejector/test_ejector.py tests/web3py/test_tx_utils.py -q'`) — 56 passed

## Checklist
- [x] Documentation updated (if required)


README.md
```diff
@@ -273,6 +273,7 @@ In manual mode all sleeps are disabled and `ALLOW_REPORTING_IN_BUNKER_MODE` is T
...
 | `DAEMON`                                                 | If False Oracle runs one cycle and ask for manual input to send report.                                                                                                     | False               | `True`                                       |
 | `TX_GAS_ADDITION`                                        | Used to modify gas parameter that used in transaction. (gas = estimated_gas + TX_GAS_ADDITION)                                                                              | False               | `100000`                                     |
+| `TX_GAS_LIMIT`                                           | Upper bound for the gas of a single transaction. The default is slightly below the mainnet block gas limit.                                                                 | False               | `55000000`                                   |
```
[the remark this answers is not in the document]
chasingrainbows: `.env.example` intentionally holds only the required connection settings - it has no optional tuning vars at all (`TX_GAS_ADDITION`, `CYCLE_SLEEP_IN_SECONDS` and the rest are also absent). The catalog for optional knobs is the README table, where `TX_GAS_LIMIT` is already listed. Adding only this one var to `.env.example` would be inconsistent.

src/modules/oracles/ejector/ejector.py
```diff
@@ -196,13 +197,30 @@ def get_validators_to_eject(
...
+                    'limit': MAX_EXIT_REQUESTS_PER_REPORT,
+                }
+            )
+            # Forced exits have priority, so the demand-based part is trimmed first.
+            forced_validators = forced_validators[:MAX_EXIT_REQUESTS_PER_REPORT]
+            validators_to_eject = validators_to_eject[: MAX_EXIT_REQUESTS_PER_REPORT - len(forced_validators)]
+
+        validators_to_eject.extend(forced_validators)
 
         logger.info(
```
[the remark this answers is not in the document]
chasingrainbows: This field is a diagnostic accumulator of the demand loop - it is compared against the predicted EL balance and it never included forced validators' balances even before this PR. The real report size is visible in `validators_to_eject_count`. Recomputing the total from the final list would change the field's meaning more than the trim does, so keeping it as is.

src/modules/oracles/ejector/ejector.py
```diff
@@ -196,13 +197,37 @@ def get_validators_to_eject(
 
                 if predictable_el_balance + gwei_to_wei(total_balance_to_eject_gwei) >= to_withdraw_amount:
                     break
+
+                if len(validators_to_eject) >= MAX_EXIT_REQUESTS_PER_REPORT:
```
hweawer: I think >

src/variables.py
```diff
@@ -47,6 +47,12 @@
 # In this case the second report will force report finalization and will consume more gas
 TX_GAS_ADDITION: Final = int(os.getenv('TX_GAS_ADDITION', 100_000))
 
+# Upper bound for the gas of a single oracle transaction.
+# The default is slightly below the 60M mainnet block gas limit, so a large report
+# (eg a VEBO exit report near the sanity checker ceiling) still fits in a block.
+TX_GAS_LIMIT: Final = int(os.getenv('TX_GAS_LIMIT', 55_000_000))
```
hweawer: 60M is a block limit - tx limit is 16,7M and it should be default

src/modules/oracles/ejector/ejector.py
```diff
@@ -196,13 +197,37 @@ def get_validators_to_eject(
...
+                    'msg': 'Report is over the max exit requests limit. The rest is reported in the next frames.',
+                    'len': total_requests,
+                    'limit': MAX_EXIT_REQUESTS_PER_REPORT,
+                }
+            )
+            # Forced exits have priority, so the demand-based part is trimmed first.
+            forced_validators = forced_validators[:MAX_EXIT_REQUESTS_PER_REPORT]
+            validators_to_eject = validators_to_eject[: MAX_EXIT_REQUESTS_PER_REPORT - len(forced_validators)]
+
+        validators_to_eject.extend(forced_validators)
```
dgusakov: Can we use `validators_to_eject = forced_validators.extend(validators_to_eject)[:MAX_EXIT_REQUESTS_PER_REPORT]` ?

src/variables.py
dgusakov: Use the exact value from https://eips.ethereum.org/EIPS/eip-7825 - 16,777,216 (2^24)

src/modules/oracles/ejector/ejector.py
```diff
@@ -204,6 +209,16 @@ def get_validators_to_eject(
...
 
+        if len(validators_to_eject) > MAX_EXIT_REQUESTS_PER_REPORT:
+            logger.warning(
+                {
+                    'msg': 'Report is over the max exit requests limit. The rest is reported in the next frames.',
+                    'len': len(validators_to_eject),
+                    'limit': MAX_EXIT_REQUESTS_PER_REPORT,
+                }
+            )
+            validators_to_eject = validators_to_eject[:MAX_EXIT_REQUESTS_PER_REPORT]
```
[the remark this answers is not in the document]
chasingrainbows: The forced-exit priority is already handled by `ValidatorExitIterator`: its sort predicate takes forced validators first inside the demand loop, and `get_remaining_forced_validators` returns only the leftover ones, which are lower priority by design (they are added only when the withdrawal queue is already covered). So the plain tail cut keeps the intended priority order and forced exits cannot be starved by demand-based ones.

src/variables.py
chasingrainbows: Fixed

src/modules/oracles/ejector/ejector.py
chasingrainbows: Raman noted this is already fine - forced go first in the iterator. Reverted the reordering, kept a plain tail cut.

src/variables.py
hweawer: https://github.com/lidofinance/lido-oracle/pull/1012/changes/8ec74e0cece7b4a34292a6447200355dd1e78f1b doesn't show the fix

hweawer (dismissed):

src/modules/oracles/ejector/ejector.py
```diff
@@ -204,6 +215,18 @@ def get_validators_to_eject(
             logger.info({'msg': 'Eject forced to exit validators.', 'len': len(forced_validators)})
             validators_to_eject.extend(forced_validators)
 
+        if len(validators_to_eject) > MAX_EXIT_REQUESTS_PER_REPORT:
+            logger.warning(
+                {
+                    'msg': 'Report is over the max exit requests limit. The rest is reported in the next frames.',
```
dgusakov:
```suggestion
                    'msg': f'Report is over the max exit requests limit. Capping to {MAX_EXIT_REQUESTS_PER_REPORT} exit requests. Dropped requests will be reported in the following frames.',
```

src/modules/oracles/ejector/ejector.py
```diff
@@ -196,6 +197,16 @@ def get_validators_to_eject(
 
                 if predictable_el_balance + gwei_to_wei(total_balance_to_eject_gwei) >= to_withdraw_amount:
                     break
+
+                if len(validators_to_eject) >= MAX_EXIT_REQUESTS_PER_REPORT:
```
F4ever: I propose moving this check into the iterator itself. It should raise `StopIteration` once the limit is reached.

You can check how this was handled before: `srv3.get_remaining_forced_validators` should already know the exact number of validators to return.

Two reasons:

1. We already have a similar limit in the balance iterator, so keeping the count there would be more convenient.
2. `get_remaining_forced_validators` should not iterate more than necessary; otherwise, we should review this function’s performance.

F4ever: use `_get_report_limits` and hardcode limit there with comment that this should be replaced with gloas

chasingrainbows: Done

src/utils/transaction.py
```diff
@@ -33,7 +33,7 @@ def build_transaction_params(w3: Web3, transaction: ContractFunction, account: L
         "nonce": w3.eth.get_transaction_count(account.address),
     }
 
-    if gas := estimate_gas(transaction, account):
+    if (gas := estimate_gas(transaction, account)) is not None:
```
F4ever: Why we comparing to None here?
Seems 0 is non valid value also

src/services/exit_order_iterator.py
F4ever, line 614: Maybe let's move check near balance check. Also it can be moved to separate func.
```
def _check_limits() -> bool:
    count_check = self.ejected_validators_count > self.max_exit_requests_per_report
    balance_check = self.max_current_exit_balance > self.exit_limit_in_gwei

    logger.log(... count_check and balance check)

    return count_check and balance_check
```

F4ever: and reuse in both places.

F4ever (dismissed):

F4ever (dismissed):

hweawer (dismissed):

hweawer (approved):

PR #1019, opened 2026-09-01
STC-887: feat: add CSM 0x02 oracle instance

hweawer:
## Description

Adds the CSM 0x02 staking module oracle.

CSM 0x02 is a separate deployment of the same CS* contract suite, so it gets its own `OracleModuleName.CSM_0X02` and entrypoint (`src/modules/oracles/staking_modules/community_staking_0x02/`) while reusing `SMPerformanceOracle`. `StakingModuleContracts` resolves Accounting, ParametersRegistry, FeeDistributor, FeeOracle and Strikes from `STAKING_MODULE_ADDRESS`, so the module needs no new contract wiring. Required-variable checks now key off `STAKING_MODULE_ORACLES` instead of listing module names one by one.

Verified on Hoodi that the new deployment is compatible with the existing `csm` module:

| | CSM | CSM 0x02 |
|---|---|---|
| CSModule | `0x79CEf36D…E59d` | `0xbb7dd81F…CAf9` |
| FeeOracle | `0xe7314f56…028B` | `0x9B8bBA11…A208` |
| HashConsensus | `0x54f74a10…C637` | `0x41142D07…BF34` |
| `getContractVersion()` | 3 | 3 |
| `getConsensusVersion()` | 4 | 4 |
| HashConsensus members | 10 | same 10, quorum 6 |
| epochsPerFrame | 3150 | 3150 |

The member set is identical, so the existing `MEMBER_PRIV_KEY` works without changes.

Changes:
- `OracleModuleName.CSM_0X02`, `STAKING_MODULE_ORACLES`, dispatch in `src/main.py`
- `community_staking_0x02/` package: `CSM0x02PerformanceOracle` + entrypoint
- `docker-compose.yml`: `csm-0x02-oracle` service, `CS_MODULE_0X02_ADDRESS`, ports 9005/9015
- `.env.example`, `README.md`, `docs/development.md`, `agents.md`
- `docs/monitoring/`: scrape job and availability alert

Notes for operators:
- The Hoodi HashConsensus `initialEpoch` is 121738 (~2026-09-09 13:33 UTC). Until then `getCurrentFrame()` reverts with `InitialEpochIsYetToArrive`; this is handled by `get_initial_or_current_frame`, so the instance can be started early and will wait.
- Frames are the same length as CSM but offset by 488 epochs, so the performance collector holds a wider union of epoch ranges (~16 days instead of ~14). No collision on collector state: the demand key is the FeeOracle address.

[...]

PR #1020, opened 2026-09-02
fix: drive the signer in fork report cycles

hweawer:
## Description

The **Mainnet Fork Tests** workflow has been failing on every branch (28 failures, all identical):

```
AttributeError: 'Web3' object has no attribute 'signer'
  src/modules/oracles/common/consensus.py:429  ->  self.w3.signer.process_members(members)
```

`tests/fork/test_lido_oracle_cycle.py` and `tests/fork/test_staking_module_oracle_cycle.py` act as each committee member with

```python
with account_from(private_key):
    module.cycle_handler()
```

`account_from` patches `variables.ACCOUNT`, but nothing in the report path reads that variable any more — since `SignerModule` landed, the active member is resolved through `w3.signer`, built once at startup in `_build_web3_base` (`runtime.py:69-78`). The fork `web3` fixture never attached it, so the cycle died before submitting anything, and patching the account had no effect either way.

`tests/fork/test_delegation_report_cycle.py` already had a local `signer_from` helper doing the right thing, which is why it stayed green.

Changes:
- `account_from` yields the account it switched to.
- `signer_from` moves out of the delegation test into `tests/fork/conftest.py`, with defaults so a plain EOA member reads as `signer_from(account)`.
- Both cycle tests rebuild the signer per member: `with account_from(pk) as account, signer_from(account):`.
- The fork `web3` fixture gets a `telemetry_data_bus` stub. Its absence never failed a test (`_try_send_telemetry` swallows everything) but it logged a failed DataBus send with a full traceback on every cycle, which buries the real failure in the CI log.

Not attaching a default signer to the `web3` fixture on purpose: an empty `SignerModule` would turn "test forgot to drive the signer" into a silent dry-run whose assertions fail confusingly, instead of an immediate `AttributeError`.

## Related Issue/Task
- Unblocks CI on #1019 and every other open PR

## How Has This Been Tested?
- [x] Local tests (`pytest -m unit`; `pytest tests/fork --collect-only` and `--setup-plan` to check the fixture graph)
- [ ] Manual testing — a mainnet-fork run needs an archive EL, a beacon node with history and KAPI, so the real verification is this PR's **Mainnet Fork Tests** run

## Checklist
- [x] Documentation updated (if required) — n/a, test-only
- [x] New tests added (if applicable) — existing fork tests restored

### DIFF
--- a/.github/workflows/mainnet_fork_tests.yml
+++ b/.github/workflows/mainnet_fork_tests.yml
@@ -13,6 +13,8 @@ on:
       - develop
     paths:
       - "src/**"
+      - "tests/fork/**"
+      - ".github/workflows/mainnet_fork_tests.yml"
 
 permissions:
   contents: read
--- a/src/modules/oracles/common/oracle_module.py
+++ b/src/modules/oracles/common/oracle_module.py
@@ -95,7 +95,8 @@ class OracleModule[W3: Web3Base](DaemonModule, ConsensusModule[W3], ABC):
         except KeysOutdatedException as error:
             logger.error({'msg': 'Keys API service returns outdated data.', 'error': str(error)})
         except Web3Exception as error:
-            logger.error({'msg': 'Web3py exception.', 'error': str(error)})
+            # A bare revert selector says nothing about which call produced it.
+            logger.error({'msg': 'Web3py exception.', 'error': ''.join(traceback.format_exception(error))})
         except IPFSError as error:
             logger.error({'msg': 'IPFS provider error.', 'error': str(error)})
         except ValueError as error:
--- a/tests/fork/conftest.py
+++ b/tests/fork/conftest.py
@@ -45,6 +45,7 @@ from src.web3py.extensions import (
     LidoValidatorsProvider,
     TransactionUtils,
 )
+from src.web3py.extensions.signer import SignerModule
 from src.web3py.extensions.staking_module import StakingModuleContracts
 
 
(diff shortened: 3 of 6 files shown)


hweawer:
CI result on this branch (run [33686078704](https://github.com/lidofinance/lido-oracle/actions/runs/33686078704)): **3 failed, 14 passed**, down from 14 failed / 0 passed on `develop`.

All 8 `test_staking_module_oracle_cycle` cases and 3 of 6 `test_lido_oracle_cycle` cases now report and pass the ref-slot assertion, so the signer wiring is confirmed fixed.

The 3 remaining failures are pre-existing and unrelated to the signer — they were simply unreachable while every cycle died on `AttributeError` before submitting anything:

1. **`test_lido_module_report[…accounting_module]`, `start_before_initial_epoch` and `start_after_initial_epoch`** — the cycle aborts during report building with a raw revert logged by the generic handler:

   ```
   ERROR src.modules.oracles.common.oracle_module:98 {'msg': 'Web3py exception.', 'error': "('0xcd0883ea', '0xcd0883ea')"}
   ```

   `0xcd0883ea` = `InitialEpochIsYetToArrive()`. Both scenarios build the report at ref slot `15129247` = `initial_epoch * 32 - 1`, i.e. the last slot **before** the fresh HashConsensus's `initialEpoch` that `new_hash_consensus` sets. Any consensus read executed at that block reverts. The oracle's own frame reads are guarded (`get_initial_or_current_frame`, `_get_processing_state`), so this is an unguarded read reached from inside the report build — the exact call site is not recoverable from the log because `oracle_module.py:98` logs the error without `exc_info`. The ejector passes those same scenarios, which fits: it does not simulate a rebase.

2. **`test_lido_module_report[…ejector_module]`, `start_before_initial_epoch`** — upstream archive RPC timing out inside anvil:

   ```
   failed to get storage for 0xdA7dE2EC… : connection error; Connection timed out (os error 110)
   ```

   The `DuplicateReport()` (`0xfd10cf73`) revert later in that test is downstream of the aborted cycle, not a separate cause. This one is a flake.

Both are separate issues from what this PR fixes; I would rather land the signer fix and chase them on their own PRs than grow this one. Happy to add `exc_info=True` to the handler in (1) so the next run names the reverting call, if that is worth a CI cycle.

PR #1025, opened 2026-09-07
fix: address Composable Security review findings for v8.1 (LDO-af6d118a)

chasingrainbows:
## Description
Fixes from the Composable Security review of Lido Oracle v8.1 (report LDO-af6d118a, commit af6d118a).

- LDO-af6d118a-L01: the third phase of the accounting report now re-resolves the active signer after the sleep, not only the blockstamp. A member or delegate change enacted during the sleep is picked up before the extra data transaction is sent.
- LDO-af6d118a-L02: when the DataBus telemetry transaction is not confirmed within the timeout, `_send_telemetry` returns None and logs a warning with the tx hash instead of reporting it as sent. Telemetry stays fire-and-forget: no exception, no retry, no extra waiting, so the report flow is not affected.
- LDO-af6d118a-R02: the warning for a reporting address that is not a HashConsensus member and has no submit role now describes the observed condition and logs the reporting address and block hash.

R01 is not changed on purpose: the redundant operand in the submit delay calculation narrows the type for pyright.

Also: `DELEGATION_CONTRACT_ADDRESS` is now passed to the oracle services in `docker-compose.yml`.

## Related Issue/Task
- Related task: STC-903
- Epic: EDF (Execution Delegation Framework)

## How Has This Been Tested?
- [x] Local tests (`make test` for accounting, consensus, telemetry and smoke tests, `make lint`)
- [ ] Manual testing
- [ ] Not tested

## Checklist
- [ ] Documentation updated (not required)
- [x] New tests added


src/web3py/extensions/telemetry_data_bus.py
```diff
@@ -167,7 +168,16 @@ def _send_telemetry(self, tx: ContractFunction, w3: Web3, account: LocalAccount)
             time.sleep(_POLL_INTERVAL_SECONDS)
 
         if tx_hash:
-            return tx_hash
+            # The transaction was broadcast but not confirmed within the time budget.
```
hweawer, line 171: nit: actually log line says the same

hweawer (approved):

PR #1027, opened 2026-09-18
fix: relax the container healthcheck budgets

hweawer:
## Problem

The oracle containers on the testnet bots host flap between healthy and unhealthy, and `docker-autoheal` restarts them. Four restarts of `lido-oracle-accounting` in under three hours, with `lido-oracle-cm` going down 11 seconds before one of them.

The image healthcheck is the aggressive part:

```dockerfile
HEALTHCHECK --interval=10s --timeout=3s
```

No `--retries`, no `--start-period`, so Docker applies 3 retries and zero startup grace: **30 seconds of slow probing is enough to mark a container unhealthy, from its first second of life.**

`infra-mainnet` sets a `healthcheck:` override only for postgres, performance-collector and performance-web. The accounting, ejector, csm, csm-0x02 and cm services have none, so these image values are what actually runs in every environment — and each module is capped at 1.0 CPU there.

The probe `exec`s into the container and shares that single-core quota with the oracle loop, while starting a Python interpreter and importing `requests`. Measured at **~0.1s of CPU per probe**; a saturated cgroup stretches that past 3s.

## Evidence this is the probe and not the endpoint

The healthcheck server answers 503 only when `now - _last_pulse > MAX_CYCLE_LIFETIME_IN_SECONDS`, and that variable is not set anywhere in infra, so it defaults to **3000s**. Pulses arrive about every **380s** — once per epoch, since `pulse()` runs only in `run_cycle()` and the skip branch in `daemon_module.py:63` returns before it. An eightfold margin, nowhere near tripping.

Log evidence agrees. `Execute module.` — the only line followed by `pulse()` — appears immediately before and after every restart, so the pulse was fresh when the container was killed. And the fatal gap was 333s while gaps of 355s and 378s passed without incident, which no staleness threshold can produce.

## Change

```dockerfile
HEALTHCHECK --interval=30s --timeout=15s --retries=5 --start-period=120s
```

Applied to the image and to `x-healthcheck-defaults` in this repo's compose file so the two do not drift. 120s of grace plus 5 × 30s, so 150s of sustained failure is needed instead of 30s.


hweawer:
Closing: the fix belongs in infra, not in the image.

External operators run this image with 3s/10s and have no problem, which matches measurement — with no CPU quota the probe takes 0.19s, so 3s is a 15x margin. The flapping is specific to our deployment, where each oracle module is capped at 1.0 CPU (`service_oracle/defaults/main.yml:23-32`) on a host carrying ~24 containers with ~22 vCPU of aggregate limits.

Worth recording why the earlier tuning did not help: the oracle template has only ever had three `healthcheck:` blocks — postgres, performance-collector and performance-web. `d7935446c` "oracle healtchecks tuning" raised retries 3 to 5 on those, but accounting, ejector, csm, csm-0x02 and cm have never had an override at all, so they kept inheriting the image defaults.

Moving to infra-mainnet: a `healthcheck:` block for the five oracle services, plus the CPU question.

